PR #2, opened 2022-10-03
[FEATURE] - Sage Link

ju-Skinner:
# Description

Fixes #(issue)

## Type of change

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- sage versions:
- os:
- browsers:
- screen readers:
- misc:

# Checklist:


libs/core/src/components.d.ts
```diff
@@ -47,15 +74,37 @@ declare namespace LocalJSX {
          */
         "middle"?: string;
     }
+    interface SageLink {
```
goodwinchris, line 77: I understand this is in a different block, but why do we _need_ to define this information twice?

libs/core/src/components/sage-link/sage-link.scss
```diff
@@ -0,0 +1,68 @@
...
+  svg {
+    margin-left: 4px;
+  }
+}
+
+:host(:focus) {
+  border-radius: 6px;
+  outline: 4px solid #8ecafb;
+  outline-offset: 4px;
+}
```
goodwinchris, line 13: If we want to use `:host` thats ok with me but we should define some clear rules for when we use `:host` and when we use top level classes like `.sage-link`

libs/core/src/components/sage-link/sage-link.scss
```diff
@@ -0,0 +1,68 @@
...
+.sage-link--md {
+  font-size: 14px;
+
+  svg {
+    height: 13px;
+    width: 13px;
+  }
+}
+
+.sage-link--lg {
```
goodwinchris, line 43: Per the spec, I was under the impression that font sizing should be derived from the parent element?

libs/core/src/components.d.ts
ju-Skinner: This file is autogenerated

libs/core/src/components/sage-link/sage-link.scss
```diff
@@ -0,0 +1,68 @@
...
+  font-size: 16px;
+
+  svg {
+    height: 15px;
+    width: 15px;
+  }
+}
+
+.sage-link--plain {
+  font-weight: 500;
```
goodwinchris, line 53: Same as above, I was under the impression that font weight should be derived from the parent element?

libs/core/src/components/sage-link/sage-link.tsx
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * The URL that the hyperlink points to.
+   */
+  @Prop() href!: string;
+
+  render() {
+    const externalIcon = (
+      <svg role="presentation" width="15" height="14" viewBox="0 0 15 14" fill="none" xmlns="http://www.w3.org/2000/svg">
+        <path
+          fill-rule="evenodd"
```
goodwinchris, line 43: We can likely clean up this SVG - fill-rules and clip-rules usually aren't needed. Further, we could do some fill cleanup etc...

libs/core/src/components.d.ts
ju-Skinner: Also, an `interface` is different than a `class`. An `interface` is a contract stating what should be in a class that inherits the interface.

libs/core/src/components/sage-link/test/sage-link.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+    });
+
+    const svg = root.shadowRoot.querySelector('svg');
+    expect(svg).not.toBeNull();
+
+    const svgPath = svg.querySelector('path');
+
+    const attr = svgPath.attributes.getNamedItem('d').value;
+    expect(attr).toEqual(
+      'M13.7458 0.75419C13.802 0.810351 13.8443 0.875115 13.8727 0.944239C13.9008 1.01249 13.9155 1.08498 13.9166 1.15765L13.9166 1.16766V5.83334C13.9166 6.1555 13.6555 6.41667 13.3333 6.41667C13.0111 6.41667 12.75 6.1555 12.75 5.83334V2.57496L7.32913 7.99581C7.10132 8.22362 6.73197 8.22362 6.50417 7.99581C6.27636 7.76801 6.27636 7.39866 6.50417 7.17086L11.925 1.75H8.66665C8.34448 1.75 8.08331 1.48884 8.08331 1.16667C8.08331 0.844503 8.34448 0.583336 8.66665 0.583336H13.3332H13.3333C13.4826 0.583336 13.6319 0.640286 13.7458 0.75419ZM1.66665 1.75C1.34448 1.75 1.08331 2.01117 1.08331 2.33333V12.8333C1.08331 13.1555 1.34448 13.4167 1.66665 13.4167H12.1666C12.4888 13.4167 12.75 13.1555 12.75 12.8333V8.75C12.75 8.42783 12.4888 8.16667 12.1666 8.16667C11.8445 8.16667 11.5833 8.42783 11.5833 8.75V12.25H2.24998V2.91666H5.74998C6.07215 2.91666 6.33331 2.6555 6.33331 2.33333C6.33331 2.01117 6.07215 1.75 5.74998 1.75H1.66665Z'
```
goodwinchris, line 52: Test will break if we do optimize the SVG - just a note

libs/core/src/components/sage-link/sage-link.scss
ju-Skinner: @goodwinchris Spec? Is that defined somewhere, e.g Figma doc.

libs/core/src/components/sage-link/test/sage-link.spec.tsx
ju-Skinner: @goodwinchris Agreed...we noted that in our pairing sessions and chalked it up to "need to get our icon library going". This will be fine for now as long as it doesn't start to grow and bleed into other components.

ju-Skinner:
This has been closed in favor of #4 which is based on the new Repo structure.

PR #5, opened 2022-10-24
[DSS-203] Image Component

anechol:
# Description

Fixes #(issue)

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [x] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- sage versions:
- os:
- browsers:
- screen readers:
- misc:

# Checklist:

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+libs/react
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,0 @@
-libs/react
--- a/libs/core/src/components/sage-image/sage-image.scss
+++ b/libs/core/src/components/sage-image/sage-image.scss
@@ -1,0 +1,3 @@
+:host {
+  display: block;
+}
--- a/libs/core/src/components/sage-image/sage-image.scss
+++ b/libs/core/src/components/sage-image/sage-image.scss
@@ -1,3 +1,12 @@
 :host {
-  display: block;
+  /**
+   * @prop --image-aspect-ratio: The image's aspect ratio.
+  */
+
+  --image-aspect-ratio: auto;
+  display: inline-block;
+}
+
+img {
+  aspect-ratio: var(--image-aspect-ratio);
 }
--- a/libs/core/src/components/sage-image/sage-image.scss
+++ b/libs/core/src/components/sage-image/sage-image.scss
@@ -9,4 +9,6 @@
 
 img {
   aspect-ratio: var(--image-aspect-ratio);
+  height: auto;
+  width: 100%;
 }
--- a/libs/core/src/components/sage-image/sage-image.tsx
+++ b/libs/core/src/components/sage-image/sage-image.tsx
@@ -1,0 +1,31 @@
+import { Component, Host, h, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'sage-image',
+  styleUrl: 'sage-image.scss',
+  shadow: true,
+})
+export class SageImage {
+  /**
+   * The image's alt tag. If none is provided,
+   * it will default to an empty string.
+  */
+ @Prop() alt? = '';
+
+ /**
+  * The image's source.
+  */
+ @Prop() src: string;
+
+  render() {
+    return (
+      <Host
+        class={{
+          'sage-image': true,
+        }}
+      >
+        <img alt={this.alt} src={this.src} />
+      </Host>
+    );
+  }
+}
--- a/libs/core/src/components/sage-image/sage-image.tsx
+++ b/libs/core/src/components/sage-image/sage-image.tsx
@@ -21,7 +21,6 @@ export class SageImage {
 
   /**
    * Indicates how the browser should load the image.
-   * Defaults to "eager".
    */
   @Prop() loading?: 'eager' | 'lazy' = 'eager';
 
--- a/libs/core/src/components/sage-image/sage-image.tsx
+++ b/libs/core/src/components/sage-image/sage-image.tsx
@@ -26,9 +26,21 @@ export class SageImage {
  @Prop() loading?: 'eager' | 'lazy' = 'eager';
 
  /**
+  * Determines the intended display size of an image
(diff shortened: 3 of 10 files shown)


.eslintignore
```diff
@@ -0,0 +1 @@
+libs/react
```
ju-Skinner: ~~you can take this out now.~~ delete the whole file

libs/core/src/components/sage-image/sage-image.scss
```diff
@@ -0,0 +1,14 @@
...
+  /**
+   * @prop --image-aspect-ratio: The image's aspect ratio.
+  */
+
+  --image-aspect-ratio: auto;
+  display: inline-block;
+}
+
+img {
+  aspect-ratio: var(--image-aspect-ratio);
```
QuintonJason, line 11: 😍 👏

QuintonJason:
When clicking through the `Properties` on the `Docs` page, some of the values aren't updating the live demo. Is this the intended behavior for the `Default` demo? (i.e.: `alt` works, but `loading` doesn't)

anechol:
> When clicking through the `Properties` on the `Docs` page, some of the values aren't updating the live demo. Is this the intended behavior for the `Default` demo? (i.e.: `alt` works, but `loading` doesn't)

Thanks for pointing this out @QuintonJason. After a bit of testing, this is due to the default story not being set up with all of the props. This particular story only has the `alt` and `src` props set up and the others cannot be updated in real time as a result. This does bring up the question of "should our default story have every prop available?" I definitely think we should discuss this in our retro.

libs/core/src/components/sage-image/test/sage-image.e2e.ts
```diff
@@ -0,0 +1,11 @@
...
+
+describe('sage-image', () => {
+  it('renders', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<sage-image></sage-image>');
+
+    const element = await page.find('sage-image');
+    expect(element).toHaveClass('hydrated');
+  });
+});
```
ju-Skinner, line 26: I'd like to see a few more test here. These test would be beneficial as they would render the actual markup we would see in the browser.

libs/core/src/components/sage-image/stories/sage-image.stories.mdx
```diff
@@ -0,0 +1,61 @@
...
+  >
+    {SizesAndSrcset.bind()}
+  </Story>
+</Canvas>
+
+## References
+
+[MDN - Image](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img)
+
+[HTML Standard - Image](https://html.spec.whatwg.org/multipage/images.html)
```
ju-Skinner: Great job on the documentations :heart:

libs/core/src/components/sage-image/stories/sage-image.stories.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+# Image
+
+An image can be used to convey supplemental information to the user or as decoration.
+
+## Guidelines
+
+### Accessibility
+
+All images, regardless of use, must have an appropriate alt text to describe the image for screen reader users. If the image only provides visual decoration to the page, be sure to add `alt=""` so that screen readers will properly skip over the element. For more information on how to properly use alt text for images, please refer to the [alt decision tree](https://www.w3.org/WAI/tutorials/images/decision-tree/).
```
ju-Skinner: Should we as the implementors provide a fallback should one not be applied? e.g the name of the file

libs/core/src/components/sage-image/sage-image.tsx
```diff
@@ -0,0 +1,70 @@
...
+   */
+  @Prop() srcset?: string
+
+  /**
+   * The width of the image in pixels. Setting this will
+   * devote space in the layout to prevent layout
+   * shifts when the image is loaded.
+   */
+  @Prop() width?: number;
+
```
ju-Skinner, line 50: we could write a helper function which could check to see if `alt` is provided, otherwise use the `filename` located in the `src` path / url. What are your thoughts?

ju-Skinner (changes requested):
Great job thus far. I left a few comments. The lack of `e2e` test is what blocks approval, otherwise great job.

libs/core/src/components/sage-image/stories/sage-image.stories.mdx
anechol: No, and I should actually consider removing or revising that sentence as the `alt` prop already defaults to an empty string. The reason for the no here is because I don't think we should assume that any image added should have context as there will inevitably be situations where the image will have no significance to the context of the page (i.e. it's decoration). Decorative images should always have an empty `alt` in order to be skipped over being read in the accessibility tree for screen reader users. This is putting more of the onus on the engineers to make the correct decision about proper usage, but that's also why I linked a page here that _should_ make that decision easy.

libs/core/src/components/sage-image/sage-image.tsx
anechol: Explained in above comment.

libs/core/src/components/sage-image/sage-image.tsx
```diff
@@ -0,0 +1,70 @@
...
+  tag: 'sage-image',
+  styleUrl: 'sage-image.scss',
+  shadow: true,
+})
+export class SageImage {
+  /**
+   * The image's alt tag. If none is provided,
+   * it will default to an empty string.
+   */
+  @Prop() alt? = '';
```
goodwinchris, line 13: Love that you supply an empty alt!

goodwinchris (approved):
This looks good to me, love the attention that went into the docs and the intention that went into a11y

ju-Skinner (approved):
Great job...Sorry it took so long to review and give approval.

QuintonJason (approved):
Looks and works great!

PR #13, opened 2023-01-23
Docs - Spread Directive

anechol:
# Description

[DSS-280](https://kajabi.atlassian.net/browse/DSS-280)

Adds a helper directive for LitHtml to handle spreading props for components within storybook. This intent of this PR is for discovery and for opening conversation with the team to determine if this is a direction we'd like pursue.

Screen Shot 2023-01-20 at 4 23 19 PM

Screen Shot 2023-01-20 at 4 23 35 PM

Added the following dependencies:
- `@open-wc/lit-helpers`
- `lit`

## Type of change

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- sage versions:
- os:
- browsers:
- screen readers:
- misc:

# Checklist:

[DSS-280]: https://kajabi.atlassian.net/browse/DSS-280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/package.json
+++ b/libs/core/package.json
@@ -53,6 +53,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.19.3",
+    "@open-wc/lit-helpers": "^0.5.1",
     "@pxtrn/storybook-addon-docs-stencil": "^6.4.1",
     "@stencil/eslint-plugin": "^0.4.0",
     "@stencil/react-output-target": "^0.3.1",
@@ -74,6 +75,7 @@
     "eslint-plugin-storybook": "^0.6.6",
     "jest": "^27.4.5",
     "jest-cli": "^27.4.5",
+    "lit": "^2.6.1",
     "lit-html": "^2.4.0",
     "npm-run-all": "^4.1.5",
     "postcss": "^8.4.18",
--- a/libs/core/src/components/sage-link/stories/sage-link.stories.mdx
+++ b/libs/core/src/components/sage-link/stories/sage-link.stories.mdx
@@ -26,7 +26,9 @@ export const TemplateWithSlot = (args) => html` <sage-link ${spread(args)}>${arg
 ### External
 
 <Canvas>
-  <Story name="External">{Template.bind()}</Story>
+  <Story name="External" args={{ external: true }}>
+    {Template.bind()}
+  </Story>
 </Canvas>
 
 ### Inline
--- a/libs/core/src/components/sage-link/stories/sage-link.stories.mdx
+++ b/libs/core/src/components/sage-link/stories/sage-link.stories.mdx
@@ -1,14 +1,14 @@
 import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs';
 import { html, render } from 'lit-html';
-import { spread, spreadProps } from '@open-wc/lit-helpers';
+import { spread } from '@open-wc/lit-helpers';
 
 import { extractArgTypes } from '@pxtrn/storybook-addon-docs-stencil';
 
 <Meta title="webcomponents/sage-link" component="sage-link" argTypes={extractArgTypes('sage-link')} args={{ href: '#some-anchor' }} />
 
-export const Template = (args) => html`<sage-link ${spreadProps(args)}></sage-link>`;
+export const Template = (args) => html`<sage-link ${spread(args)}></sage-link>`;
 
-export const TemplateWithSlot = (args) => html` <sage-link ${spreadProps(args)}>${args.slot}</sage-link>`;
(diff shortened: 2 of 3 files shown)


libs/core/package.json
```diff
@@ -74,6 +75,7 @@
     "eslint-plugin-storybook": "^0.6.6",
     "jest": "^27.4.5",
     "jest-cli": "^27.4.5",
+    "lit": "^2.6.1",
```
ju-Skinner, line 78: This feels a little confusing...Are we building `lit` or `stencil` components? 😆  Luckily it's only a development dependency. So it's not a big deal to me.

libs/core/src/components/sage-link/stories/sage-link.stories.mdx
```diff
@@ -1,12 +1,14 @@
...
 import { html, render } from 'lit-html';
+import { spread } from '@open-wc/lit-helpers';
 
 import { extractArgTypes } from '@pxtrn/storybook-addon-docs-stencil';
 
 <Meta title="webcomponents/sage-link" component="sage-link" argTypes={extractArgTypes('sage-link')} args={{ href: '#some-anchor' }} />
 
-export const Template = (args) => html`<sage-link external=${args.external} font-size=${args.fontSize} href=${args.href} variant=${args.variant}></sage-link>`;
-export const TemplateWithSlot = (args) => html` <sage-link external=${args.external} font-size=${args.fontSize} href=${args.href} variant=${args.variant}>${args.slot}</sage-link>`;
+export const Template = (args) => html`<sage-link ${spread(args)}></sage-link>`;
```
ju-Skinner, line 9: So my initial thought was a bit of a concern as this lends itself to hide details or cause side affects. As I sit here now, I believe if we set the standard to only use this in Storybook it will be fine to do so. 

To further expand on the need to do this versus being explicit. When a developer would look at code examples, should an attribute not be used it would still render in the output showing an empty value. This could cause bad practices and misuse of the components, so the `spread` would reduce unused attributes.

ju-Skinner (approved):
Approving based on comments and agreement on how we use `spread`.

libs/core/package.json
anechol: Lit was added to solve a console error where the spread directive was searching for a missing dependency. If we did decide to move forward with this hypothetically, would it make more sense to create our own solution or leverage what's already been made?

libs/core/src/components/sage-link/stories/sage-link.stories.mdx
anechol: I'm not opposed to using this if it's only for Storybook and provides us benefits in our docs. I like the approach to being more specific, personally. The question I have for use of `spread` is that do we _need_ to add it? Could we come up with an alternate solution instead, e.g. having one/the initial story dedicated to sandboxing and all other stories dedicated to common patterns? Furthermore, should common patterns be a section in our docs page?

goodwinchris: I'm torn too. I don't mind using `spread` but I do also wonder if we could just split up our code examples and take on the addition of a template for each example. Maintaining too many templates could be a good "canary in the coal mine" for when a component is getting too complex. 

That said, I don't mind if we want to use `spread` instead of adding multiple templates.

QuintonJason: I like the idea, but I can see where confusion could arise. Currently we have the issue where the attributes are showing in the DOM due to our main template containing all props. Because of that, it's helpful for that template but I also like the explicit way of setting them to reduce confusion as well. I don't mind multiple template files since it could ensure correct usage.

anechol: Based on the conversation here, it seems as if there's no real definitive direction that we would like to pursue regarding adding this helper. I'm going to vote that we dismiss using `spread` and come up with an alternate solution for now like creating different stories for different use cases. If that gets too out of hand, then maybe we can reconsider `spread` usage in the future. Any objections?

I am more than happy to spend some time looking into an avenue for how this can look in the docs as long as it doesn't interfere with Quinton's work on docs improvement.

QuintonJason (approved):

libs/core/src/components/sage-link/stories/sage-link.stories.mdx
goodwinchris: Sounds good to me, love having this PR here (even if its closed) so we can come back to spread if we ever change our minds. Thank you for digging into this!

PR #15, opened 2023-02-02
[DSS-43]: Tabs Component(s)

goodwinchris:
# Description

Addition of `Sage-Tabs`, `Sage-Tab`, and `Sage-Tabpanel` components. 

CleanShot 2023-02-17 at 17 11 56

Fixes DSS-43

## Testing

1. Navigate to component in storybook: http://localhost:6006/?path=/story/webcomponents-tabs--primary
2. Confirm basic functionality is in place
3. Confirm updating of expected aria/a11y attributes as outlined here: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/
4. Confirm keyboard navigation works as expected (arrow keys, home key, end key)

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

# Checklist:

# Followup tasks: 
- [ ] Ensure tabs/tabpanels have unique ID's by concat'ing their parentComponent id into their unique identifiers.


libs/core/.eslintrc.json
```diff
@@ -7,6 +7,8 @@
   "plugins": ["@typescript-eslint"],
   "rules": {
     "@stencil/decorators-context": "off", // This is set to off until support for ESLINT 8, othwerise remove rule
-    "@stencil/strict-boolean-conditions": "off" // This is set to off until support for ESLINT 8, othwerise remove rule
+    "@stencil/strict-boolean-conditions": "off", // This is set to off until support for ESLINT 8, othwerise remove rule
+    "@stencil/decorators-style": "off",
+    "@stencil/own-methods-must-be-private": "off"
```
ju-Skinner: Can you explain the need for adding these?

libs/core/.stylelintrc.json
```diff
@@ -26,6 +26,7 @@
           "custom-elements"
         ]
       }
-    ]
+    ],
+    "max-nesting-depth": 3
```
ju-Skinner, line 30: Same comment as above

libs/core/src/components/sage-tab/sage-tab.scss
```diff
@@ -0,0 +1,177 @@
...
+  background-color: transparent;
+  border: 0;
+  box-shadow: none;
+  color: inherit;
+  cursor: pointer;
+  font-family: inherit;
+  padding: 0;
+}
+
+// delete later
```
ju-Skinner, line 20: when would be an appropriate time to delete it?

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Sets the related tab name, this name must match a `sage-tabpanel`'s tab property, required
+  */
+  @Prop() tab: string;
+
+  /**
+   * Keeps track of the activeTab, this property is passed in by parent component
+  */
+  @Prop({mutable: true}) activeTab: string;
```
ju-Skinner: Why the use of mutable?

libs/core/.eslintrc.json
goodwinchris, line 12: Check in on the need for these

libs/core/src/components/sage-tab/sage-tab.scss
goodwinchris, line 20: Open followup ticket for this work

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Keeps track of if the tabpanel is selected, this property is computed on `componentWillUpdate()`
+  */
+  @Prop({mutable: true}) selected = false;
+ 
+  /**
+   * Emits an event upon tab click for `sage-tab` and `sage-tabpanel` to listen for
+   */
+  @Event({composed: false}) tabClick: EventEmitter<object>;
```
goodwinchris: Check if composed still needed

ju-Skinner: Why did you elect to set `composed` to false?

libs/core/src/components/sage-tab/stories/sage-tab.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+<ArgsTable of="sage-tab" />
+
+## Title
+
+<Canvas>
+  <Story name="Default">
+    {Template.bind()}
+  </Story>
+</Canvas>
```
ju-Skinner: Can you follow the new pattern established with the `sage-button` stories?

libs/core/src/components/sage-tab/test/sage-tab.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+    });
+    expect(page.root).toEqualHtml(`
+      <sage-tab slot="tabs">
+        <button aria-controls="undefined-panel" aria-selected="false" class="sage-tab" role="tab" tabindex="-1">
+          <div class="sage-tab__content"></div>
+        </button>
+      </sage-tab>
+    `);
+  });
+});
```
ju-Skinner, line 76: We can add some more tests from a "unit" testing standpoint. Examples, setting props and what the expected output may be.

libs/core/src/components/sage-tabpanel/stories/sage-tabpanel.stories.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+## Properties
+
+<ArgsTable of="sage-tabpanel" />
+
+<Canvas>
+  <Story name="Default">
+    {Template.bind()}
+  </Story>
+</Canvas>
```
ju-Skinner: Same comment regarding the new Story structure

libs/core/src/components/sage-tabpanel/test/sage-tabpanel.spec.tsx
```diff
@@ -0,0 +1,16 @@
...
+      components: [SageTabpanel],
+      html: `<sage-tabpanel></sage-tabpanel>`,
+    });
+    expect(page.root).toEqualHtml(`
+      <sage-tabpanel slot="tabpanels">
+        <div class="sage-tabpanel" id="undefined-panel" role="tabpanel"></div>
+      </sage-tabpanel>
+    `);
+  });
+});
```
ju-Skinner, line 45: Same comment regarding unit testing the props and expected behavior

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,164 @@
+import { Component, Element, Host, h, Prop, Listen } from '@stencil/core';
+@Component({
```
ju-Skinner, line 7: add new line here

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,164 @@
...
+  tag: 'sage-tabs',
+  styleUrl: 'sage-tabs.scss',
+  shadow: true,
+  // scoped: true,
```
ju-Skinner: delete this comment if it's not needed.

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,164 @@
...
+    });
+    this.tabPanels.forEach(child => {
+      child['activeTab'] = this.activeTab.toString();
+      child['parentComponent'] = this.componentId.toString();
+      child['variant'] = this.variant.toString();
+    });
+  }
+
+  @Listen('keydown', {
+    passive: true,
```
ju-Skinner: Explain more about the use of `passive` here

libs/core/src/components/sage-tabs/stories/sage-tabs.stories.mdx
```diff
@@ -0,0 +1,133 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs';
```
ju-Skinner: Same things as the other two story files.

libs/core/src/components/sage-tabpanel/sage-tabpanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Sets the related tab name, this name must match a `sage-tab`'s tab property, required
+  */
+  @Prop() tab: string;
+
+  /**
+   * Keeps track of the activeTab, this property is passed in by parent component
+  */
+  @Prop({mutable: true}) activeTab: string;
+  /**
```
ju-Skinner, line 22: new line

libs/core/src/components/sage-tabpanel/sage-tabpanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+   * Keeps track of the parentComponent unique id, this property is passed in by parent component
+  */
+  @Prop({mutable: true}) parentComponent: string;
+  /**
```
ju-Skinner, line 28: new line

ju-Skinner (changes requested):
The rejection is around the Storybook files and a few coding styles.

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const tabList = Array.from(this.el.querySelectorAll('[role="tab"]'));
+    const activeEl = this.getActiveElement();
+    const tabLocations = this.getTabLocations(tabList);
+    const firstTabNumber = 0;
+    const lastTabNumber = tabLocations.length - 1;
+
+    if (ev.key === 'ArrowLeft') {
+      if (activeEl.id === tabLocations[0].id) {
+        tabLocations[lastTabNumber].focus();
+        this.activeTab = tabLocations[lastTabNumber].id;
```
QuintonJason: This may not be needed. The issue I'm raising here is, do we want the tab content to open while the keyboard user traverses the tab-list, then activate their choice by clicking space. Since don't know what kind of content will exist auto loading on focus could give a stuttered experience

libs/core/.eslintrc.json
goodwinchris: We can remove line 11

libs/core/src/components/sage-tab/sage-tab.scss
goodwinchris: https://kajabi.atlassian.net/browse/DSS-345

libs/core/.eslintrc.json
goodwinchris: Most methods can be private with no issue, however, there is a problem with linting not understanding that methods attached to a `@Listen` or `@Event` decorator will be invoked automatically when those decorators are called. 

When these methods are private, the linter is smart enough to "know" the method hasn't been explicitly invoked in that TSX file which his throws a warning about the method never being used. If the method is not private then the linter doesn't throw an error about the method never being used. 

I'm happy to solve this issue in a different way if you have any thoughts on a better approach. I do think its best for us to mark most of our methods as private.

libs/core/.stylelintrc.json
goodwinchris: Nesting in our CSS, [as seen here](https://github.com/Kajabi/sage/pull/15/files#diff-aef3889db0e6b395692b418b189dc651d500646a8d7ef82e7ee74b5f671e65c6R78-R97), is used to scope logic into a modifier class. I find this the easiest way to parse the code but if others have thoughts on this and would prefer to write it out all in one selector with less nesting I'd be open to that. Curious what the team thinks!

libs/core/src/components/sage-tab/sage-tab.scss
goodwinchris: I've opened a ticket for this work: https://kajabi.atlassian.net/browse/DSS-345

libs/core/src/components/sage-tab/sage-tab.tsx
goodwinchris: After reading through the stencil docs, I've found that a mutable prop allows stencil to know a prop will change over time without having to use a `@State` decorator which, from what I've ready, its recommend only for use when absolutely necessary. Because of this, I've opted for a mutable prop over state.

libs/core/src/components/sage-tab/sage-tab.tsx
goodwinchris: This is no longer needed. The `composed: false` is supposed to scope an event to the nearest shadow dom edge but in practice I found this doesn't work for our use case. I dug into this but came up empty handed. To solve this issue I've passed the parentCompomentID as well to make sure we can scope our listeners for these events instead.

libs/core/src/components/sage-tabs/sage-tabs.tsx
goodwinchris: I believe this is no longer needed, I was playing around with it when trying to scope `@Event` properly, removing now.

libs/core/src/components/sage-tabs/sage-tabs.tsx
goodwinchris: The [W3C best practices notes](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/):
>It is recommended that tabs activate automatically when they receive focus as long as their associated tab panels are displayed without noticeable latency. This typically requires tab panel content to be preloaded. Otherwise, automatic activation slows focus movement, which significantly hampers users' ability to navigate efficiently across the tab list.

I'd opt to defaulting to automatic switching and iterating the component if we ever come across an issue with noticeable latency. The component is built with the ability to pass all content directly into the tabpanel on load.

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Emits an event upon tab click for `sage-tab` and `sage-tabpanel` to listen for
+   */
+  @Event() tabClick: EventEmitter<object>;
+  private onTabClick(tab, parentComponent) {
+    this.activeTab = tab;
+    this.tabClick.emit([tab, parentComponent]);
+  }
+
+  private matchActiveTab() {
```
anechol: This function exists in multiple files. Is it possible to consolidate and share this between all necessary files without making it `private` in each subcomponent?

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,84 @@
...
+    const availabilityTabEdgeInlineStart = (
+      <span class="sage-tab-edge" role="presentation"></span>
+    )
+
+    const availabilityTabEdgeInlineEnd = (
+      <span class="sage-tab-edge sage-tab-edge--end" role="presentation"></span>
+    )
+
+    return (
+      <Host variant={this.variant} slot="tabs">
```
anechol: If we're using named slots, I would love to see these documented in Storybook as the props are.

pixelflips (changes requested):
Overall these are looking great. Just a couple of small issues I noticed. These may be Storybook issues, but let me know.

1. Could the items in the sidebar be capitalized to match other components?
Screenshot 2023-02-27 at 11 52 41 AM

2. When switching variants (canvas view) the tabs don't seem to load.
Screenshot 2023-02-27 at 11 52 54 AM

3. The links here (Tab & TabPanel) don't navigate to the subcomponents.
The URL does seem to change, but the UI does not.
Screenshot 2023-02-27 at 11 53 14 AM

libs/core/.stylelintrc.json
anechol: I think this warrants a broader discussion on how we want to organize our CSS. Because we have the `:host` selector to use, I imagine we could leverage this to eliminate the need for nesting or nesting too deeply.

libs/core/src/components/sage-tabs/stories/sage-tabs.docs.stories.mdx
```diff
@@ -0,0 +1,69 @@
...
+
+Handles tabpanel logic, requires a matching `tab` property to be passed
+
+## Accessibility
+
+This component has been built to align with the [W3C Aria best practices for tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/). You get all tabbing, keyboard navigation, focus states, and aria information applied automatically.
+
+In order for all accessibility to properly compile, you must provide a few required properties to the component. 
+
+**1. A well named `tablist-label`:** Provides context to screen readers about what kind of tablist information is being presented
```
anechol: Extreme nitpick but "well-named"

anechol (commented):
Overall, really good work! Some comments below and a more overarching question: do we find it necessary to have the subcomponents that make up tabs exist in the sidebar? Would this be too confusing for engineers?

One more thing: the spacing between the tab and the tabpanel is tight. Is there direction on how we should handle this spacing?

Screen Shot 2023-02-27 at 3 26 56 PM

goodwinchris:
> 



> Overall, really good work! Some comments below and a more overarching question: do we find it necessary to have the subcomponents that make up tabs exist in the sidebar? Would this be too confusing for engineers?
> 
> One more thing: the spacing between the tab and the tabpanel is tight. Is there direction on how we should handle this spacing?
> 
> Screen Shot 2023-02-27 at 3 26 56 PM

**Parent and Child Components:** When Julian and I paired on this approach we thought the multiple components wouldn't be too confusing for engineers, I'm open to revisiting this conversation though. For what its worth, this approach is aligned most tab systems I found and I believe it helps developers to build a mental map of their structure and expected content.

**Spacing between the Tab and the Tabpanel:** I checked in with Cam, he thinks for now we should leave no spacing between the tabs and tabpanels. He noted in the future he may want to add spacing but for now it seems best to not be prescriptive about default spacing and instead allow users to add what they need.

libs/core/src/components/sage-tabpanel/readme.md
```diff
@@ -0,0 +1,19 @@
...
+
+
+<!-- Auto Generated Below -->
+
+
+## Properties
+
+| Property          | Attribute          | Description                                                                                  | Type     | Default     |
+| ----------------- | ------------------ | -------------------------------------------------------------------------------------------- | -------- | ----------- |
+| `activeTab`       | `active-tab`       | Keeps track of the activeTab, this property is passed in by parent component                 | `string` | `undefined` |
```
pixelflips: I understand this (activeTab) gets added automatically, but when looking at the tabPanel page it seems like this is a prop that can be passed in, which was confusing to me.

May be worth checking with the team to see if this is desired behavior or how we might want to handle these cases to help eliminate any confusion.

anechol:
> > 
> 
> > Overall, really good work! Some comments below and a more overarching question: do we find it necessary to have the subcomponents that make up tabs exist in the sidebar? Would this be too confusing for engineers?
> > One more thing: the spacing between the tab and the tabpanel is tight. Is there direction on how we should handle this spacing?
> > Screen Shot 2023-02-27 at 3 26 56 PM
> 
> **Parent and Child Components:** When Julian and I paired on this approach we thought the multiple components wouldn't be too confusing for engineers, I'm open to revisiting this conversation though. For what its worth, this approach is aligned most tab systems I found and I believe it helps developers to build a mental map of their structure and expected content.
> 
> **Spacing between the Tab and the Tabpanel:** I checked in with Cam, he thinks for now we should leave no spacing between the tabs and tabpanels. He noted in the future he may want to add spacing but for now it seems best to not be prescriptive about default spacing and instead allow users to add what they need.

All good! In the case of spacing, if that's our approach going forward, would it make sense to create a specific custom prop that engineers can use to hook into to change that particular spacing as needed? Something similar to the Input component?

libs/core/src/components/sage-tab/sage-tab.tsx
goodwinchris: I love this idea of documenting our slots in some kind of argsTable like way. Do you know of a way to do this?

In this case, the user isn't expected to pass slot information into their `sage-tab`s or `sage-tabpanel`s but I'd still be happy to document them if we feel it'd be helpful!

libs/core/src/components/sage-tabs/stories/sage-tabs.docs.stories.mdx
goodwinchris: I'll update this to be "descriptive"

libs/core/src/components/sage-tab/sage-tab.tsx
anechol: That makes sense then for this case. We can document slots the same way we do for component props and custom props. Here's an [example](https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/accordion/accordion.tsx) in Ionic. It will automatically populate Storybook.

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,84 @@
+import { Component, Element, Host, h, Prop, Event, EventEmitter, State } from '@stencil/core';
+
+@Component({
+  tag: 'sage-tab',
+  styleUrl: 'sage-tab.scss',
+  shadow: false,
```
QuintonJason, line 6: Is it a best practice to explicitly set it to `false` or can this be omitted?

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,84 @@
...
+  */
+  @Prop({mutable: true}) activeTab: string;
+
+  /**
+   * Keeps track of the parentComponent unique id, this property is passed in by parent component
+  */
+  @Prop({mutable: true}) parentComponent: string;
+
+  /**
+   * Keeps track of if the expected tab variant, this property is passed in by parent component
```
QuintonJason: remove `if`

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,84 @@
...
+   * Emits an event upon tab click for `sage-tab` and `sage-tabpanel` to listen for
+   */
+  @Event() tabClick: EventEmitter<object>;
+  private onTabClick(tab, parentComponent) {
+    this.activeTab = tab;
+    this.tabClick.emit([tab, parentComponent]);
+  }
+
+  private matchActiveTab() {
+    if (this.activeTab && this.activeTab === this.tab) {
```
QuintonJason: Regarding `this.activeTab && this.activeTab` is that part of the conditional needed as it seems like it can go? Would it work the same as `this.activeTab === this.tab`?

libs/core/src/components/sage-tab/sage-tab.tsx
QuintonJason: +1 for documenting the slot name and description

libs/core/src/components/sage-tabs/test/sage-tabs.e2e.ts
```diff
@@ -0,0 +1,237 @@
...
+        <sage-tabpanel tab="two">Content 2</sage-tabpanel>
+      </sage-tabs>
+  `);
+    const element = await page.find(`sage-tabs`);
+    expect(element).toHaveClass('sage-tabs--availability');
+    const tabEdges = await page.find(`.sage-tab-edge`);
+    expect(tabEdges).not.toBeNull();
+  });
+
+  it('renders availability filter', async () => {
```
QuintonJason: filter variant

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,161 @@
...
+  @Prop() tablistLabel: string;
+ 
+  /**
+    * Sets unique id on tabs component, required
+  */
+  @Prop() componentId: string;
+  
+  /**
+   * Sets tabs variant styles as outlined in Figma documentation, optional
+   * @defaultValue primary
```
QuintonJason: Can this be optional with a default value? Not sure if a conflict with occur

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,161 @@
...
+    */
+  @Prop() variant?: 'primary' | 'availability' | 'filter' = 'primary';
+
+  /**
+   * Sets default active tab, required
+   */
```
QuintonJason, line 40: Default and active together are feel a little weird since it changes based on user feedback

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,161 @@
...
+    for (let i = 0; i < tabList.length; i += 1) {
+      const tab = tabList[i];
+
+      tabs.push(tab);
+    }
+    return(tabs);
+  }
+
+  // Copied code - might be overkill or ideal to make global?
+  private getActiveElement(root: Document | ShadowRoot = document): Element | null {
```
QuintonJason: Why do you feel this might be overkill?

QuintonJason:
I don't think we need the story file for the `Tab` or `TabPanel`. They only contain one item for `tabpanel` and `tabcontent`. In the examples in the `Tabs`, the only thing that seems to change was the variants. Probably need a consensus, but I'd be in favor of removing the first 2 and putting refs to their `ArgsTable` if need be. Thoughts?

QuintonJason (changes requested):
Looking and working good! Left some comments, let me know you need any additional clarity.

libs/core/src/components/sage-tab/sage-tab.tsx
goodwinchris: For the subcomponents (`sage-tab` and `sage-tabpanel`), leaving the shadow dom open allows them to communicate with their parent (`sage-tabs`) while staying scoped to their parent shadowdom

libs/core/src/components/sage-tabs/sage-tabs.tsx
goodwinchris: Originally I thought this logic may be more than we needed as it checks for more than just shadowroot active elements but now that I've been using it I think its actually working well.

goodwinchris:
> I don't think we need the story file for the Tab or TabPanel. They only contain one item for tabpanel and tabcontent. In the examples in the Tabs, the only thing that seems to change was the variants. Probably need a consensus, but I'd be in favor of removing the first 2 and putting refs to their ArgsTable if need be. Thoughts?

I don't have any issue consolidating further though I also don't mind having them called out explicitly, open to what others prefer

libs/core/src/components/sage-tabpanel/sage-tabpanel.tsx
```diff
@@ -0,0 +1,48 @@
...
+   * Keeps track of the tabpanel selected state, this property is passed by parent component 
+  */
+  /** @internal */
+  @Prop({mutable: true}) selected = false;
+
+  render() {
+    return (
+      <Host slot="tabpanels">
+        <div 
+          role="tabpanel"
```
teenwolfblitzer, line 38: Since we're aligning to the W3C implementation, should each panel also be set to `tabindex="0"`?

goodwinchris: Good callout, I'll update!

libs/core/src/components/sage-tabpanel/sage-tabpanel.tsx
```diff
@@ -0,0 +1,49 @@
...
+          id={this.tab + '-panel'}
+          tabindex="0"
```
teenwolfblitzer, line 40: ❤️

teenwolfblitzer (approved):
👍 LGTM, working nicely! Really appreciate the detail with kb accessibility.

pixelflips (approved):
👍🏼 LGTM! Thanks for fixing up the issues I pointed out! Great work!

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,70 @@
...
+   * Keeps track of the parentComponent unique id, this property is passed by parent component
+  */
+  /** @internal */
+  @Prop({mutable: true}) parentComponent: string;
+
+  /**
+   * Keeps track of if the expected tab variant, this property is passed by parent component
+  */
+   /** @internal */
+  @Prop() variant: string;
```
QuintonJason, line 26: How does `@internal` work in this scenario? I ask because it looks like the `variant` is ultimately defined here even though it gets passed through to the parent. Mainly trying to get more of an understanding since I would initially expect to see the corresponding test in the tab test. Being that the prop doesn't show on the `Docs` page, it them seems necessary to have the test in the tabs test file. Is this way preferred for items that contain subcomponents?

goodwinchris: I did some digging on the Ionic repo to see how they were handling props that weren't intended to be exposed to the people using components. They use `@internal` for that. I haven't been able to find good documentation on it but it doesn't seem to change any of the innerworkings of the components while hiding the props from the storybook args table automatically. 

In this usecase, that is exactly what we want. We want stencil to treat the `Prop()` decorator exactly the same but we don't want to expose Props to developers and confuse them. This allows us to only expose the props we expect developers to change and interact with. As you can't have just 1 tab with a different variant, with `Sage-Tab` and `Sage-Tabpanel` the variant is passed down from the parent intentionally. Exposing it to developers only adds confusion to our docs.

anechol: I think this is what they're referencing: https://tsdoc.org/pages/tags/internal/

goodwinchris: Good find! Heres an example form the Ionic repo: https://github.com/ionic-team/ionic-framework/blob/545f0a9309bb453a78708bb382124ab285c041a2/core/src/components/router-outlet/route-outlet.tsx

libs/core/src/components.d.ts
```diff
@@ -149,11 +149,53 @@ export namespace Components {
...
+        "selected": boolean;
+        /**
+          * Sets the related tab name, this name must match a `sage-tab`'s tab property
+         */
+        "tab": string;
+        "variant": string;
+    }
+    interface SageTabs {
+        /**
+          * Sets starting active tab and maintains active tab as component re-renders
```
ju-Skinner: Since this is a string, I would change the description a bit.

`Sets the starting active tab name and maintains the name as the component re-renders`

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,76 @@
...
+  styleUrl: 'sage-tab.scss',
+  shadow: false,
+})
+export class SageTab {
+  @Element() el: HTMLElement;
+
+  /**
+   * Sets the related tab name, this name must match a `sage-tabpanel`'s tab property
+  */
+  @Prop() tab!: string;
```
ju-Skinner: Here in the description you mention "related tab name", I think it would be best to rename this prop to `name`. It would be redundant to name the property `tabName`.

libs/core/src/components/sage-tab/sage-tab.tsx
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Keeps track of if the expected tab variant, this property is passed by parent component
+  */
+  /** @internal */
+  @Prop() variant: string;
+
+  /**
+   * Keeps track of if the expected tab variant, this property is passed by parent component
+  */
+  /** @internal */
```
ju-Skinner: Looks like you need to update the description of the prop here. Probably from a copy and paste.

libs/core/src/components/sage-tabpanel/sage-tabpanel.tsx
```diff
@@ -0,0 +1,49 @@
...
+  styleUrl: 'sage-tabpanel.scss',
+  shadow: false,
+})
+export class SageTabpanel {
+  @Element() el: HTMLElement;
+
+  /**
+   * Sets the related tab name, this name must match a `sage-tab`'s tab property
+  */
+  @Prop() tab!: string;
```
ju-Skinner: possibly a rename to `tabName` or `tabname` if we don't want to deal with the hyphen at the element attribute level.

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,132 @@
...
+  @Prop() componentId!: string;
+  
+  /**
+   * Sets tabs variant styles as outlined in Figma documentation
+  */
+  @Prop() variant!: 'primary' | 'availability' | 'filter';
+
+  /**
+   * Sets starting active tab and maintains active tab as component re-renders
+   */
```
ju-Skinner, line 40: Since this is a string, I would change the description a bit.

`Sets the starting active tab name and maintains the name as the component re-renders`

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,132 @@
+import { Component, Element, Host, h, Prop, Listen } from '@stencil/core';
+import { move } from 'fs-extra';
```
ju-Skinner: I don't think this needs to be here.

ju-Skinner (changes requested):
The requested changes are due to the import in `libs/core/src/components/sage-tabs/sage-tabs.tsx` that shouldn't be there `import { move } from 'fs-extra'`. 

There are additional comments left as well.

libs/core/src/components/sage-tabpanel/sage-tabpanel.tsx
goodwinchris: I've updated it to `name`, that should work fine and is in alignment with HTML attribute naming.

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -0,0 +1,131 @@
...
+   * Sets tabs variant styles as outlined in Figma documentation
+  */
+  @Prop() variant!: 'primary' | 'availability' | 'filter';
+
+  /**
+   * Sets the starting active tab name and maintains the name as the component re-renders
+   */
+  @Prop({mutable: true}) activeTabName!: string;
+  
+  @Prop({mutable: true}) activeTabIndex!: number;
```
ju-Skinner: we need to add JSDoc here, I'd suggest just making this `@internal` as well.

QuintonJason:
tabs
@goodwinchris  Whenever you tab into the tablist, the `left` and `right` arrow keys throw an error. If you click the tab, you can use the `left` and `right` keys with no problems

ju-Skinner:
> tabs [ tabs ](https://user-images.githubusercontent.com/1241836/223537082-301a24df-9d28-4561-a533-9bc4c5008ad3.gif) [ ](https://user-images.githubusercontent.com/1241836/223537082-301a24df-9d28-4561-a533-9bc4c5008ad3.gif) @goodwinchris Whenever you tab into the tablist, the `left` and `right` arrow keys throw an error. If you click the tab, you can use the `left` and `right` keys with no problems

@QuintonJason, @goodwinchris, and I have fixed this in this [commit](https://github.com/Kajabi/sage/pull/15/commits/7e1586ec8998e6842fdf42b9cf4ef20e72973656)

QuintonJason (approved):
Looks and works great!

anechol (approved):
Lots of awesome work here!

ju-Skinner (approved):
Great job...Thanks for pushing through this.

PR #27, opened 2023-02-23
[CHORE] - Dependency Updates

ju-Skinner:
# Description

It was noticed that Lerna and NX both had some nice updates to their tooling. So it felt appropriate to update them.

Lerna v6.5 updates - [here](https://blog.nrwl.io/whats-new-with-lerna-6-5-4a4f69371504)
NX Updates - [here](https://blog.nrwl.io/nx-15-4-vite-4-support-a-new-nx-watch-command-and-more-77cbf6c9a711)


teenwolfblitzer (approved):
👍🏼 no concerns here, scripts are running well. Interesting to see that the react dependencies were downgraded though.

ju-Skinner:
> 👍🏼 no concerns here, scripts are running well. Interesting to see that the react dependencies were downgraded though.

@monkeypox8 There have been some talks about how React 18 has been having issues. We had a `peerDependency` on 18 prior to this, which would mean the containing app would need React 18. So I elected to downgrade to V17 until 18 appears to be more stable.

QuintonJason (approved):
Definitely looking forward to how we utilize the new features

ju-Skinner:
> Definitely looking forward to how we utilize the new features

@QuintonJason Lerna updates will be utilized in the build/release process.  NX updates, specifically the `nx watch` for watching files for the various targets. e.g linting, tests

PR #31, opened 2023-03-09
[DSS-43]: Hotfix - Scope key input active tab changes to components which matching parentId's

goodwinchris:
# Description
On pages where multiple `sage-tabs` are on the page, we were accidentally changing the active tab of each component even if that wasn't the component the keydown was triggered from.

[DSS-43: More complex Web Component](https://kajabi.atlassian.net/browse/DSS-43)

before | after
--|--
CleanShot 2023-03-09 at 16 40 52|CleanShot 2023-03-09 at 16 42 13

# Checklist:


ju-Skinner (approved):
LGTM but I'm biased 😉

libs/core/src/components/sage-tabs/sage-tabs.tsx
```diff
@@ -51,15 +51,16 @@ export class SageTabs {
     }
   }
 
-  @Listen('keydown', {passive: true, target: 'window'})
+  @Listen('keydown', {})
   handleKeyDown(ev: KeyboardEvent) {
     const keySet = ["ArrowLeft", "ArrowRight", "Home", "End"];
```
teenwolfblitzer, line 56: This is more out of curiosity than an actual issue: on my machine, `HOME` and `END` also affect the window scroll in addition to the tabs. Might be due to Storybook's app frame/iframe, though

teenwolfblitzer (approved):
👍 Works great, no concerns here

libs/core/src/components/sage-tabs/sage-tabs.tsx
goodwinchris: Thats a good callout! I'll add a `preventDefault()` to our logic.

goodwinchris: @monkeypox8: https://github.com/Kajabi/sage/pull/31/commits/2f5223b42aeba82579bd0912e5649fd3e1106ff2

QuintonJason (approved):
Good stuff

PR #39, opened 2023-03-17
[DSS-354] Tokens - Add Sage Tokens to Repo

anechol:
# Description

Adds an initial set of Sage tokens to the repo. Further formatting and fine-tuning of names will be done in a future PR. 

[DSS-354](https://kajabi.atlassian.net/browse/DSS-354)

[DSS-354]: https://kajabi.atlassian.net/browse/DSS-354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/src/global/styles/tokens/tokens.json
```diff
@@ -0,0 +1,437 @@
+{
+  "sage": {
+    "color": {
+      "white": {
+        "white": {
+          "value": "#ffffff",
```
pixelflips: Seems all the other color values are uppercase, should this one be as well?

libs/core/src/global/styles/tokens/tokens.json
```diff
@@ -0,0 +1,437 @@
...
+          "type": "color",
+          "description": "Backgrounds, primary/destructive CTA text"
+        }
+      },
+      "black": {
+        "black": {
+          "value": "#000000",
```
pixelflips: No preference really, but should we use the shorthand hex notation where possible?

pixelflips (approved):
A couple of tiny non-blocking comments, but LGTM! Nice work! 👍🏼

ju-Skinner (approved):
LGTM...great initial commit to getting us on our way with Tokens.

libs/core/src/global/styles/tokens/tokens.json
anechol: Good callout, I will set them all to uppercase.

libs/core/src/global/styles/tokens/tokens.json
anechol: I've updated this for now. Previously, I had issues with linting rules, but it doesn't seem to be a problem at the moment.

goodwinchris (approved):

PR #42, opened 2023-03-30
Tokens - Adds Line Height, Letter Spacing, and Border Radius

anechol:
Adds missing tokens for line height, letter spacing, and border radius. Also adds design system name as prefix in build script.


libs/core/src/global/styles/tokens/_font-size.scss
```diff
@@ -0,0 +1,11 @@
+:root {
+  --sage-font-size-h1: 40px;
```
pixelflips, line 2: Is the plan to move from rem to px or is that not yet decided?

pixelflips (approved):
One small non-blocking comment. LGTM!

goodwinchris (approved):

ju-Skinner (approved):
Great work!!!!

PR #45, opened 2023-04-04
[DSS-379]: Divider

pixelflips:
# Description

Adds Divider component to the system including:

- Horizontal default
- Vertical variant
- Optional offset to create a full-bleed or full-height appearance.

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- os: macOS 13.3

# Checklist:

### DIFF
--- a/libs/core/src/components/sage-divider/sage-divider.scss
+++ b/libs/core/src/components/sage-divider/sage-divider.scss
@@ -1,20 +1,19 @@
-/*
-TODO: update offset values to use
-our spacing tokens when available.
-*/
+:host {
+  --divider-background-color: var(--sage-color-neutral-grey-400);
+}
 
 $sage-divider-offsets: (
-  xxs: 4px,
-  xs: 8px,
-  sm: 16px,
-  md: 24px,
-  lg: 32px,
-  xl: 48px,
-  xxl: 64px,
+  xxs: #{var(--sage-spacing-xxs)},
+  xs: #{var(--sage-spacing-xs)},
+  sm: #{var(--sage-spacing-sm)},
+  md: #{var(--sage-spacing-md)},
+  lg: #{var(--sage-spacing-lg)},
+  xl: #{var(--sage-spacing-xl)},
+  xxl: #{var(--sage-spacing-xxl)},
 );
 
 .sage-divider {
-  background: #D3D5D9;
+  background: var(--divider-background-color);
   border: 0;
   display: flex;
   height: 1px;
@@ -29,11 +28,11 @@ $sage-divider-offsets: (
 
 @each $offset, $value in $sage-divider-offsets {
   .sage-divider--offset-#{$offset} {
-    margin: 0 (-$value);
+    margin: 0 calc($value * -1);
     padding: 0 $value;
 
     &.sage-divider--vertical {
-      margin: (-$value) 0;
+      margin: calc($value * -1) 0;
       padding: $value 0;
     }
   }
--- a/libs/core/src/components/sage-divider/sage-divider.scss
+++ b/libs/core/src/components/sage-divider/sage-divider.scss
@@ -20,11 +20,11 @@ $sage-divider-offsets: (
   height: 1px;
   margin: 0;
   width: 100%;
+}
 
-  &--vertical {
-    height: 100%;
-    width: 1px;
-  }
+.sage-divider--vertical {
+  height: 100%;
+  width: 1px;
 }
 
 @each $offset, $value in $sage-divider-offsets {
--- a/libs/core/src/components/sage-divider/sage-divider.scss
+++ b/libs/core/src/components/sage-divider/sage-divider.scss
@@ -1,0 +1,44 @@
+$sage-divider-offsets: (
+  xxs: 4px,
+  xs: 8px,
+  sm: 16px,
+  md: 24px,
+  lg: 32px,
+  xl: 48px,
+  xxl: 64px,
+);
+
+.sage-divider {
+  background: #D3D5D9;
+  border: 0;
+  display: flex;
+  height: 1px;
+  margin: 0;
+  width: 100%;
+
+  &--vertical {
+    height: 100%;
+    width: 1px;
+  }
+}
+
+@each $key,
+$value in $sage-divider-offsets {
+  $offset: "#{$key}";
+
+  .sage-divider--offset-#{$offset} {
+    margin: 0 (-$value);
+    padding: 0 $value;
+
+    &.sage-divider--vertical {
+      margin: (-$value) 0;
+      padding: $value 0;
+    }
+  }
+}
+
+@media (max-width: 767px) {
+  .sage-divider--vertical {
+    display: none;
+  }
+}
--- a/libs/core/src/components/sage-divider/sage-divider.scss
+++ b/libs/core/src/components/sage-divider/sage-divider.scss
@@ -1,25 +1,25 @@
 :host {
-  --divider-background-color: var(--sage-color-neutral-grey-400);
-  --divider-offset-xxs: var(--sage-spacing-xxs);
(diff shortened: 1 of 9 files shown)


libs/core/src/components/sage-divider/sage-divider.scss
```diff
@@ -0,0 +1,48 @@
...
+  &--vertical {
+    height: 100%;
+    width: 1px;
+  }
+}
+
+@each $key, $value in $sage-divider-offsets {
+  $offset: "#{$key}";
+
+  .sage-divider--offset-#{$offset} {
```
ju-Skinner: This seems to be an unnecessary layer of abstraction. Mainly due to just reassigning the variable name `key` to `offset`, when we can just do this

```
@each $offset, $value in $sage-divider-offsets
```

libs/core/src/components/sage-divider/stories/sage-divider.docs.stories.mdx
```diff
@@ -0,0 +1,56 @@
...
+### Offsets
+
+Both horizontal and vertical dividers can be offset to create a full width/height bleed appearance.
+
+Offset sizes follow the global spacing sizes: `xxs`:4px, `xs`:8px, `sm`:12px,`md`:24px, `lg`:32px, `xl`:48px, `xxl`:64px.
+
+<Canvas>
+  <sage-divider offset="lg" />
+</Canvas>
+
```
ju-Skinner, line 51: Maybe add a short statement describing what this is below. I was a bit confused when I was looking at it. It wasn't until I clicked `show code`, I had the aha moment.

e.g Below you will see a vertical divider using a large offset.

ju-Skinner (changes requested):
Left a couple of comments. 

* Comment regarding @each is the blocker.
* Storybook docs comment is more of a suggestion. 

Outside of those two, everything looks good. Great Job

libs/core/src/components/sage-divider/stories/sage-divider.docs.stories.mdx
pixelflips: @ju-Skinner I added a callout/description right below the "Offset" heading explaining what offset is and mentioned both vertical and horizontal. 

Happy to break them up, but it seems a like a bit of overkill to have two separate sentences for similar examples. Do we feel more copy is better?

ju-Skinner: Let me re-iterate "Storybook docs comment is more of a **suggestion**".

My comment was directed to lines 52-56. Sorry if that may have caused confusion. 

When looking at the docs page. I see the two canvases. The first canvas feels related to what you have listed above. The second however seems like it's just there. My suggestion was to add a short sentence to behave like a figure caption. 

image

pixelflips: No confusion, I just felt like the "Both horizontal and vertical dividers can be offset..." sentence being right above the two examples reduced the amount of copy to read. 

All good and added the suggested copy for clarity.

libs/core/src/components/sage-divider/sage-divider.scss
```diff
@@ -0,0 +1,46 @@
...
+
+.sage-divider {
+  background: #D3D5D9;
+  border: 0;
+  display: flex;
+  height: 1px;
+  margin: 0;
+  width: 100%;
+
+  &--vertical {
```
goodwinchris: I'm not the biggest fan of concatenating class names when it can be easily avoided as it makes searching/find and replacing more challenging. I wanted to voice it but ultimately its your call here 👍

libs/core/src/components/sage-divider/sage-divider.tsx
```diff
@@ -0,0 +1,44 @@
...
+  tag: 'sage-divider',
+  styleUrl: 'sage-divider.scss',
+  shadow: true,
+})
+export class SageDivider {
+  /**
+   * Sets divider to display vertically
+   * @defaultValue false
+   */
+  @Prop() vertical = false;
```
goodwinchris: Does it make sense to have this prop be named `direction` or `variant` and pass a default value of `horizontal`?

libs/core/src/components/sage-divider/sage-divider.scss
pixelflips: Updated.

libs/core/src/components/sage-divider/sage-divider.tsx
pixelflips: I don't understand the benefit or difference here. Also, seems the logic is much simpler when using use a boolean and just one class, but maybe I'm missing something.

goodwinchris: Yep, could be fine either way. In my mind, having a divider be horizontal by default feels like it makes the vertical a secondary option where as defining `direction` makes each equal.

Definitely not a thing that you need to change if you don't like that solution.

anechol: Another topic we can discuss further in our upcoming meeting, but I think the intention here was to better align with our definition of a `variant` for consistency.

libs/core/src/components/sage-divider/sage-divider.scss
```diff
@@ -0,0 +1,46 @@
+/*
+TODO: update offset values to use
+our spacing tokens when available.
+*/
+
+$sage-divider-offsets: (
+  xxs: 4px,
```
anechol: Let's use our tokens here

libs/core/src/components/sage-divider/stories/sage-divider.docs.stories.mdx
```diff
@@ -0,0 +1,58 @@
...
+## Properties
+
+<ArgsTable of="sage-divider" />
+
+### Default
+
+The default divider displays a horizontal line.
+
+<Story story={stories.Default} />
```
anechol, line 22: I might be missing (bad eyesight), but I currently cannot see the dividers at all.
Screen Shot 2023-04-21 at 4 28 17 PM

libs/core/src/components/sage-divider/sage-divider.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   *
+   * Adds offset margin/padding to expand the width (horizontal) or the height (vertical) of divider.
+   */
+  @Prop() offset: 'xxs' | 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'xxl';
+
+  private classNames() {
```
anechol, line 20: We'll discuss this in more detail in the future, but I wonder if this should evolve into a global utility function.

anechol (changes requested):
Great work on this! A few thing below...

libs/core/src/components/sage-divider/stories/sage-divider.docs.stories.mdx
pixelflips: @anechol Not at all. Looks like something is def off. I just double-checked and I do indeed see dividers. Sometimes when I switch branches I get odd behaviors. Maybe try restarting?

Screenshot 2023-04-21 at 2 43 14 PM

anechol: Yep that was def the issue, thanks for reminding me

libs/core/src/components/sage-divider/sage-divider.scss
pixelflips: Done.

anechol (approved):
LGTM

ju-Skinner (approved):
Great job...Thanks for going back and refactoring based on our teams' meeting yesterday.

PR #47, opened 2023-04-06
[DSS-382] Checkbox

pixelflips:
# Description

Adds Checkbox component to the system including:

- Default variant
- Invalid variant
- Indeterminate variant
- Checked/unchecked states
- Optional messaging

Fixes [DSS-382: Checkbox](https://kajabi.atlassian.net/browse/DSS-382)

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [x] tested manually
- [x] other:

**Test Configuration**:

- os: macOS 13.3

# Checklist:


libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,127 @@
...
+.sage-checkbox--indeterminate {
+  input:checked {
+    &::after {
+      border-bottom: 0;
+      transform: rotate(87deg) translate3d(-114%, 25%, 0);
+    }
+  }
+}
+
+.sage-checkbox--invalid {
```
goodwinchris: Does it make sense to append classes or would it be simpler to to build a selector for the state of the checkbox? 

e.g. `input:invalid`

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    if (this.indeterminate) {
+      const indeterminateClassName = 'sage-checkbox--indeterminate';
+      className += ' ' + indeterminateClassName;
+    }
+
+    if (this.disabled) {
+      const disabledClassName = 'sage-checkbox--disabled';
+      className += ' ' + disabledClassName;
+    }
```
goodwinchris: Similar question to what I asked in the scss file - do we need modifier classes to target these states? If we do want to do it with classes should we explore using state classes? 

e.g. `is-checked`, `is-invalid`, etc

pixelflips: Based on our convo, I've updated these to use the state classes. Take another look when you have time.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -90,36 +120,6 @@ label {
...
-    }
-  }
-
-  label,
-  .sage-checkbox__message {
-    color: var(--invalid-color);
-  }
-}
-
 .sage-checkbox--disabled {
```
goodwinchris: Looks like we'll need to clean this one up too

pixelflips: Refactored and removed completely. 👍🏼

libs/core/src/components/sage-checkbox/test/sage-checkbox.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+  });
+  it('renders checked input when checked prop is set', async () => {
+    const { root } = await newSpecPage({
+      components: [SageCheckbox],
+      html: `<sage-checkbox checkbox-id="default" label="Label text" checked />`,
+    });
+    const input = root?.shadowRoot?.querySelector('input');
+    expect(input?.checked).toBe(true);
+  });
+  it('renders disabled input when checked prop is set', async () => {
```
goodwinchris: Typo here, should be "when disabled prop is set"

pixelflips: 🦅

libs/core/src/components/sage-checkbox/test/sage-checkbox.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const { root } = await newSpecPage({
+      components: [SageCheckbox],
+      html: `<sage-checkbox checkbox-id="default" label="Label text" disabled />`,
+    });
+    const input = root?.shadowRoot?.querySelector('input');
+    expect(input?.disabled).toBe(true);
+  });
+  it('renders with error when invalid prop is set', async () => {
```
goodwinchris: I don't see the error text in the expected HTML block below

pixelflips: this is probably just a bad description as it's not supposed to display any message text. let me try and reword.

libs/core/src/components/sage-checkbox/test/sage-checkbox.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+        <mock:shadow-root>
+          <div class="sage-checkbox is-invalid">
+            <input type="checkbox" id="default">
+            <label htmlfor="default">Label text</label>
+          </div>
+        </mock:shadow-root>
+      </sage-checkbox>
+    `);
+  });
+  it('renders indeterminate when indeterminate prop is set', async () => {
```
goodwinchris: I'd love if we could nest some checkboxes to confirm indeterminate state applies as expected.

libs/core/src/components/sage-checkbox/test/sage-checkbox.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+      </sage-checkbox>
+    `);
+  });
+  it('renders required input when required prop is set', async () => {
+    const { root } = await newSpecPage({
+      components: [SageCheckbox],
+      html: `<sage-checkbox checkbox-id="default" label="Label text" required />`,
+    });
+    const input = root?.shadowRoot?.querySelector('input');
+    expect(input?.required).toBe(true);
```
goodwinchris, line 129: I have no issue with this approach but I'm curious why you chose to check it with `input?.required` instead of checking against the `page.root` HTML.

libs/core/src/components/sage-checkbox/test/sage-checkbox.e2e.ts
```diff
@@ -0,0 +1,126 @@
+import { newE2EPage } from '@stencil/core/testing';
```
goodwinchris, line 1: All these tests look good to me but can we add some for user flows as well? Can we test when a user clicks on a checkbox and confirm it sets as checked or inverse. Can we confirm invalid states when required show error messages, etc?

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  &:focus-visible {
+    box-shadow: 0 0 0 4px var(--focus-color);
+    outline: none;
+  }
+}
+
+label {
```
goodwinchris, line 115: Might make sense to update to `input + label` - your call

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,106 @@
+import { Component, h, Prop, Host } from '@stencil/core';
+
+@Component({
+  tag: 'sage-checkbox',
+  styleUrl: 'sage-checkbox.scss',
+  shadow: true,
+})
+export class SageCheckbox {
+  /**
+   * Whether or not the checkbox is checked.
```
goodwinchris: Seeing inconsistencies here with some descriptions ending with a period and some not, can we unify?

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,106 @@
...
+  @Prop() message: string;
+
+  /**
+   * String used for checkbox `name` attribute.
+   */
+  @Prop() name: string;
+
+  /**
+   * Whether or not checkbox should display as indeterminate. Prop is for visual styling only.
+   * Only JavaScript can set the objects `indeterminate` property. See [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes)
```
goodwinchris: Could be  a good link to add at the end of the story in an `Additional Resources` section

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Whether or not the checkbox is disabled.
+   */
+  @Prop() disabled: boolean;
+
+  /**
+   * Whether or not the checkbox is invalid.
+   */
+  @Prop() invalid: boolean;
```
goodwinchris, line 22: Text input and textarea have events emit when inputs are changed - could be something to add

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,106 @@
...
+            type="checkbox"
+            id={this.checkboxId}
+            name={this.name}
+            value={this.value}
+            checked={this.checked}
+            required={this.required}
+            disabled={this.disabled}
+          />
+          <label htmlFor={this.checkboxId}>{this.label}</label>
+          {message}
```
goodwinchris: I think this could be consolidated to one line: 
`{this.message && <div class={'sage-checkbox__message'}>{this.message}</div>}`

libs/core/src/components/sage-checkbox/test/sage-checkbox.spec.tsx
pixelflips: no rhyme or reason really. I believe I saw it used in another component, but felt to me it was better to be able to check the input vs. just rendering out the markup.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
pixelflips: Genuinely curious about this one. If the styles are scoped to the component, why do you feel the adjacent combinator would be needed?

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
pixelflips: Gave this a shot, but let me know if you have any feedback.

goodwinchris: Looks mostly good to me, we'll need tests for it! Input and Textarea and my textarea PR have some examples. Let me know if you wanna pair on it

libs/core/src/components/sage-checkbox/sage-checkbox.scss
goodwinchris: Only reason I can see is if someone passed a label into a slot. Upon further investigation I don't see a slot so probably unnecessary although if you do want to be able to nest checkboxes within checkboxes we'll need a slot for that, right?

goodwinchris: nvm - checkboxes can't be nested, disregard.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,125 @@
...
+        transform: rotate(87deg) translate3d(-114%, 25%, 0);
+      }
+    }
+  }
+}
+
+input {
+  appearance: none;
+  border: 1px solid var(--border-color);
+  border-radius: 4px;
```
QuintonJason: Can `border-radius` be updated to be a var since we'll have a token for that?

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,125 @@
...
+  height: 16px;
```
QuintonJason: Can you put the repeated values of `height` and `width` in a var?

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,125 @@
...
+    background: var(--hover-background-color);
+    border-color: var(--hover-border-color);
+  }
+
+  &:checked {
+    background: var(--checked-color);
+    border-color: var(--checked-color);
+
+    &::after {
+      border: 1px solid #FFF;
```
QuintonJason: Update color to be a var

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * The value of the checkbox that is submitted with a form.
+   */
+  @Prop() value: string;
+
+  /**
+   * Emits a boolean indicating whether the checkbox is currently checked or unchecked.
+   */
+  @Event() sageCheckbox: EventEmitter<boolean>;
```
QuintonJason: Can the name of the event be updated to include the kind of event, something like `sageCheckboxChange`?

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const target = event.target as HTMLInputElement;
+    const isChecked = target.checked;
+
+    this.sageCheckbox.emit(isChecked);
+  }
+
+  private classNames() {
+    let className = `sage-checkbox`;
+
+    if (this.invalid) {
```
QuintonJason: It's probably fine to append classes this way as it lends itself more to the React side. I'm mine I did the following to achieve this behavior:
```
class={`
            sage-component
            ${this.prop ? (`sage-component--${this.prop}`) : ''}
            ${this.prop ? 'sage-prop' : ''}
          `}

What are your thoughts on this approach?

File: libs/core/src/components/sage-checkbox/stories/sage-checkbox.docs.stories.mdx [hunk 19]
```diff
@@ -0,0 +1,129 @@
...
+<Canvas>
+  <sage-checkbox checkbox-id="checked" label="Label" checked />
+</Canvas>
+
+
+### Indeterminate
+
+Component `indeterminate` prop is for visual styling only.
+
+Only JavaScript can set the objects `indeterminate` property. See [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes)
```
QuintonJason, line 50: 🥇

libs/core/src/components/sage-checkbox/test/sage-checkbox.e2e.ts
```diff
@@ -0,0 +1,148 @@
...
+
+describe('sage-checkbox', () => {
+  it('renders', async () => {
```
QuintonJason: These tests are specs in that they are testing whether are not something is there. You already have those in the `spec` test. Those are the ones giving you the 100 percent coverage. These are more like toggling state. They mostly involve some type of event occurring. 

One example can be to switch from not checked -> indeterminate -> checked and check stuff along the way or flows similar to that in your component.

QuintonJason (changes requested):
This is great work. Nice docs as well. Let comments for action items and considerations.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
pixelflips: Done

libs/core/src/components/sage-checkbox/sage-checkbox.scss
pixelflips: Done

libs/core/src/components/sage-checkbox/sage-checkbox.scss
pixelflips: Done

libs/core/src/components/sage-checkbox/sage-checkbox.scss
pixelflips: Done

libs/core/src/components/sage-checkbox/test/sage-checkbox.e2e.ts
pixelflips: @QuintonJason I'd love to chat more on this one. Please ping me when you have some time.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
pixelflips: I think I based this on a similar solution in the button component. I think the `markup`/`return` is easier to read, but no real preference here so not 100% either way.

libs/core/src/components/sage-checkbox/test/sage-checkbox.spec.tsx
goodwinchris: My bad, this isn't a thing. Disregard.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,128 @@
+:host {
+  --border-color: #D3D5D9;
```
goodwinchris: I know most of these seem self explanatory but we can add @prop info here for storybook - [see input](https://github.com/Kajabi/sage/blob/next/libs/core/src/components/sage-input/sage-input.scss)

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * String used for label text next to checkbox.
+   */
+  @Prop() label: string;
+
+  /**
+   * String used for message below checkbox.
+   */
+  @Prop() message: string;
```
goodwinchris: Should this be `hintMessage`?

pixelflips: @goodwinchris I took another look at the Figma spec and design refers to this as supporting text. Any thoughts on whether I should stick with hint or supporting? `supportingMessage` does seem a bit lengthy, right?

goodwinchris: I don't mind being verbose if it matches. Doesn't feel make or break to me but maybe worth confirming with design if we see it as a hint message in an attempt to keep the naming aligned?

QuintonJason (changes requested):
The code side looks good to go. I overlook one thing in the docs. For `Disabled`, `Message`, and `Invalid` sections, you don't need all the permutations of the Checkbox. You'll also need a short description for `Disabled` and `Invalid`.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
pixelflips: Based on convo with design I have updated this to `helperMessage`. This seems to be the naming they'd like to use across form elements.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
pixelflips: Going to hold off on this for now. I can see some being useful and some not needed. This should probably be a team discussion on how we'd like to handle these.

pixelflips:
> The code side looks good to go. I overlook one thing in the docs. For `Disabled`, `Message`, and `Invalid` sections, you don't need all the permutations of the Checkbox. You'll also need a short description for `Disabled` and `Invalid`.

@QuintonJason good call. cleaned these up and added desc as suggested. Take a look when you can. Def much cleaner IMO.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
ju-Skinner: If aren't allowing the customer/devs to tweak these then they don't need to have TSDoc comments.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,117 @@
+import { Component, h, Prop, Host, Event, EventEmitter } from '@stencil/core';
+
+@Component({
+  tag: 'sage-checkbox',
+  styleUrl: 'sage-checkbox.scss',
+  shadow: true,
+})
+export class SageCheckbox {
+  /**
+   * Whether or not the checkbox is checked.
```
ju-Skinner: Can we update the wording here to be more complete?  We want this to read as documentation to the end-user. e.g Indicates if the checkbox is checked or not. Just an example here

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+  @Prop() checked: boolean;
+
+  /**
+   * Whether or not the checkbox is disabled.
+   */
+  @Prop() disabled: boolean;
+
+  /**
+   * Whether or not the checkbox is invalid.
+   */
```
ju-Skinner: Same comment as above

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * If `true`, the checkbox will visually appear as indeterminate.
+   * Only JavaScript can set the objects `indeterminate` property. See [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes).
+   */
+  @Prop() indeterminate: boolean;
+
+  /**
+   * Whether or not the checkbox is required.
+   */
```
ju-Skinner: Same as the previous "Whether...." statements above.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
ju-Skinner: I don't have a strong opinion either way. However, I will throw another possible solution into the mix. 😂 

```
const classNames = ['sage-checkbox'];

if (this.invalid) { classNames.push('is-invalid'); }
if (this.indeterminate) { classNames.push('is-indeterminate'); }
if (this.disabled) { classNames.push('is-disabled'); }

return classNames.join('  ');
```

If you decide to keep what you currently have, I would suggest not defining a variable for a single use. e.g const invalidClassname = 'is-valid'. It was longer to type that versus `className += '  is-invalid';`

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    const target = event.target as HTMLInputElement;
+    const isChecked = target.checked;
+
+    this.sageCheckboxChange.emit(isChecked);
+  }
+
+  private classNames() {
+    let className = `sage-checkbox`;
```
ju-Skinner: no need for ` (backtick), that is used for string interpolation e.g `some-string-${some.code.property}` would return "some-string-foo"

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+      className += ' ' + disabledClassName;
+    }
+
+    return className;
+  }
+
+  render() {
+    return (
+      <Host>
+        <div class={this.classNames()}>
```
ju-Skinner: Could we assign the class to `<Host class={this.classNames()}>`? If I recall, `<Host>` will output a `<div>`

libs/core/src/components/sage-checkbox/stories/sage-checkbox.docs.stories.mdx
```diff
@@ -0,0 +1,92 @@
...
+
+### Default
+
+<Story story={stories.Default} />
+
+<Canvas>
+  <sage-checkbox component-id="default" label="Label" />
+</Canvas>
+
+### Checked
```
ju-Skinner, line 36: Can we add verbiage here for `checked` property?

libs/core/src/components/sage-checkbox/test/sage-checkbox.e2e.ts
```diff
@@ -0,0 +1,44 @@
...
+
+describe('sage-checkbox', () => {
+  it('toggles checked and unchecked', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<sage-checkbox checkbox-id="default" label="Label text" />');
+    const component = await page.find('sage-checkbox');
+    expect(component).toHaveClass('hydrated');
+    const checkbox = await page.find('sage-checkbox >>> input');
+    await checkbox.click();
+    expect(await checkbox.getProperty('checked')).toBeTruthy();
```
ju-Skinner, line 13: whitespace is our friend, it assists with legibility.  Add newline at line 9

libs/core/src/components/sage-checkbox/test/sage-checkbox.e2e.ts
```diff
@@ -0,0 +1,44 @@
...
+    await checkbox.click();
```
ju-Skinner, line 15: add a new line here. It will be the same for the rest of this file. Also, we want to have a new line when starting a new test `it(....)`.

libs/core/src/components/sage-checkbox/test/sage-checkbox.spec.tsx
```diff
@@ -0,0 +1,158 @@
+import { newSpecPage } from '@stencil/core/testing';
+import { SageCheckbox } from '../sage-checkbox';
+
+describe('sage-checkbox', () => {
```
ju-Skinner, line 4: Same as the previous file.

ju-Skinner (changes requested):
See comments left. Feel free to reach out if you need assistance.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
pixelflips: The `<HOST>` does not output a div and complicates applying the styles since the classes are then applied to elements outside the shadow root. Let me know if I may be misunderstanding the ask.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
pixelflips: Not sure I understand here. Happy to change, but I don't feel like "indicates" is the correct wording. As a boolean, it's not an indicator, it either is or isn't. 

Let me know if you have any additional suggestions.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,128 @@
...
+  --border-radius: 4px;
```
anechol: Tokens are now available for use, if you'd like to add these in. Let's also try to stick to our naming convention for component specific tokens. If it feels weird, we can discuss further in our future meeting on consistency.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,128 @@
...
+}
+
+label {
+  margin-left: 10px;
+}
+
+.sage-checkbox__message {
+  color: var(--message-color);
+  font-size: 14px;
+  margin-left: 26px;
```
anechol: Should we consider using logical properties here?

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
anechol: "Determines," maybe?

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
anechol: Can you explain how this complicates applying styles in this case? We can make a final decision on this in our future meeting, but it'll be helpful to know your thoughts on this going into that.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
```diff
@@ -0,0 +1,106 @@
...
+            id={this.componentId}
+            name={this.name}
+            value={this.value}
+            checked={this.checked}
+            required={this.required}
+            disabled={this.disabled}
+            onChange={event => this.handleCheckboxChange(event)}
+          />
+          <label htmlFor={this.componentId}>{this.label}</label>
+          {this.helperMessage && <div class={'sage-checkbox__message'}>{this.helperMessage}</div>}
```
anechol: Mentioned this on the textarea component, but I'm on the fence about adding help messages to these components as it does break our expectations for having components like these serve as atoms. Any thoughts here?

anechol (changes requested):
Nice work on this! Got a few comments and suggestions below.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
pixelflips: I'm not aware of the naming conventions, is that something we've documented somewhere? Seems I may have missed it.

anechol: I can send you a link to the Slack thread I made about this, and yes, this will be officially documented.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
pixelflips: When I move the classes to the host, the classes then get applied to the element outside of the shadow-root, which removed/broke all the styles nested/applied to the current class. 

I tried nesting those same styles under the :host to no avail. Maybe my set up is incorrect. I'd be happy to pair up if you have some time.

libs/core/src/components/sage-checkbox/sage-checkbox.tsx
pixelflips: No real preference here. I think we may need to consult with design as most of the input elements have this.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
pixelflips: @anechol Thanks so much for the info and pairing up. I updated the naming and also used the tokens. They'll prob need some adjustments, but please have another look when you can.

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,128 @@
...
+        border-bottom: 0;
+        transform: rotate(87deg) translate3d(-114%, 25%, 0);
+      }
+    }
+  }
+}
+
+input {
+  appearance: none;
+  border: 1px solid var(--checkbox-border-color);
```
anechol: We now have a border shorthand token if you want to use that here

libs/core/src/components/sage-checkbox/sage-checkbox.scss
```diff
@@ -0,0 +1,128 @@
+:host {
+  --checkbox-background-color-disabled: var(--sage-color-neutral-grey-200);
```
anechol, line 2: Love the names here!

anechol (approved):
LGTM

pixelflips:
Refactored to remove the wrapper div. Requesting additional reviews.

cc: @ju-Skinner @anechol - TIA!

anechol (approved):

libs/core/src/components/sage-checkbox/test/sage-checkbox.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+  it('renders', async () => {
+    const page = await newSpecPage({
+      components: [SageCheckbox],
+      html: `<sage-checkbox />`,
+    });
+    expect(page.root).toEqualHtml(`
```
ju-Skinner, line 11: Add a whitespace here and on line 25

ju-Skinner (approved):
Great job!!!

PR #48, opened 2023-04-11
[DSS-372]: Textarea Component

goodwinchris:
# Description
Addition of Textarea component with minor alignments to Input component documentation.

Fixes [DSS-372: Form Textarea](https://kajabi.atlassian.net/browse/DSS-372)

CleanShot 2023-04-11 at 13 20 53@2x

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [x] accessibility tests
- [x] tested manually

**Test Configuration**:

- sage versions:
- os:
- browsers:
- screen readers:
- misc:

# Checklist:


libs/core/src/components/sage-input/sage-input.scss
```diff
@@ -111,7 +109,7 @@ input {
     color: var(--color-placeholder);
   }
 
-  &:has(~.sage-input__error-text) {
+  &:has(~ .sage-input__error-text) {
```
goodwinchris: Just a note: This css doesn't work in FF yet but it is coming. It works in Firefox with flags enabled. Just something to keep an eye on for now I think.

libs/core/src/components/sage-textarea/sage-textarea.scss
```diff
@@ -0,0 +1,128 @@
...
+    border-color: var(--border-color-focus-visible);
+    outline: 4px solid var(--focus-visible-outline-color);
+  }
+
+  &::placeholder {
+    color: var(--color-placeholder);
+    font-style: italic;
+  }
+
+  &:has(~ .sage-textarea__error-text) {
```
goodwinchris: Same note as above: This css doesn't work in FF yet but it is coming. It works in Firefox with flags enabled. Just something to keep an eye on for now I think.

libs/core/src/components.d.ts
```diff
@@ -5,6 +5,7 @@
  * It contains typing information for all components that exist in this project.
  */
 import { HTMLStencilElement, JSXBase } from "@stencil/core/internal";
+import { ChangeEvent } from "react";
```
goodwinchris: I'm not sure where this came from but I'm happy to remove it if needed.

libs/core/src/components/sage-textarea/sage-textarea.scss
anechol: Could we place an `@supports` block here along with a fallback for FF?

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+@Component({
+  tag: 'sage-textarea',
+  styleUrl: 'sage-textarea.scss',
+  shadow: true,
+})
+export class SageTextarea {
+  @Element() el: HTMLDivElement;
+
+  /**
+   * Indicates whether or not the textarea is disabled
```
anechol, line 19: Nice work on the documentation here on all these props. Love to see it

libs/core/src/components/sage-textarea/sage-textarea.scss
```diff
@@ -0,0 +1,128 @@
+:host {
+  // display: block;
+  /**
+
+    * @prop --background: Background of the textarea
```
anechol: Is it expected that this background is shorthand? If not, can we specify whether it's a color or something else?

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+    this.sageTextareaInput.emit();
+  };
+
+  render() {
+    return (
+      <Host
+        aria-disabled={this.disabled ? 'true' : null}
+      >
+        <div class="sage-textarea">
+          {this.labelText && <label htmlFor={this.componentId}>{this.labelText}</label>}
```
anechol: Should we consider making a label its own component since more than one component will use it?

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+            disabled={this.disabled}
+            id={this.componentId}
+            name={this.name}
+            placeholder={this.placeholderText}
+            readOnly={this.readonly}
+            required={this.required}
+            rows={this.rows}
+            onChange={this.onTextareaInputEvent}
+          >{this.value}</textarea>
+          {this.hintText && <p class="sage-textarea__hint-text">{this.hintText}</p>}
```
anechol: I'm on the fence about adding hint text here, because that breaks our intention of keeping components as single atoms.

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+          {this.invalid && <p class="sage-textarea__error-text">{this.errorText}</p>}
```
anechol: Same as above.

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+   * Indicates whether or not the textarea is invalid or throws an error
+   * @defaultValue null
+   */
+  @Prop({mutable: true}) invalid?: boolean = null;
+
+  /**
+   * A unique identifier for the textarea
+   * @defaultValue null
+   */
+  @Prop() componentId?: string = null;
```
pixelflips: Wondering about the naming here, since `inputId` & `checkboxId` were used in those components. Probably worth a discussion around if we should standardize some prop naming that will be available across components.

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * Text to be displayed as the textarea label
+   * @defaultValue null
+   */
+  @Prop() labelText?: string = null;
```
pixelflips: Same thought about naming here. Others just used `label`

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * Event emitted whenever the value of the textarea changes
+   */
+  @Event() sageTextareaInput: EventEmitter<ChangeEvent>;
+  private onTextareaInputEvent = (ev: Event) => {
+    const textarea = ev.target as HTMLTextAreaElement;
+    if (this.required === true) {
+      const validity = textarea.checkValidity();
+      (validity === false) ? this.invalid = true : this.invalid = false;
```
pixelflips: Is the validity check something we would also want to standardize across input-related components?

libs/core/src/components/sage-textarea/stories/sage-textarea.docs.stories.mdx
```diff
@@ -0,0 +1,117 @@
...
+#### When to use
+- You want a response longer than one line
+- You want to capture rich text input
+
+#### When not to use
+- To gather basic information, use an [Input](/docs/components-input--text)
+- You want a response that is one line of text, use an [Input](/docs/components-input--text)
+
+### Accessibility
+- The textarea component comes with a built-in label through the `label` property, it is strongly encouraged that you include it.
```
pixelflips: I think the prop should be `labelText` here.

libs/core/src/components/sage-textarea/test/sage-textarea.spec.tsx
```diff
@@ -0,0 +1,271 @@
...
+          <div class="sage-textarea">
+            <label htmlFor="testId">label</label>
+            <textarea class="sage-textarea__field" id="testId"></textarea>
+          </div>
+        </mock:shadow-root>
+      </sage-textarea>
+    `);
+  });
+
+
```
pixelflips: Extra line here.

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+  render() {
+    return (
+      <Host
+        aria-disabled={this.disabled ? 'true' : null}
+      >
+        <div class="sage-textarea">
+          {this.labelText && <label htmlFor={this.componentId}>{this.labelText}</label>}
+          <textarea class="sage-textarea__field"
```
pixelflips: Is the class needed? Not seeing any styles attached.

pixelflips (commented):
This is looking really great. A couple of small things and some areas where we may need to decide as a team. Let me know if anything is unclear.

libs/core/src/components/sage-textarea/sage-textarea.tsx
anechol: Agree here.

anechol (changes requested):
Great work so far! Some things to consider/change listed below.

anechol (changes requested):
Great work so far! Some things to consider/change listed below.

libs/core/src/components/sage-textarea/sage-textarea.tsx
goodwinchris: I got this wrong, I think it should be `textareaId`

libs/core/src/components/sage-textarea/sage-textarea.scss
```diff
@@ -0,0 +1,128 @@
...
+    outline-color: var(--focus-visible-outline-color-error);
+  }
+}
+
+.sage-textarea__error-text,
+.sage-textarea__hint-text {
+  font-size: var(--hint-font-size);
+  font-weight: var(--hint-font-weight);
+  line-height: var(--hint-line-height);
+  margin-bottom: 0;
```
QuintonJason: `margin-bottom` -> `margin-block-end`

libs/core/src/components/sage-textarea/sage-textarea.scss
```diff
@@ -0,0 +1,128 @@
...
+  margin-top: var(--hint-margin-top);
```
QuintonJason: `margin-top` -> `margin-block-start`

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+  shadow: true,
+})
+export class SageTextarea {
+  @Element() el: HTMLDivElement;
+
+  /**
+   * Indicates whether or not the textarea is disabled
+   * @defaultValue false
+   */
+  @Prop() disabled? = false;
```
QuintonJason: Since the prop has an `?` this implies that it's optional. The `@defaultValue` implies that the prop is required. One or the other has to be removed, it's your discretion.

libs/core/src/components/sage-textarea/sage-textarea.tsx
QuintonJason: Text is used in the description, does it also need to be in the prop name?

libs/core/src/components/sage-textarea/sage-textarea.tsx
QuintonJason: Utils?

libs/core/src/components/sage-textarea/sage-textarea.tsx
QuintonJason: I had this thought during Input creation and still not sure if it's best, but I think about it

libs/core/src/components/sage-textarea/stories/sage-textarea.docs.stories.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+# Textarea
+A textarea field coupled with a label and error text which allows for user input and feedback.
+
+## Guidelines
+### Usage
+
+#### When to use
+- You want a response longer than one line
+- You want to capture rich text input
```
QuintonJason: I don't think this is accurate. It's still plain text.

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Specifies the name, submitted with the form name/value pair
+   */
+  @Prop() name: string;
+
+  /**
+   * Specifies a short hint that describes the expected value of the textarea
+   * @defaultValue null
+   */
+  @Prop() placeholderText?: string = null;
```
QuintonJason: I normally see this implementation as placeholder={placeholder}.

libs/core/src/components/sage-textarea/sage-textarea.tsx
goodwinchris: Update: will keep as `componentId` to align with discussion we had around naming

libs/core/src/components/sage-textarea/stories/sage-textarea.stories.js
```diff
@@ -0,0 +1,97 @@
...
+export const Hint = BaseTemplate.bind({});
+Hint.args = {
+  name: 'Hint',
+  labelText: 'Name',
+  componentId: 'uniqueIdHint',
+  hintText: 'Hint'
+};
+Hint.parameters = { ...defaultParameters };
+
+export const Error = BaseTemplate.bind({});
```
QuintonJason: When testing this, removing the errorText doesn't remove the error styles

libs/core/src/components/sage-textarea/sage-textarea.tsx
goodwinchris: Reverted back as per discussion

libs/core/src/components/sage-textarea/test/sage-textarea.spec.tsx
```diff
@@ -0,0 +1,271 @@
...
+    const page = await newSpecPage({
+      components: [SageTextarea],
+      html: `<sage-textarea value="initial"></sage-textarea>`,
+    });
+
+    const sageTextarea = page.root;
+    expect(sageTextarea?.value).toBe('initial');
+
+    const textarea = sageTextarea?.shadowRoot?.querySelector<HTMLTextAreaElement>('textarea');
+    expect(textarea?.innerHTML.toString()).toBe('initial');
```
QuintonJason, line 255: Because the expected string value changes, does that make this an e2e test vs a unit test or am I incorrect?

QuintonJason (changes requested):
Good stuff! Left some comments and it should be good to go

libs/core/src/components/sage-textarea/sage-textarea.scss
goodwinchris: 👍 - updating in input as well to keep alignment between both

libs/core/src/components/sage-textarea/sage-textarea.tsx
goodwinchris: @QuintonJason My understanding is that a value is marked as optional when it isn't required for someone using the web component to provide a value for that prop. That still leaves us with the choice to provide a default value for the prop which a user can override by supplying their own, optionally. 

I checked the stencil docs to confirm this notion and was unable to find anything noting that optional props can't have default values, further the [docs show an optional label prop with a default value](https://stenciljs.com/docs/properties#optional-types).

Am I missing something with how to use optional?

libs/core/src/components/sage-textarea/stories/sage-textarea.docs.stories.mdx
goodwinchris: You're right, currently, it doesn't. I'll remove this. In the future when we have advanced text formatting we'll want to note it again.

libs/core/src/components/sage-textarea/stories/sage-textarea.stories.js
goodwinchris: That is correct, removing error text shouldn't remove error styles. Instead entering valid input into the textarea OR removing the required attribute should remove error styles.

libs/core/src/components/sage-textarea/sage-textarea.tsx
goodwinchris: Both this hint text and error text follow the pattern established in the Input component, I'm temped to leave it how we have it and revisit if needed.

libs/core/src/components/sage-textarea/stories/sage-textarea.stories.js
```diff
@@ -2,12 +2,12 @@ import { html } from 'lit-html';
...
-  error-text="${args.errorText}"
-  hint-text="${args.hintText}"
+  error-message="${args.errorMessage}"
+  hint-message="${args.hintMessage}"
   invalid="${args.invalid}"
-  label-text="${args.labelText}"
+  label="${args.label}"
   name="${args.name}"
-  placeholder-text="${args.placeholderText}"
+  placeholderMessage="${args.placeholderMessage}"
```
goodwinchris: To fix

libs/core/src/components/sage-textarea/sage-textarea.tsx
anechol: With my comment below about helper/error text, I think it could serve as a conversation piece. I'll get something in Slack some time this week to discuss.

libs/core/src/components/sage-textarea/stories/sage-textarea.stories.js
QuintonJason: That sounds right. In that case, do we even need this has stuff: https://github.com/Kajabi/sage/blob/adf91b7f6a265d7993de41ff6bf23870f7b283d3/libs/core/src/components/sage-textarea/sage-textarea.scss#L117?

goodwinchris: For now with the FF fallback, no. But I think there's value to it once we can remove the `@supports` because it saves us from having to add an `is-invalid` class to the component to do all our styling with.

libs/core/src/components/sage-textarea/sage-textarea.scss
QuintonJason: I think it's best to save it for when we have our consistancy pass.

goodwinchris: Ok, will revert

libs/core/src/components/sage-textarea/sage-textarea.tsx
QuintonJason: You right on this, nevermind

libs/core/src/components/sage-textarea/sage-textarea.tsx
QuintonJason: @QuintonJason  This currently exists in utils but needs tests

libs/core/src/utils/utils.spec.ts
```diff
@@ -17,3 +20,9 @@ describe('format', () => {
     expect(format('Joseph', 'Quincy', 'Publique')).toEqual('Joseph Quincy Publique');
   });
 });
+
+describe('isRequired', () => {
```
QuintonJason, line 24: Note to @QuintonJason Make better tests in a followup PR. Checking for validity is dependent on the element type which is different per form element.

libs/core/src/components/sage-input/stories/sage-input.docs.stories.mdx
```diff
@@ -30,18 +30,17 @@ import * as stories from './sage-input.stories.js';
...
-- You want a response longer than one line, use a textarea (TODO: create textarea and link here)
+- You want a response longer than one line, use a [Textarea](/docs/components-textarea--default)
 - You want a response from a predefined list, use a select (TODO: create select and link here)
 
 
 ### Accessibility
-- The input component comes with a built-in label through the `label` property.   It is strongly encouraged that you include it
-- Please provide clear and descriptive error messages
+- The input component comes with a built-in label through the `label` property, it is strongly encouraged that you include it
+- Please provide clear and descriptive error, hint, and placeholder messages
```
anechol, line 43: How do you feel about making a note here that using a placeholder is *not* a replacement for a label?

libs/core/src/components.d.ts
```diff
@@ -597,7 +663,7 @@ declare namespace LocalJSX {
          */
         "invalid"?: boolean;
         /**
-          * Text to be displayed as the form label
+          * Text to be displayed as the input label
```
pixelflips, line 666: Should we say textarea instead of input, so it matches other descriptions? This is a nitpick and not a blocker.

libs/core/src/components/sage-input/stories/sage-input.docs.stories.mdx
```diff
@@ -30,18 +30,17 @@ import * as stories from './sage-input.stories.js';
 ### Usage
 
 #### When to use
-
 - To gather basic usage information
 - You want a response that is one line of text
 
 #### When not to use
-- You want a response longer than one line, use a textarea (TODO: create textarea and link here)
+- You want a response longer than one line, use a [Textarea](/docs/components-textarea--default)
```
pixelflips, line 37: ❤️

libs/core/src/components/sage-textarea/sage-textarea.scss
```diff
@@ -0,0 +1,96 @@
...
+  font-size: var(--label-font-size);
+  font-weight: var(--label-font-weight);
+  line-height: var(--label-line-height);
+  margin-block-end: var(--label-margin-block-end);
+}
+
+.sage-textarea__field {
+  background-color: var(--background-color);
+  border: 1px solid var(--border-color-default);
+  border-radius: var(--sage-border-radius-md);
```
pixelflips, line 53: Could this be updated to a local var?

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,164 @@
...
+export class SageTextarea {
+  @Element() el: HTMLDivElement;
+
+  /**
+   * A unique identifier for the textarea
+   */
+  @Prop() componentId?: string;
+
+  /**
+   * Indicates whether or not the textarea is disabled
```
pixelflips, line 19: For the boolean props, I believe we agreed to use "Determines" instead of "Indicates". Not a blocker, since it will depend on Ashley's upcoming guidelines/documentation

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,164 @@
...
+    this.sageTextareaChange.emit();
+  };
+
+  private textareaClassNames = () => {
+    let className = `sage-textarea__field`;
+    if (this.invalid && this.invalid === true) {
+      const invalidClassName = `is-invalid`;
+      className += ' ' + invalidClassName;
+    }
+    return className;
```
pixelflips: Another approach could be used here that would remove the need for the `+= ' ' +`

```
private textareaClassNames() {
  const classNames = ['sage-textarea__field'];

  if (this.invalid && this.invalid === true) {
    classNames.push('is-invalid');
  }

  return classNames.join('  ');
}
```

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Create id for messaging
+   */
+  private messageId = (id: string, messageType: string) => {
+    return `${id}__${messageType}-message`;
+  };
+
+  /**
+   * Assign aria-description id to relate messages with input
```
pixelflips, line 111: ❤️

pixelflips (approved):
A few tiny nitpicks and some suggestions, but nothing blocking. Overall LGTM! 👍🏼 - Nice work on getting this over the finish line.

libs/core/src/components.d.ts
ju-Skinner: This should be removed

libs/core/src/components/sage-textarea/stories/sage-textarea.stories.js
```diff
@@ -0,0 +1,97 @@
...
+export const Hint = BaseTemplate.bind({});
+Hint.args = {
+  componentId: 'uniqueIdHint',
+  hintMessage: 'Hint',
+  label: 'Name',
+  name: 'Hint',
+};
+Hint.parameters = { ...defaultParameters };
+
+export const Error = BaseTemplate.bind({});
```
pixelflips: Could we call this Invalid to match the other inputs?

anechol (approved):
Just a non-blocking comment. Really nice work on this!

anechol (approved):
Just a non-blocking comment. Really nice work on this!

libs/core/src/components/sage-textarea/sage-textarea.tsx
QuintonJason: I was saving this for the completion of the guidelines work. It would likely be handled in a sweeping change

libs/core/src/components.d.ts
QuintonJason: I believe that's correct as this lives in the `SageInput` interface

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,169 @@
...
+   * The value of the textarea
+   */
+  @Prop({mutable: true}) value?: string;
+
+  /**
+   * Event emitted whenever the value of the textarea changes
+   */
+  @Event() sageTextareaChange: EventEmitter<TextareaChangeEventDetail>;
+
+
```
ju-Skinner: unneeded line break here

libs/core/src/components/sage-textarea/sage-textarea.tsx
```diff
@@ -0,0 +1,169 @@
+import { Component, Element, Host, h, Prop, Event, EventEmitter } from '@stencil/core';
+import { isRequired } from '../../utils/utils';
+import { TextareaChangeEventDetail } from './textarea-interface';
+
+
```
ju-Skinner: unneeded linebreak here as well

libs/core/src/components/sage-textarea/test/sage-textarea.spec.tsx
ju-Skinner: This is ok to do in a spec.

libs/core/src/utils/utils.spec.ts
ju-Skinner: It may be better to add a comment in the code, so that there is a reminder.

```
//TODO: Make better test for isRequired function
```

ju-Skinner (approved):
Looks good. Left a couple of comments around formatting, which should be fixed prior to merging.

PR #50, opened 2023-04-13
chore: add engines to package.json

ju-Skinner:
# Description
This assists developers with knowing which node versions are compatible with the Design System. 

- [X] Chore (other changes that don't affect source or test files)

# How Has This Been Tested?
N/A

**Test Configuration**:

- Sage versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

- [X] No options apply

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,12 @@
 {
-  "name": "sage",
   "private": true,
   "workspaces": [
     "apps/*",
     "libs/*"
   ],
+  "engines": {
+    "node": ">= 16"
+  },
   "scripts": {
     "bootstrap": "lerna bootstrap",
     "build.all": "nx run-many --target=build",
--- a/.github/workflows/actions/build-core/action.yml
+++ b/.github/workflows/actions/build-core/action.yml
@@ -1,6 +1,11 @@
 name: 'Build Sage Core'
 description: 'Builds Sage Core library'
 
+inputs:
+  node-version:
+    description: 'The version of node to use'
+    required: true
+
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -1,10 +1,12 @@
 {
-  "name": "sage",
   "private": true,
   "workspaces": [
     "apps/*",
     "libs/*"
   ],
+  "engines": {
+    "node": ">= 16"
```
voodooGQ, line 8: Any reason we're starting with 16 instead of 18? 16 stopped receiving security support on 10-18-2022

ju-Skinner: Yes, you are right. Does that mean that people should be forced to use 18 to develop within our ecosystem? If I recall, we are saying that we know that our code works with this minimum version of node. End of Life for 16 is later this year, you can find out more [here](https://nodejs.dev/en/about/releases/)

rwc9u (approved):

package.json
voodooGQ: You're correct, but with that stance we have to have some sort of testing scenarios in place to verify 16 and 18 both work (probably from CI). If that's in the plans then I think this is totally reasonable, if not I'd say that we should probably put it on the docket.

ju-Skinner: See the latest [commit](https://github.com/Kajabi/sage/pull/50/commits/a94f8d2ec4e3ddead98d190bdf831b054eabc509)

voodooGQ (approved):

PR #51, opened 2023-04-18
[DSS-386] Tokens - Adds function to format shorthand tokens

anechol:
# Description

Fixes #(issue)

## Type of change

- [x] Refactor of build code

How to test:
1. Run `yarn nx run core:build.tokens`
2. Check that tokens generated/updated in `libs/core/src/global/styles/tokens` are formatted correctly with shorthand token options available.

# Checklist:

### DIFF
--- a/libs/core/scripts/build-tokens.ts
+++ b/libs/core/scripts/build-tokens.ts
@@ -12,7 +12,7 @@ const propertyOrder = {
  * Runs the process to build tokens from JSON and save them to their
  * appropriate files.
  *
- * @param data Tokens data in JSON format.
+ * @param data Tokens data in JSON format
  */
 const build = (data: object) => {
 
@@ -36,7 +36,7 @@ build(tokens);
 /*** Helper Methods ***/
 
 /**
- * Generates the custom shorthand for the CSS Custom Property
+ * Generates the custom shorthand for the CSS Custom Property.
  *
  * @param customPropName Name of the Custom Property
  * @param cssPropertyName Name of the CSS Property
@@ -93,7 +93,7 @@ const determinePropertyOrder = (cssProperty, data) => {
  * Reads the index.scss file if it exists and adds an import statement per category name.
  * If the file doesn't exist, it will be created.
  *
- * @param categoryName The name of the token category.
+ * @param categoryName The name of the token category
  */
 const formatIndexFile = (categoryName: string) => {
 	const indexFile = `${process.cwd()}/src/global/styles/tokens/index.scss`;
@@ -111,6 +111,7 @@ const formatIndexFile = (categoryName: string) => {
 }
 
 /**
+ * Receives values of CSS property names and joins them.
  * 
  * @param data Array data to be joined
  * @returns CSS property values joined by a space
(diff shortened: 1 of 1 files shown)


libs/core/scripts/build-tokens.ts
```diff
@@ -44,43 +111,58 @@ const formatIndexFile = (categoryName: string) => {
...
- * appropriate files.
- * 
- * @param data Tokens data in JSON format.
+ * Takes the JSON data and converts it to formatted CSS
+ * custom properties.
+ *
+ * @param tokens Tokens data in JSON format.
+ * @param cssPropertyName Name of the css property
+ * @param base Base string to format custom properties.
+ * @returns Formatted CSS with all custom properties.z
```
ju-Skinner: looks like an extra character `z`

libs/core/scripts/build-tokens.ts
```diff
@@ -44,43 +111,58 @@ const formatIndexFile = (categoryName: string) => {
 }
 
 /**
- * Saves formatted tokens to their own SCSS partial files.
  * 
+ * @param data Array data to be joined
+ * @returns CSS property values joined by a space
+ */
+const getValues = (data) => Object.values(data).join(' ');
```
ju-Skinner, line 118: need a `tab` here

ju-Skinner (changes requested):
formatting change. Nothing major

libs/core/scripts/build-tokens.ts
anechol: Fixed this

libs/core/scripts/build-tokens.ts
```diff
@@ -1,32 +1,99 @@
...
+
+	} catch (e) {
+		console.log(
+			"Provide a correct argument - a relative path to design tokens.\n",
+			e
+		)
+	}
+}
+
+build(tokens);
```
teenwolfblitzer: You were right: looks like moving this line up to kick off the build process is the reason for the error I'm seeing. I'd suggest moving it back to the end of the file, along with the entire `build` function block to keep it all together.

```
Provide a correct argument - a relative path to design tokens.
 ReferenceError: Cannot access 'tokensToCss' before initialization
    at /Users/kevin.chang/Sites/sage/libs/core/scripts/build-tokens.js:34:34
```

teenwolfblitzer (changes requested):
Looking great! The execution order is breaking the build for me locally, however.

libs/core/scripts/build-tokens.ts
anechol: This is fixed

ju-Skinner (approved):
Great job!!!!

libs/core/scripts/build-tokens.ts
ju-Skinner: Just as a heads up @monkeypox8, we would like to keep our functions/methods in alphabetical order. This works since they have to be compiled into javascript in most cases.

teenwolfblitzer: Nice, I'm a fan of this as a convention. My only concern would be with synchronous functions similar to here, but in all likelihood they'll be pretty rare.

teenwolfblitzer (approved):
👍 LGTM, nice work!

QuintonJason (approved):
Looks and works great!

PR #53, opened 2023-04-26
[DSS-384] Avatar - Adds component

anechol:
# Description

Adds a new Avatar component.

[Figma](https://www.figma.com/file/k6e1a1j4OS8yYr54H1v0B4/%5BWIP%5D--%F0%9F%A7%A9-Pine-components?type=design&node-id=24106%3A18511&t=FIv3ICJSC9DFcped-1)

[DSS-384](https://kajabi.atlassian.net/browse/DSS-384)

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [x] accessibility tests
- [x] tested manually

**Test Configuration**:

- Sage versions: latest
- OS: 12.6
- Browsers: Latest

# Checklist:

[DSS-384]: https://kajabi.atlassian.net/browse/DSS-384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,62 @@
...
+   * Preset sizes for the avatar. If a custom size is desired,
+   * use the `--size` custom property instead.
+   * @defaultValue md
+   */
+  @Prop() size?: 
+  | 'xl'
+  | 'lg'
+  | 'md' 
+  | 'sm' 
+  | 'xs';
```
ju-Skinner: I don't see a default value set here. You want to add `= "md"` to the end. This will fit with what you have in the tsdoc comment.

libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,62 @@
...
+   * Determines whether the badge is visible or not
+   * @defaultValue false
+   */
+  @Prop() badge? = false;
+
+  /**
+   * The src for a custom user image.
+   * @defaultValue null
+   */
+  @Prop() image?: string;
```
ju-Skinner: Here you will probably want to change from `string` to `string | null`, because null cannot be casted to a string.

ju-Skinner (changes requested):
These are a few changes I saw that we needed prior to running locally. I'll circle back to view Storybook after lunch.

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,74 @@
...
+
+:host(.sage-avatar--admin) div {
+  border-radius: 12px;
+}
+
+div {
+  --radius-round: 100px;
+
+  align-items: center;
+  background-color: var(--sage-color-primary-200);
```
pixelflips: Should this be `primary-100`? The background is not matching the figma file.

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,74 @@
...
+  block-size: var(--size);
+  border: var(--sage-border-default);
```
pixelflips: I've been assigning all of our tokens to variables instead of assigning them directly. I thought I remembered something about using that approach instead, but could be wrong.

libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,62 @@
...
+    return (
+      <Host
+        class={this.classNames({
+          'sage-avatar': true,
+          ...(this.size && { [`sage-avatar--${this.size}`] : []}),
+          ...(this.variant === 'admin' && { [`sage-avatar--${this.variant}`] : []}),
+        })}
+      >
+        <div>
+          {this.image ? <img alt='Profile' src={this.image} /> : <sage-icon name="user-filled" size="normal"></sage-icon>}
```
pixelflips: Should we allow the image `alt` to be customized?

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,74 @@
...
+    width: $value;
+  }
+}
+
+:host(.sage-avatar--admin) div {
+  border-radius: 12px;
+}
+
+div {
+  --radius-round: 100px;
```
pixelflips: I believe `50%` could be used here to create a circle.

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,74 @@
...
+
+  align-items: center;
+  background-color: var(--sage-color-primary-200);
+  block-size: var(--size);
+  border: var(--sage-border-default);
+  border-color: transparent;
```
pixelflips: Is a border actually needed if it's set to be transparent?

libs/core/src/components/sage-avatar/sage-avatar.tsx
anechol: I removed the default value here because it was overriding the `--size` custom property due to CSS specificity. If we still want a default here, I can find some way to increase the specificity to have the custom prop always override the preset values _or_ use `!important` (I do not desire this route).

anechol: I don't think we established how we want to tackle values that have preset and custom options, but I am more than happy to set a precedent, see how the rest of the team feels about it, and add it to the best practices documentation.

libs/core/src/components/sage-avatar/stories/sage-avatar.docs.stories.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+# Avatar
+
+Avatar is used for showing a thumbnail representation of an admin, customer, or business.
+
+## Properties
+
+<ArgsTable of="sage-avatar" />
+
+### Default
```
pixelflips: The default and admin stories are showing up the same for me and the props are not populating in the admin example:

Screenshot 2023-06-05 at 11 14 26 AM

libs/core/src/components/sage-avatar/stories/sage-avatar.stories.js
```diff
@@ -0,0 +1,46 @@
+import { html } from 'lit-html';
+import imgFile from '../assets/demi_wilkinson.png';
```
pixelflips: Nitpick, but could this be converted to a `.jpg` to discourage developers from possibly using a `.png` for non-transparent images/photos?

pixelflips (changes requested):
This is off to a really great start. A few comments and seeing some issues with the variant documentation examples.

libs/core/src/components/sage-avatar/sage-avatar.tsx
ju-Skinner: If you removed it, then you proably want to remove it from the tsdocs. It currently shows `@defaultValue md`, but we don't have a defaultValue.  You will need to recompile or have your server running to generate the new `components.d.ts` file.

libs/core/src/components/sage-avatar/stories/sage-avatar.docs.stories.mdx
```diff
@@ -0,0 +1,57 @@
+import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs';
+
+import { extractArgTypes } from '@pxtrn/storybook-addon-docs-stencil';
+
+import * as stories from './sage-avatar.stories.js';
+
+<Meta title="components/Avatar" component="sage-avatar" argTypes={extractArgTypes('sage-avatar')} />
```
QuintonJason: Can the `<Meta />` include `args={{}}` of the boolean props set to false? This should allow the toggles in storybook to reflect the correct without having to click on them

libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  @Prop({ reflect: true }) variant?: 'customer' | 'admin' = 'customer'
+
+  private classNames = (props: object) => {
+    return Object.keys(props).join(' ');
+  }
+
+  render() {
+    return (
+      <Host
+        class={this.classNames({
```
QuintonJason: 💯

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,74 @@
+:host {
+  /**
+    * @prop --size: Custom size of avatar. Setting this will *not* override a set `size` prop value.
+  */
+
+  --size: var(--sage-spacing-2xl);
```
QuintonJason: Can you add the `display: inline-block`? When inspecting the element doesn't fill the entire parent

QuintonJason (changes requested):
Looking good. Left a couple comments.

libs/core/src/components/sage-avatar/sage-avatar.tsx
anechol: In this case, I'm not entirely sure. I can make a prop for this, no problem. I'm wondering if an avatar's use case will be so different that it will warrant needing a new alt? Any thoughts here? 🤔

libs/core/src/components/sage-avatar/sage-avatar.scss
anechol: Ha, thanks! Not sure how I missed this

libs/core/src/components/sage-avatar/sage-avatar.scss
anechol: You're right, I will get this updated

libs/core/src/components/sage-avatar/sage-avatar.tsx
pixelflips: Yeah, I was a bit on the fence about it so I was just curious to see what you thought. I can see it being beneficial for something like a contacts list where there could be quite a few different photos. I don't think it's a showstopper for now.

libs/core/src/components/sage-avatar/sage-avatar.scss
anechol: Thanks for pointing this out. For some reason, I was always under the impression that there needed to be a defined border in order for the radius to take effect, but apparently not. TIL.

libs/core/src/components/sage-avatar/stories/sage-avatar.docs.stories.mdx
anechol: Minor bug with the admin story, this is fixed in the coming commits.

libs/core/src/components/sage-avatar/stories/sage-avatar.stories.js
anechol: This is a good idea. I'm updating this.

libs/core/src/components/sage-avatar/sage-avatar.scss
anechol: Will change this.

libs/core/src/components/sage-avatar/sage-avatar.tsx
anechol: That's a good point. It wouldn't hurt to add one. If we feel otherwise, we can always reconsider.

libs/core/src/components/sage-avatar/stories/sage-avatar.stories.js
```diff
@@ -0,0 +1,48 @@
...
+>
+</sage-avatar>`;
+
+const AdminTemplate = (args) => html`<sage-avatar size="${args.size}" variant="${args.variant}"></sage-avatar>`;
+
+const BadgeTemplate = (args) => html`<sage-avatar badge="${args.badge}" size="${args.size}"></sage-avatar>`;
+
+const CustomSizeTemplate = (args) => html`<sage-avatar style="--size: 128px;"></sage-avatar>`;
+
+const ImageTemplate = (args) => html`<sage-avatar alt="${args.alt}" image="${args.image}" size="${args.size}"></sage-avatar>`;
```
pixelflips: I think with the templates defined this way it's messing with the storybook controls slightly. I am not positive, but for example, when looking at the image on the canvas tab, switching variants does not seem to apply. 

I think you could add additional props to the templates or update each to use the BaseTemplate. Let me know if that makes no sense.

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,79 @@
...
+
+@each $size, $value in $sage-avatar-sizes {
+  :host(.sage-avatar--#{$size}) div {
+    height: $value;
+    width: $value;
+  }
+}
+
+:host(.sage-avatar--admin) div {
+  border-radius: 12px;
```
pixelflips: May want to check with design here. I was seeing some usage of 12 in other components and it turned out it was incorrect. Would be great if they'd agree to one of the token sizes. Definitely not a show-stopper.

libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+  styleUrl: 'sage-avatar.scss',
+  shadow: true,
+})
+export class SageAvatar {
+
+  /**
+   * The alt for a custom user image.
+   * @defaultValue null
+   */
+  @Prop() alt?: string | null;
```
pixelflips: 🔥

libs/core/src/components/sage-avatar/stories/sage-avatar.stories.js
```diff
@@ -0,0 +1,48 @@
+import { html } from 'lit-html';
+import imgFile from '../assets/demi_wilkinson.jpg';
```
pixelflips: ❤️

pixelflips (changes requested):
This is super close and looks great. Mostly just seeing some minor issues with the storybook controls. UI all looks great!

I missed this during my first round of review, but I am not seeing any interactive states (hover, focus, etc...), and wondering your thoughts on those since Avatar could be used in the dropdown.

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,79 @@
...
+  // Percentage is average size of icon in relation to total avatar size of
+  // all preset sizes found in Figma
+  // Used to allow icons to scale to container size
+  height: 33.53%;
+  width: 33.53%;
+}
+
+.sage-avatar__badge {
+  background-color: var(--badge-background-color);
+  border: var(--sage-border-default);
```
pixelflips: Could use the `--border` prop here?

libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Determines the variant of avatar. Changes appearance accordingly.
+   * @defaultValue customer
+   */
+  @Prop({ reflect: true }) variant?: 'customer' | 'admin' = 'customer'
+
+  private renderBadge = () => (
+    this.badge
+      && <sage-icon class="sage-avatar__badge" name="check-circle-filled" size="normal"></sage-icon>
```
ju-Skinner: you could pass your 33.53% here to the `size` prop, which would remove the need for the style in the `sage-avatar.scss`. 

Also named values for `size` are below...This may be why you were having issues with appropriate sizing. Default is `regular` if you don't pass `size`
```
'small'   // 12px
    | 'regular'  // 16px
    | 'medium'  // 20px
    | 'large'   // 24px
    | 'auto'
```

libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,77 @@
...
+  );
+
+  private renderIconOrImage = () => (
+    this.image
+      ? <img alt={this.alt} src={this.image} />
+      : <sage-icon name="user-filled" size="normal"></sage-icon>
```
ju-Skinner: see comment above

ju-Skinner (changes requested):
See comments below

libs/core/src/components/sage-avatar/stories/sage-avatar.stories.js
anechol: This should be fixed now.

libs/core/src/components/sage-avatar/sage-avatar.tsx
anechol: This has been addressed.

libs/core/src/components/sage-avatar/sage-avatar.tsx
anechol: Same as above

anechol:
Refactored the size prop to remove unnecessary styles and class names. Added a `dropdown` prop so that the Avatar can be used as a dropdown trigger.

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,68 @@
...
+  background: transparent;
+  border: 0;
+  border-radius: 50%;
+  cursor: pointer;
+  display: flex;
+  padding: 0;
+
+  &:focus-visible {
+    // outline: var(--sage-border-focus); Border radius for outline does not work in Safari
+    box-shadow: var(--box-shadow-focus); // Remove when outline radius is supported in Safari
```
pixelflips: I think this is showing as 2px and should be 4px.

libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,124 @@
...
+      // Percentage is average size of icon in relation to total avatar size 
+      // of all preset sizes found in Figma. 
+      // Used to allow icons to scale to container size
+      : <sage-icon name="user-filled" size="33.53%"></sage-icon>
+  );
+
+  private classNames = () => (
+    {
+      'sage-avatar': true,
+      [`sage-avatar--has-image`]: this.image !== undefined || null, // Remove when FF supports :has selector
```
pixelflips: I believe this conditional is adding the `--has-image` class at all times. If you check the canvas tab, the background is not being applied by default due to the has-image class.

Screenshot 2023-06-09 at 3 35 40 PM

I refactored the classNames function slightly and it seems to work:

```
private classNames() {
    const classNames = ['sage-avatar'];

    if (this.image) {
      classNames.push('sage-avatar--has-image');
    }
    if (this.variant) {
      classNames.push('sage-avatar--' + this.variant);
    }

    return classNames.join('  ');
  }
```

There may be a better adjustment or approach, but hopefully, this can provide some direction.

pixelflips (changes requested):
Found 2 additional issues and then I think it's ready to go!

libs/core/src/components/sage-avatar/sage-avatar.scss
```diff
@@ -0,0 +1,68 @@
+:host {
+  display: inline-block;
+}
+
+div {
+  // These custom props are not reachable
```
QuintonJason: 🏆 This is a great callout and something we'll need to revisit throughout the codebase

libs/core/src/components/sage-avatar/sage-avatar.scss
QuintonJason: +1

libs/core/src/components/sage-avatar/sage-avatar.tsx
QuintonJason: Same callout with the `sage-avatar--has-image` class being present on the Canvas tab demos

QuintonJason (changes requested):
Looking good. Left a couple comments

libs/core/src/components/sage-avatar/sage-avatar.scss
anechol: Thanks for this. I was looking at the border tokens and not the Avatar. Probably another thing to ask design about in regards to consistency.

libs/core/src/components/sage-avatar/sage-avatar.tsx
anechol: Thanks for this rec. This was a minor logical issue that I ended up pairing with Julian on.

anechol: Thanks for calling this out. This was a minor logical issue I paired with Julian on.

libs/core/src/components/sage-avatar/sage-avatar.tsx
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * The src for a custom user image.
+   * @defaultValue null
+   */
+  @Prop() image?: string | null = null;
+
+  /**
+   * Size of the avatar. Value can be preset or custom.
+   */
+  @Prop({ reflect: true }) size?:
```
pixelflips: Should we document the preset sizes in some way? The default displays in the props table but all the others do not.

pixelflips (approved):
One last minor non-blocking comment, otherwise LGTM! Really great work!! 👍🏼

libs/core/src/components/pds-avatar/assets/demi_wilkinson.png
ju-Skinner, line 1: This file needs to be removed

libs/core/src/components/pds-avatar/test/pds-avatar.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+    await page.setContent('<pds-avatar></pds-avatar>');
+
+    const element = await page.find('pds-avatar');
+    expect(element).toHaveClass('hydrated');
+  });
+
+  it('renders admin variant with correct class', async () => {
+    const page = await newE2EPage();
+
+    await page.setContent('<pds-avatar variant="admin"></pds-avatar>');
```
ju-Skinner, line 14: nitpick ⚠️ :  move the line up and have newline after this. Similar to what you have on lines 5-8. Apply to the rest of the file, please and thanks.

ju-Skinner (changes requested):
Changes requested are related to the `.png` has been added, when it should have been removed.

QuintonJason (changes requested):
Screen Shot 2023-06-13 at 11 43 58 AM

On the Dropdown `Canvas` example the size isn't coming though

ju-Skinner (approved):
When you merge this please "Squash and merge".

QuintonJason (approved):
Good to go!

PR #58, opened 2023-05-11
Tokens - Fixes Build Issue

anechol:
# Description

Fixes build function not using the proper argument. This caused an issue of not being able to properly build tokens when an object key was passed into the function call.

## Type of change


pixelflips (approved):
LGTM! 🚀

ju-Skinner (approved):

PR #80, opened 2023-06-15
[DSS-416]: feat(progress) - Progress

pixelflips:
# Description

Adds progress component to the system including:

- Default variant
- Animations
- Percentage-based filled state
- Custom fill colors
- Optional percent text

Fixes [DSS-416](https://kajabi.atlassian.net/browse/DSS-416)

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# How Has This Been Tested?

To test run the `yarn nx run core:test` command and verify coverage and passing results.

Manual tests were also performed.

- [x] unit tests
- [x] e2e tests
- [x] tested manually

**Test Configuration**:

- os: macOS 13.4

# Checklist:

[DSS-416]: https://kajabi.atlassian.net/browse/DSS-416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/src/components/pds-progress/pds-progress.scss
```diff
@@ -0,0 +1,78 @@
+:host {
+  --progress-fill-color: var(--pine-color-primary-300);
+  align-items: center;
```
QuintonJason, line 5: Can the custom props have an extra line between them and the css properties?

libs/core/src/components/pds-progress/pds-progress.scss
```diff
@@ -0,0 +1,78 @@
...
+
+.pds-progress-bar {
+  --background-progress-bar: var(--pine-color-neutral-grey-300);
+  --border-radius: var(--pine-border-radius-xs);
+  align-items: center;
+  background-color: var(--background-progress-bar);
+  border-radius: var(--border-radius);
+  display: flex;
+  flex-direction: row;
+  height: 8px;
```
QuintonJason: Can this repeated value be updated to a var?

libs/core/src/components/pds-progress/test/pds-progress.e2e.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  it('renders with a custom percent values', async () => {
+    const page = await newE2EPage();
+    await page.setContent('<pds-progress percent="25"></pds-progress>');
+
+    const element = await page.find('pds-progress >>> progress');
+    expect(element.getAttribute('value')).toBe('25');
+  });
+
+  it('renders with animated progress', async () => {
```
QuintonJason: These tests seem to be covered in the unit tests since a value isn't changing, just being checked. One e2e test I could see is the percentage being one value, then being updated to another

QuintonJason (changes requested):
Looking good! Added some comments

QuintonJason:
Screen Shot 2023-06-20 at 10 41 00 AM

There is also a need for the percentage to be displayed visually as a tooltip either above or below the progress bar that currently exists in our Sage implementation

pixelflips:
> There is also a need for the percentage to be displayed visually as a tooltip either above or below the progress bar that currently exists in our Sage implementation

@QuintonJason I didn't add it in this implementation since the tooltip is now a separate component which was not the case in Sage. I guess we could add an additional story but that felt like overkill to me but maybe not 🤷🏼 . Let me know what you think.

QuintonJason (changes requested):
There is also a need for the percentage to be displayed visually as a tooltip either above or below the progress bar that currently exists in our Sage implementation

@QuintonJason I didn't add it in this implementation since the tooltip is now a separate component which was not the case in Sage. I guess we could add an additional story but that felt like overkill to me but maybe not 🤷🏼 . Let me know what you think.

I don't think it's overkill since it already exists the designs and in dev for rails and react. It could be created similar to that part in the React implementation. I didn't want to introduce a situation going into a newer system, we don't have things they would expect to exist

pixelflips:
> I don't think it's overkill since it already exists the designs and in dev for rails and react. It could be created similar to that part in the React implementation. I didn't want to introduce a situation going into a newer system, we don't have things they would expect to exist

@QuintonJason I just meant overkill in the sense that we don't need to support the implementation of the tooltip inside of progress like the previous system. We can just use the tooltip and add an example/link on the progress page for clarity. See the screenshot as an example of what I am thinking:

Screenshot 2023-06-20 at 10 26 29 AM

Let me know if that clarifies and helps to better explain what I was trying to say.

QuintonJason (approved):
Looks good

ju-Skinner (approved):

PR #82, opened 2023-06-28
[DSS-431]: feat(copytext) - Copy Text

pixelflips:
# Description

Adds the CopyText component to the system including:

- Default UI
- Border options
- Full-width options
- Truncation options

Button:
- Adds `unstyled` variant to allow the button component to be used to compose the copyText component.

Fixes [DSS-431](https://kajabi.atlassian.net/browse/DSS-431)

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [x] tested manually
- [x] other:

**Test Configuration**:

- os: macOS 13.4.1
- browsers: Chrome(latest), Edge(latest), Firefox(latest), Safari(latest)

# Checklist:

[DSS-431]: https://kajabi.atlassian.net/browse/DSS-431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/components/pds-button/test/pds-button.spec.tsx
+++ b/libs/core/src/components/pds-button/test/pds-button.spec.tsx
@@ -34,6 +34,22 @@ describe('pds-button', () => {
     `);
   });
 
+  it('renders unstyled button', async () => {
+    const {root} = await newSpecPage({
+      components: [PdsButton],
+      html: `<pds-button variant="unstyled"></pds-button>`,
+    });
+    expect(root).toEqualHtml(`
+      <pds-button variant="unstyled">
+        <mock:shadow-root>
+          <button class="pds-button pds-button--unstyled" type="button">
+            <slot></slot>
+          </button>
+        </mock:shadow-root>
+      </pds-button>
+    `);
+  });
+
   it('renders disabled button', async () => {
     const {root} = await newSpecPage({
       components: [PdsButton],
@@ -55,7 +71,7 @@ describe('pds-button', () => {
       components: [PdsButton],
       html: `<pds-button icon="trashIcon"></pds-button>`,
     });
-    const svg = root.shadowRoot.querySelector('svg');
+    const svg = root?.shadowRoot?.querySelector('svg');
     expect(svg).not.toBeNull();
   });
 
@@ -71,7 +87,7 @@ describe('pds-button', () => {
     const form = root.doc.querySelector<HTMLFormElement>("form");
     const eventSpy = jest.fn();
     form?.addEventListener("reset", eventSpy());
-    const button = document.querySelector<HTMLPdsButtonElement>('pds-button');
+    const button = document.querySelector<HTMLButtonElement>('pds-button');
     button?.click();
     await root.waitForChanges();
     expect(eventSpy).toHaveBeenCalled();
--- a/libs/core/src/components/pds-copytext/pds-copytext.tsx
+++ b/libs/core/src/components/pds-copytext/pds-copytext.tsx
@@ -17,7 +17,7 @@ export class PdsCopytext {
   /**
    * The string that is displayed and that is also copied to the clipboard upon interaction.
    */
-  @Prop() value: string;
+  @Prop() value!: string;
 
   async copyToClipboard(value: string) {
     try {
--- a/libs/core/src/components/pds-copytext/pds-copytext.tsx
+++ b/libs/core/src/components/pds-copytext/pds-copytext.tsx
@@ -6,8 +6,17 @@ import { Component, Host, h, Prop } from '@stencil/core';
   shadow: true,
 })
 export class PdsCopytext {
+  /**
+   * Determines whether copytext should have a border.
+   */
   @Prop() border = true;
+  /**
+   * Determines whether copytext should expand to the full width of its container.
+   */
   @Prop() fullWidth = false;
+  /**
+   * The string that is displayed and that is also copied to the clipboard upon interaction.
+   */
   @Prop() value: string;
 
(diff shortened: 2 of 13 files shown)


libs/core/src/components/pds-copytext/test/pds-copytext.spec.tsx
```diff
@@ -0,0 +1,149 @@
+import { newSpecPage } from '@stencil/core/testing';
```
pixelflips, line 1: > **Note**

Coverage is currently less than desired, but this is due to a limitation with jest and the `document.execCommand` used in the fallback code. Everything else has coverage except for the safari fallback but the fallback is covered in the e2e test.

pixelflips:
> **Note**
> Spec test coverage is currently less than desired, but this is due to a limitation with jest and the `document.execCommand` used in the fallback for Safari. Everything else has coverage and the fallback is covered in the e2e test.

libs/core/src/components/pds-button/test/pds-button.spec.tsx
```diff
@@ -71,7 +87,7 @@ describe('pds-button', () => {
     const form = root.doc.querySelector<HTMLFormElement>("form");
     const eventSpy = jest.fn();
     form?.addEventListener("reset", eventSpy());
-    const button = document.querySelector<HTMLPdsButtonElement>('pds-button');
+    const button = document.querySelector<HTMLButtonElement>('pds-button');
```
ju-Skinner, line 90: Was there a reason for this change?

anechol (approved):
Looks awesome! Love that we're using an unstyled button to build this instead of a new button component altogether. 
One non-blocking thing: could we add a style for changing the cursor on hover? Currently it's difficult to tell if the button is interactable.

pixelflips:
> Looks awesome! Love that we're using an unstyled button to build this instead of a new button component altogether. One non-blocking thing: could we add a style for changing the cursor on hover? Currently it's difficult to tell if the button is interactable.

@anechol I have no problem updating the cursor, but I feel like that should be controlled via the button component which currently does not have the cursor adjustments. Let me know your thoughts.

libs/core/src/components/pds-copytext/pds-copytext.tsx
```diff
@@ -0,0 +1,98 @@
...
+@Component({
+  tag: 'pds-copytext',
+  styleUrl: 'pds-copytext.scss',
+  shadow: true,
+})
+export class PdsCopytext {
+  /**
+   * Determines whether `copytext` should have a border.
+   * @defaultValue true
+   */
```
ju-Skinner, line 12: Do we want to reflect this in the DOM?

E.g `{ reflect: true }`

libs/core/src/components/pds-copytext/stories/pds-copytext.docs.stories.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+import { extractArgTypes } from '@pxtrn/storybook-addon-docs-stencil';
+
+import * as stories from './pds-copytext.stories.js';
+
+<Meta title="components/CopyText" component="pds-copytext" argTypes={extractArgTypes('pds-copytext')} />
+
+# Copy Text
+
+A small component to use in places where "copy" text is provided to be copied elsewhere.
```
ju-Skinner: Can we remove the word `small`?

ju-Skinner (approved):
I didn't see anything blocking. Comments left are non-blocking.

libs/core/src/components/pds-copytext/stories/pds-copytext.docs.stories.mdx
pixelflips: Done.

libs/core/src/components/pds-copytext/pds-copytext.tsx
pixelflips: I am sure I am totally misunderstanding here, but would using the "bordered" class name be another way to reflect it in the DOM?

libs/core/src/components/pds-button/test/pds-button.spec.tsx
pixelflips: The HTMLPdsButtonElement was showing as: 

`Cannot find name 'HTMLPdsButtonElement'.`

Updating resolved that. Let me know if there may be an alternative way to go about it.

anechol:
> > Looks awesome! Love that we're using an unstyled button to build this instead of a new button component altogether. One non-blocking thing: could we add a style for changing the cursor on hover? Currently it's difficult to tell if the button is interactable.
> 
> @anechol I have no problem updating the cursor, but I feel like that should be controlled via the button component which currently does not have the cursor adjustments. Let me know your thoughts.

@pixelflips As I'm poking my head around in research, it looks like the cursor change is not required for buttons as that feedback is traditionally used for links only (TIL!). We should be good if the visual feedback is strong enough to indicate that the button is being hovered, but given the change in border and background, it leaves me wondering if the current change we have is strong enough when considering users with low vision 🤔 
This is probably a conversation to have with design at some point.

I'm fine with this being merged.

QuintonJason (approved):
Good stuff

libs/core/src/components/pds-copytext/pds-copytext.tsx
pixelflips: @ju-Skinner after our convo, I added and double-checked. I totally see what you meant and that's dope! Appreciate the clarification!

PR #92, opened 2023-08-16
[DSS-440]: feat(sortable) - Sortable Component

pixelflips:
# Description

Adds the Sortable component and SortableItem subcomponent to the system including:

- Default UI
- Border option
- Divider option
- Handle options
- Actions: SortableItem

Fixes [DSS-440: Sortable](https://kajabi.atlassian.net/browse/DSS-440)

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [x] tested manually
- [x] other:

**Test Configuration**:

- os: macOS 13.5.1

# Checklist:


libs/core/src/components/pds-sortable-item/pds-sortable-item.scss
```diff
@@ -0,0 +1,79 @@
+:host(.pds-sortable-item) {
+  --background-color-interactive: var(--pine-color-neutral-grey-200);
+  --background-color: var(--pine-color-base-white);
+  --border-radius: var(--pine-border-radius-md);
+  --border: var(--pine-border-default);
+  --box-shadow: var(--pine-box-shadow-lg);
+  --color-interactive-handle: var(--pine-color-primary-300);
+  --margin-inline-end: var(--pine-spacing-md); //24px
```
QuintonJason: Regarding the comments, do they need to be here for some and not all? Are these lines commented with the value because it is spacing related?

libs/core/src/components/pds-sortable-item/pds-sortable-item.scss
```diff
@@ -0,0 +1,79 @@
...
+  --padding-inline: var(--pine-spacing-md); //24px
+
+  align-items: center;
+  display: flex;
+  flex-wrap: wrap;
+  padding-block: var(--padding-block-xs);
+  padding-inline: var(--padding-inline);
+
+  .pds-sortable-item__actions {
+    margin-left: auto
```
QuintonJason: Use a logical prop here

libs/core/src/components/pds-sortable-item/pds-sortable-item.scss
```diff
@@ -0,0 +1,79 @@
...
+
+  .pds-sortable-item__handle {
+    line-height: 1;
+    margin-inline-end: var(--margin-inline-end);
+  }
+
+  // styles when sortable is bordered
+  .pds-sortable--bordered & {
+    &:first-child {
+      border-top-left-radius: var(--border-radius);
```
QuintonJason: Use a logical prop for the border-radius

libs/core/src/components/pds-sortable-item/pds-sortable-item.scss
```diff
@@ -0,0 +1,79 @@
...
+
+    &:last-child {
+      border-bottom-left-radius: var(--border-radius);
+      border-bottom-right-radius: var(--border-radius);
+    }
+  }
+
+  // styles when sortable is divided
+  .pds-sortable--divided & {
+    border-bottom: var(--border);
```
QuintonJason: Use a logical prop for the border

libs/core/src/components/pds-sortable-item/test/pds-sortable-item.e2e.ts
```diff
@@ -0,0 +1,11 @@
+import { newE2EPage } from '@stencil/core/testing';
```
QuintonJason: This e2e test seems to be covered by the first spec test. Can this be removed?

libs/core/src/components/pds-sortable/pds-sortable.tsx
```diff
@@ -0,0 +1,83 @@
...
+   */
+  @Prop() componentId!: string;
+
+  /**
+   * Deternines whether `sortable` items should be divided with border.
+   */
+  @Prop({ reflect: true }) dividers = false;
+
+  /**
+   * Assigns the handle type to use for sortable items.
```
QuintonJason: Can this be updated to speak more to what the type will affect?

Something like 'Determines the clickable area for the `pds-sortable-item`'. What are your thoughts on this?

libs/core/src/components/pds-sortable/pds-sortable.tsx
```diff
@@ -0,0 +1,83 @@
...
+   */
+  @Prop() handleType: 'handle' | 'row' = 'row';
+
+  /**
+   * Event emitted when a sortable item is moved.
+   */
+  @Event() pdsSortableItemMoved: EventEmitter;
+
+  private container: HTMLElement;
```
QuintonJason: I'm not sure if this is needed, but all components are allowed on `@Element` as this one seems to fit. Does using `@Element` vs the private var break anything in the code or tests or is that not needed for you implementation?

libs/core/src/components/pds-sortable/stories/pds-sortable.docs.mdx
```diff
@@ -0,0 +1,227 @@
...
+</pds-sortable>
+`}>
+  <pds-sortable component-id="handle">
+    <pds-sortable-item handle>Item 1</pds-sortable-item>
+    <pds-sortable-item handle>Item 2</pds-sortable-item>
+    <pds-sortable-item handle>Item 3</pds-sortable-item>
+  </pds-sortable>
+</Canvas>
+
+#### Handle Only
```
QuintonJason: Can this be labeled `handleType` or handle vs row?

libs/core/src/components/pds-sortable-item/pds-sortable-item.tsx
```diff
@@ -0,0 +1,46 @@
...
+  tag: 'pds-sortable-item',
+  styleUrl: 'pds-sortable-item.scss',
+  scoped: true,
+})
+export class PdsSortableItem {
+  /**
+   * Determines whether `sortable-item-actions` slot should be enabled.
+   * @defaultValue false
+   */
+  @Prop() actions = false;
```
QuintonJason: What are your thoughts on having the word `has` prepended to props that are actually conditionals for the existence of slot content? We currently don't have a rule for it, but it does seem like an opportunity for this type of naming convention. Something like `hasActions`

QuintonJason (changes requested):
This is great work! Left a couple comments

QuintonJason:
Screen_Shot_2023-08-30_at_10_03_56_AM

Calling out 3 items in this pic:

1. When the description wraps, the entire description goes on the next line
2. When the description wraps, the actions slot goes to the next line
3. When multiple items exist within the `actions`, there is no spacing between them

libs/core/src/components/pds-sortable-item/pds-sortable-item.scss
pixelflips: Will remove, these were notes to self while developing to make sure I had the correct values. Good catch.

libs/core/src/components/pds-sortable-item/pds-sortable-item.scss
pixelflips: Done

libs/core/src/components/pds-sortable-item/pds-sortable-item.scss
pixelflips: Done

libs/core/src/components/pds-sortable-item/pds-sortable-item.scss
pixelflips: Done

libs/core/src/components/pds-sortable/pds-sortable.tsx
pixelflips: Updated.

libs/core/src/components/pds-sortable/stories/pds-sortable.docs.mdx
pixelflips: Updated and included both.

libs/core/src/components/pds-sortable-item/pds-sortable-item.tsx
pixelflips: I don't hate it, but I would say we should add it to the housekeeping doc, and if we all agree we can revisit and update all components.

pixelflips:
> Screen_Shot_2023-08-30_at_10_03_56_AM
> Calling out 3 items in this pic:
> 
> 1. When the description wraps, the entire description goes on the next line
> 2. When the description wraps, the actions slot goes to the next line
> 3. When multiple items exist within the `actions`, there is no spacing between them


1 & 2 should be resolved. 

As for 3, I think this brings up a larger conversation around slots/layout in general. In my opinion, the spacing of elements in the slot would be handled by whatever elements/content is put inside that slot. 

Also, I will prob need a follow-up here once we have the layout and breakpoints in place.

Let me know if you have any additional thoughts!

libs/core/src/components/pds-sortable-item/test/pds-sortable-item.e2e.ts
pixelflips: It can be removed, but I feel like at least one simple/default e2e test that uses an actual browser is better than nothing at all. I could definitely be wrong here though so let me know!

libs/core/src/components/pds-sortable-item/pds-sortable-item.tsx
pixelflips: Okay after more thinking on this one, I do think I should rename it, but to something like `enableActions` since it's not an actual check if actions exist, it just enables the slot.

libs/core/src/components/pds-sortable-item/test/pds-sortable-item.e2e.ts
QuintonJason: There may not necessarily be a need for all components to have an e2e test file, since not every component will have a flow that needs to be checked.

libs/core/src/components/pds-sortable-item/pds-sortable-item.tsx
anechol: I agree that we should have a conversation on this. I would love if we can continue to have consistent naming conventions.

anechol (approved):
This is awesome work here. I love the idea of adding a full demo example to the stories. On a non-blocking note, since we are including the sortable item as a subcomponent, would it make sense to move it inside the root sortable folder much like we do with tab and tabpanel inside tabs?

pixelflips:
@anechol I've updated the folder structure and all seems to be a-ok! Thanks for the suggestion!

QuintonJason (approved):
Nice work

libs/core/src/components/pds-sortable/pds-sortable-item/test/pds-sortable-item.e2e.ts
ju-Skinner, line 1: Do we need this test? Or could we add more tests to this? It could be things as simple as checking various attributes based on props being set.  

We can discuss offline and circle back to put the outcome here if need be.

libs/core/src/components/pds-sortable/pds-sortable.tsx
ju-Skinner: > I'm not sure if this is needed, but all components are allowed on `@Element` as this one seems to fit. Does using `@Element` vs the private var break anything in the code or tests or is that not needed for you implementation?



I noticed [this](https://github.com/Kajabi/pine/pull/92/files#r1310412576) was resolved...what was the response and outcome to @QuintonJason question?

ju-Skinner (changes requested):
First off let me start by saying "GREAT JOB!!". 🎩 I know you had some struggles working through this but you got to the finish line. There is only one reason why I rejected it. See below. 

**Rejected** due to `pds-sortable-item.e2e.ts`. Either we should add more specs or just remove the file altogether.

For additional comments, you can find them below in their open comment boxes. 

**Non-Blocking**: Another concern with this PR is the dependency on Sortable and what it does to the package size. But this is something we can look at later.

ju-Skinner:
When we get ready to merge this. lets use squash and merge. Make the title of the commit to be something like `feat(pds-sortable): implementation to both pds-sortbable and pds-sortable-item`

libs/core/src/components/pds-sortable/pds-sortable.tsx
pixelflips: I did run into some issues, but don't recall exactly what. Let me give it another look.

libs/core/src/components/pds-sortable/pds-sortable-item/test/pds-sortable-item.e2e.ts
pixelflips: Since this is the second mention of this, we can remove it. I believe the attributes will be covered in the spec test.

libs/core/src/components/pds-sortable/pds-sortable.tsx
pixelflips: Ok not entirely sure what I was doing wrong in the previous attempt, but updated and all seems fine. Thanks for catching this!

pixelflips: @QuintonJason great catch here! I have made the adjustments and all should be good now!

:hat-tip to @ju-Skinner as well for help specifically related to this area!

ju-Skinner (approved):
Great job. Thanks for your patience and persistence.

Mark-Miller-Kajabi:
Please feel free to merge. 

The MPL license for Axe-core has been in the REPO and has been vetted as being ok in this repository by legal and infosec. 

I would like to see the critical vulnerability addressed, but i do understand it is a Dependency of a dependency making it tough.

PR #93, opened 2023-08-22
chore: update icons from latest version of Figma file

ju-Skinner:
# Description
Ran the Figma export process to get the latest icons.


libs/icons/changelogs/2023-08-22-changelog.html
```diff
@@ -0,0 +1,173 @@
+<!doctype html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>2023-08-22 - {{version}} Changelog</title>
```
pixelflips, line 5: Not a blocker, but `{{version}}` outputs as text.

pixelflips (approved):
One tiny non-blocking comment but overall LGTM! 👍🏼 Great work!

anechol (approved):
🚀

libs/icons/changelogs/2023-08-22-changelog.html
ju-Skinner: Yeah, that's still in the works. Which will happen during nightly builds.

PR #94, opened 2023-08-22
chore: upgrade storybook to v7.3

ju-Skinner:
# Description

Upgrades to the latest version of Storybook.


anechol (approved):
Smooth upgrade and all stories work great!

pixelflips (approved):
🔥🔥🔥 - Totally digging how the docs section is now included and the default in the sidebar. Thanks for handling this upgrade! LGTM!

libs/core/src/components/pds-tabs/stories/pds-tabs.docs.mdx
```diff
@@ -1,19 +1,11 @@
...
 
 import * as stories from './pds-tabs.stories.js';
 
-<Meta
-  title="components/Tabs"
-  component="pds-tabs"
-  argTypes={extractArgTypes('pds-tabs')}
-  subcomponents={{
-    PdsTab: 'pds-tab',
-    PdsTabPanel: 'pds-tabpanel',
```
pixelflips, line 14: I did notice the subcomponent tabs are no longer showing up on the tabs documentation. Is this intentional or due to upgrading? Not a blocker, but would be great to have those if possible.

ju-Skinner: Yes, @pixelflips there is some funky behavior with subcomponents in the latest version. Let me see what I can dig back up.

pixelflips: Ok cool, can def be handled in a follow-up if needed. Thanks for checking though!

ju-Skinner: Here is the information that I obtained when I was going through each of the components and ran into `subcomponents`.

https://github.com/storybookjs/storybook/issues/20782#issuecomment-1482771013

pixelflips: Thanks for sharing! I have also used the subcomponents on the sortable, so I will check out the info, but if you have any suggestions let me know!

PR #103, opened 2023-09-11
DSS-467: chore:(input) - Housekeeping

pixelflips:
# Description

Housekeeping-related adjustments to input component:

- Remove CSS props exposed in props table
- Updates colors, spacing, fonts, etc to use tokens
- Corrects disabled state font colors
- Adjust placeholder story/docs to display placeholder text
- Adds MDN link to additional resources section
- Documentation adjustments

Fixes [DSS-467](https://kajabi.atlassian.net/browse/DSS-467)

## Type of change

- [x] Chore (changes that don't affect source or test files)

# How Has This Been Tested?

Navigate to the Input component. Verify changes applied and that the input and documentation render as expected

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-467]: https://kajabi.atlassian.net/browse/DSS-467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/components.d.ts
+++ b/libs/core/src/components.d.ts
@@ -233,39 +233,39 @@ export namespace Components {
          */
         "componentId": string;
         /**
-          * Indicates whether or not the input field is disabled
+          * Determines whether or not the input field is disabled.
          */
         "disabled"?: boolean;
         /**
-          * Specifies the error message and provides an error-themed treatment to the field
+          * Specifies the error message and provides an error-themed treatment to the field.
          */
         "errorMessage"?: string;
         /**
-          * Displays a message or hint below the input field
+          * Displays a message or hint below the input field.
          */
         "helperMessage"?: string;
         /**
-          * Indicates whether or not the input field is invalid or throws an error
+          * Determines whether or not the input field is invalid or throws an error.
          */
         "invalid"?: boolean;
         /**
-          * Text to be displayed as the input label
+          * Text to be displayed as the input label.
          */
         "label"?: string;
         /**
-          * Specifies the name. Submitted with the form name/value pair
+          * Specifies the name. Submitted with the form name/value pair.
          */
         "name"?: string;
         /**
-          * Specifies a short hint that describes the expected value of the input field
+          * Specifies a short hint that describes the expected value of the input field.
          */
         "placeholder"?: string;
         /**
-          * Indicates whether or not the input field is readonly
+          * Determines whether or not the input field is readonly.
          */
         "readonly"?: boolean;
         /**
-          * Indicates whether or not the input field is required
+          * Determines whether or not the input field is required.
          */
         "required"?: boolean;
         /**
@@ -274,7 +274,7 @@ export namespace Components {
          */
         "type": string;
         /**
-          * The value of the input
+          * The value of the input.
          */
         "value"?: string;
     }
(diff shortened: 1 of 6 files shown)


libs/core/src/components.d.ts
```diff
@@ -233,39 +233,39 @@ export namespace Components {
          */
         "componentId": string;
         /**
-          * Indicates whether or not the input field is disabled
+          * Determines whether or not the input field is disabled.
```
ju-Skinner: The word "determines" seems to imply an automatic or programmatic setting of the value. Can we discuss a better way to describe our boolean values? I want to note that I don't think we should have a blanket rule for all boolean descriptions.

ju-Skinner (changes requested):
Team discussion is needed. See the comment left below.

libs/core/src/components.d.ts
pixelflips: "Determines" was decided by a previous team discussion, but if we want to revisit I'd say let's do it and then make the updates with whatever decision is decided in a follow-up ticket/pr.

anechol (approved):

libs/core/src/components.d.ts
ju-Skinner: I don’t remember us having a full team discussion on this, but I do remember our brief chat about the difference between ‘Indicates’ and ‘Determines’.

We should not merge this but rather, let’s discuss the details now to ensure clarity moving forward.

pixelflips: And just "whether or not" terminology before that... clarification is fine, but since this is scoped to the input component, I'd disagree with holding this up since it may eventually affect a wide variety of components (~12).

I'd be happy to roll back the text change here and wait for the team discussion if that works better?

monicawheeler: It would be ideal to address the phrasing like, "If true... If false"

pixelflips: @ju-Skinner updated after team convo. Check it out when time permits.

libs/core/src/components.d.ts
```diff
@@ -233,39 +233,39 @@ export namespace Components {
          */
         "componentId": string;
         /**
-          * Indicates whether or not the input field is disabled
+          * If true, the input field is disabled.
```
monicawheeler: Could we say something more explanatory, like this
```suggestion
          * If true, the user cannot interact with the field.
```

libs/core/src/components.d.ts
```diff
@@ -233,39 +233,39 @@ export namespace Components {
...
          */
         "errorMessage"?: string;
         /**
-          * Displays a message or hint below the input field
+          * Displays a message or hint below the input field.
          */
         "helperMessage"?: string;
         /**
-          * Indicates whether or not the input field is invalid or throws an error
+          * If true, the input field is invalid.
```
monicawheeler: What does it mean for the input field to be invalid? Does this change the appearance? Can the user interact with it? Do we know if it triggers a message? 

Can we be more descriptive here so that it's clear the treatment when the input field is invalid? Like, "If true, the input field no longer accepts alien emojis"

libs/core/src/components.d.ts
```diff
@@ -233,39 +233,39 @@ export namespace Components {
...
          */
         "name"?: string;
         /**
-          * Specifies a short hint that describes the expected value of the input field
+          * Specifies a short hint that describes the expected value of the input field.
          */
         "placeholder"?: string;
         /**
-          * Indicates whether or not the input field is readonly
+          * If true, the input field is readonly.
```
monicawheeler: Same sort of thing with "invalid" - what does readonly do to the field? Is it available to screenreaders?

libs/core/src/components.d.ts
```diff
@@ -233,39 +233,39 @@ export namespace Components {
...
          */
         "readonly"?: boolean;
         /**
-          * Indicates whether or not the input field is required
+          * If true, the input field is required.
```
monicawheeler: Also, same here with the required boolean. Does this visually apply styles when required? Does it prevent the field from being submitted, therefore stopping form submission?

monicawheeler (changes requested):
Hey there, Phillip! I added some comments around the "If true..." definitions as I still felt like I didn't understand what occurred with each boolean. Let me know if you want to chat on the comments. Thanks!

pixelflips:
@monicawheeler all great questions and suggestions. I'd propose that those go into the [follow-up task](https://kajabi.atlassian.net/browse/DSS-469) since it seems that there are still questions. 

We can then reboot/document any decisions in Jira vs. this input-related housekeeping PR. 

Thoughts?

monicawheeler:
> @monicawheeler all great questions and suggestions. I'd propose that those go into the follow up task since it seems that there are still questions.
> 
> We can then reboot/document any decisions in Jira vs this input related housekeeping PR.
> 
> Thoughts?

Sounds good. Let's remove the updates to the wording on this PR and use the new Jira task for the updates to copy.

monicawheeler (approved):
Thanks for being open to feedback on the boolean documentation! Looks good 👍

ju-Skinner (approved):

PR #112, opened 2023-09-29
DSS-408: feat(table): Table Component

pixelflips:
# Description

Adds the table component to the system including:

- Default UI
- Compact option
- Responsive option
- Selectable option (checkbox)
- Fixed column option
- Basic column sorting

**Checkbox:**

- Adds `label-hidden` prop for allowing checkbox labels to be visually hidden.

Fixes [DSS-408](https://kajabi.atlassian.net/browse/DSS-408)

|  Screenshot  |
|--------|
|Screenshot 2023-11-22 at 1 10 25 PM|

## Type of change

# How Has This Been Tested?
To test run the `npx nx run core:test` command and verify the passing results.

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-408]: https://kajabi.atlassian.net/browse/DSS-408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/src/components/pds-table/pds-table-checkbox-cell/pds-table-checkbox-cell.tsx
```diff
@@ -0,0 +1,17 @@
...
+export class PdsTableCheckboxCell {
+
+  render() {
+    return (
+      <Host role="gridcell">
+        <slot></slot>
+      </Host>
+    );
+  }
+}
```
ju-Skinner: Could we get rid of this and just have a `pds-table-cell` with `pds-checkbox` being used? A few reasons for asking my question.

1. I see you modified `pds-checkbox`, assuming it was for the use of `pds-table` in some fashion

2. This component seems to be used only in a small instance. Which I don't believe we need to create a specialized component for it.

libs/core/src/components/pds-table/pds-table-cell/stories/pds-table-cell.stories.js
```diff
@@ -0,0 +1,24 @@
...
+export default {
+  argTypes: extractArgTypes('pds-table-cell'),
+  component: 'pds-table-cell',
+  decorators: [withActions],
+  title: 'components/Table/Cells',
+};
+
+const BaseTemplate = (args) => html`
+<pds-table>
+  <pds-table-row>
```
ju-Skinner: I don't see the use of `pds-table-body` here, so is it (`pds-table-body`) something that we need?

libs/core/src/components/pds-table/pds-table-cell/test/pds-table-cell.e2e.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const element = await page.find('pds-table-cell');
+    expect(element).toHaveClass('hydrated');
+  });
+
+  it('renders with compact styles', async () => {
+    const page = await newE2EPage();
+    await page.setContent(`
+      <pds-table compact>
+        <pds-table-body>
```
ju-Skinner, line 16: Piggybacking on my comment above, is there a reason you used it here, but not in the above example?

libs/core/src/components/pds-table/pds-table-cell/test/pds-table-cell.e2e.ts
```diff
@@ -0,0 +1,44 @@
...
+          <pds-table-row>
+            <pds-table-cell truncate>Very long content that should be truncated</pds-table-cell>
+          </pds-table-row>
+        </pds-table-body>
+      </pds-table>
+    `);
+
+    const tableCell = await page.find('pds-table-cell');
+    expect(tableCell).toHaveClass('is-truncated');
```
ju-Skinner, line 42: Add an expectation that will show that the actual text is truncated.

libs/core/src/components/pds-table/pds-table-head-cell/pds-table-head-cell.tsx
```diff
@@ -0,0 +1,105 @@
...
+      this.tableScrolling = true;
+    } else {
+      this.tableScrolling = false;
+    }
+  };
+
+  private handleSortClick = () => {
+    if (this.sortable) {
+      const column = this.hostElement.innerText.trim();
+      this.sortingDirection = this.sortingDirection === 'asc' ? 'desc' : 'asc';
```
ju-Skinner, line 46: Is this right? The code below is how it reads. 

```
If (this.sortingDirection === 'asc')
  this.sortingDirection 'desc';
else
  this.sortingDirection =  'asc';
end
```

Is this function's purpose to `toggle` the direction? If so, let's rename it as such `toggleSort`

libs/core/src/components/pds-table/pds-table-head-cell/test/pds-table-head-cell.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const element = await page.find('pds-table-head-cell');
+    expect(element).toHaveClass('hydrated');
+  });
+
+  it('renders with fixed styles when part attribute is set', async () => {
+    const page = await newE2EPage();
+    await page.setContent(`
+      <pds-table fixed-column>
+        <pds-table-head>
```
ju-Skinner, line 16: no use of `pds-table-body`. Do we want to be consistent or look into making sure the schema requires the use `pds-table-body`?

libs/core/src/components/pds-table/pds-table-head-cell/stories/pds-table-head-cell.stories.js
```diff
@@ -0,0 +1,24 @@
...
+export default {
+  argTypes: extractArgTypes('pds-table-head-cell'),
+  component: 'pds-table-head-cell',
+  decorators: [withActions],
+  title: 'components/Table/Head',
+};
+
+const BaseTemplate = (args) => html`
+<pds-table>
+  <pds-table-head>
```
ju-Skinner, line 14: no use of pds-table-body. Do we want to be consistent or look into making sure the schema requires the use pds-table-body?

libs/core/src/components/pds-table/pds-table-head/pds-table-head.tsx
```diff
@@ -0,0 +1,31 @@
...
+      const tableCell = this.hostElement.querySelector('pds-table-head-cell:first-child');
+      tableCell?.classList.add("is-fixed");
+    }
+  }
+  render() {
+    return (
+      <Host role="row">
+        {this.tableRef && this.tableRef.selectable && (
+          <pds-table-checkbox-cell part={this.tableRef.fixedColumn ? 'checkbox-cell' : ''}>
+          </pds-table-checkbox-cell>
```
ju-Skinner: This is where i think we can change this to become something like the example below.

```
<pds-table-cell part={this.tableRef.fixedColumn ? 'checkbox-cell': '')>
  <pds-checkbox labelHidden />
</pds-table-cell>
```

libs/core/src/components/pds-table/pds-table-head/test/pds-table-head.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+      </pds-table-head>
+    `);
+  });
+
+  it('renders with is-fixed class when tableRef has fixedColumn', async () => {
+    const page = await newSpecPage({
+      components: [PdsTableHead, PdsTable],
+      html: `
+        <pds-table fixed-column="true">
+          <pds-table-head>
```
ju-Skinner, line 25: Same comment regarding the lack of `pds-table-body`.

libs/core/src/components/pds-table/pds-table-row/pds-table-row.tsx
```diff
@@ -0,0 +1,77 @@
...
+      tableCell?.classList.add("is-fixed");
+    }
+  }
+
+  private generateUniqueId = () => {
+    const randomString = Math.random().toString(36).substring(2, 8);
+    const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
+    const uniqueId = `${randomString}-${timestamp}`;
+
+    return uniqueId;
```
ju-Skinner, line 53: I like the thought, but not sure we to write a custom function, when we can use `UUID`.

libs/core/src/components/pds-table/stories/pds-table.stories.js
```diff
@@ -0,0 +1,227 @@
...
+const BaseTemplate = (args) => html`
+<pds-table
+  compact="${args.compact}"
+  component-id="${args.componentId}"
+  fixed-column="${args.fixedColumn}"
+  responsive="${args.responsive}"
+  selectable="${args.selectable}"
+>
+  <pds-table-head>
+    <pds-table-head-cell class="${args.compact ? 'hydrated is-compact' : 'hydrated'}">Column Title</pds-table-head-cell>
```
ju-Skinner: Does `is-compact` need to be passed to each table cell manually like this? If so, why? 

1. Could we find a way to improve this for the developer experience? 
2. Are there instances where a single cell may not use `is-compact`?

libs/core/src/components/pds-table/stories/pds-table.stories.js
```diff
@@ -0,0 +1,227 @@
...
+    <pds-table-head-cell class="${args.compact ? 'hydrated is-compact' : 'hydrated'}">Column Title</pds-table-head-cell>
```
ju-Skinner: Isn't `hydrated` used and set by the stencil-core runtime? I don't think we need to have that a part of the Template

ju-Skinner (changes requested):
This was my first pass. There are a lot of comments, they should make sense. If something is unclear, reply with your questions and I'll do my best to clear things up.

The request for the changes is due to some inconsistencies around the use of `pds-table-body` and I would like to see an attempt to remove `pds-table-checkbox-cell` in favor of using `pds-table-cell` and `pds-checkbox`.

libs/core/src/components/pds-table/pds-table-cell/test/pds-table-cell.e2e.ts
pixelflips: The text is only visually truncated using CSS with the class. How else would I accomplish that case?

libs/core/src/components/pds-table/pds-table-checkbox-cell/pds-table-checkbox-cell.tsx
pixelflips: 1. Updated checkbox to allow the label to be visually hidden.
2. Will look into this one.

libs/core/src/components/pds-table/pds-table-cell/stories/pds-table-cell.stories.js
pixelflips: Just missed it here. Will add.

libs/core/src/components/pds-table/pds-table-cell/test/pds-table-cell.e2e.ts
pixelflips: Just missed it above.

libs/core/src/components/pds-table/pds-table-row/pds-table-row.tsx
pixelflips: Can you elaborate since you previously suggested this approach? We wrote it together.

libs/core/src/components/pds-table/pds-table-head/test/pds-table-head.spec.tsx
pixelflips: This is a test for the table head, not sure why I need to include the body?

libs/core/src/components/pds-table/stories/pds-table.stories.js
pixelflips: This was an attempt to get the switches in storybook to work. The code works properly but the switch in the storybook controls does not. Should I remove and we can revisit the controls later?

libs/core/src/components/pds-table/pds-table-cell/test/pds-table-cell.e2e.ts
ju-Skinner: No need to if it's just "visually" hidden.

libs/core/src/components/pds-table/pds-table-row/pds-table-row.tsx
ju-Skinner: I don't recall assisting with this, it's probably not worth the trouble to add another NPM dependency. Please feel free to disregard this comment / thread

libs/core/src/components/pds-table/stories/pds-table.stories.js
ju-Skinner: Yeah, we can revisit before merging into the base branch.

libs/core/src/components/pds-table/stories/pds-table.stories.js
pixelflips: It does not, this was an attempt to get storybook controls to work. Will mark this one as resolved based on the convo below. The compact only has to be added to the table and will handle all the cells.

libs/core/src/components/pds-table/pds-table-checkbox-cell/pds-table-checkbox-cell.tsx
pixelflips: I was able to adjust and remove the need for the `pds-table-checkbox-cell`. Update incoming.

libs/core/src/components/pds-table/pds-table-head/pds-table-head.tsx
pixelflips: Was able to remove the need for the checkbox cell here. Marking as resolved.

libs/core/src/components/pds-table/pds-table-head-cell/pds-table-head-cell.tsx
pixelflips: This toggles the direction of the head cell arrow and emits the event. Will rename.

libs/core/src/components/pds-table/pds-table-head-cell/test/pds-table-head-cell.e2e.ts
pixelflips: This is a test for the table head cell, not sure why I need to include the body?

libs/core/src/components/pds-table/pds-table-head-cell/stories/pds-table-head-cell.stories.js
pixelflips: This is a story for the table head cell, not sure why I need to include the body?

ju-Skinner: You're fine. I didn't look at the complete code, my apologies.

libs/core/src/components/pds-table/pds-table-head-cell/test/pds-table-head-cell.e2e.ts
ju-Skinner: same comment as before. Not realizing it was just for the `table-head` only.

libs/core/src/components/pds-table/pds-table-head/test/pds-table-head.spec.tsx
ju-Skinner: same as other table-head comments

ju-Skinner (approved):
great work and effort here. 👏🏾  Lets get this merged in prior to the #123. Then we can update the table documentation in #123 to match the new format.

anechol (approved):
Really, really awesome and thorough work here! I'm excited for us finally getting this into the system and having a proper table. One minor thing to note, non-blocking:

Could we potentially have our subcomponent properties exist as different tabs within Storybook? This is something we currently do for SB in Sage.

Pine: 
Screenshot 2024-01-11 at 10 01 56 AM
Sage: 
Screenshot 2024-01-11 at 10 03 55 AM

pixelflips:
> Really, really awesome and thorough work here! I'm excited for us finally getting this into the system and having a proper table. One minor thing to note, non-blocking:
> 
> Could we potentially have our subcomponent properties exist as different tabs within Storybook? This is something we currently do for SB in Sage.
> 

@anechol I will look into this since I will be converting the documentation here to the new format in a separate follow-up.

PR #125, opened 2023-12-12
ci: create release workflow

ju-Skinner:
# Description

Creates the workflows necessary to deploy the packages to NPM.

* release: This is the main workflow that executes the child workflows and steps
* release-packages: This workflow has the ownership of releasing each package. Currently, that is `core` and `react`
* cd: This will be the Continuous Deployment workflow. We as a team will need to iron out how we want this to look.


pixelflips (approved):
LGTM!

QuintonJason (approved):

anechol (approved):

PR #127, opened 2023-12-15
chore: rename `react` package to `core-react`

ju-Skinner:
# Description

In the sake that there may be confusion, I elected to rename the `pine-ds/react` package to `pine-ds/core-react.` The name will then imply that this is the `core` React design system components.

Ticket: https://kajabi.atlassian.net/browse/DSS-497


pixelflips (approved):

anechol (approved):
Makes sense to me

PR #132, opened 2024-01-11
Discovery: Style Dictionary and Tiered Tokens

anechol:
## Description
This PR is to display discovery findings regarding updating the Pine tokens system as well as presenting updated naming conventions for token values. **Nothing has been added to Storybook so there is no need to run the project.**

### Why Style Dictionary?

In our current implementation of design tokens, we built a custom script that converts a JSON file built in Tokens Studio into CSS custom properties. While this currently functions fine, the grand question is: is this scalable? Adding in Style Dictionary aims to keep scalability in mind. In the event that we need to expand tokens to include native mobile devices or tokens in a CSS preprocessor, Style Dictionary can cover those use cases without us having to write manual code to convert our JSON to those formats.

## Tiered Tokens

Taking inspiration from other design systems, a tiered token system was implemented. This system is helpful for managing color or system themes such as dark mode or others we may need to add. All tier names are subject to change.

### Tier 1

Raw and base tokens will exist as tier 1. These are colors, spacing, fonts, box shadows, etc. in the most raw form. These tokens are meant to be consumed by Tier 2 tokens. Ideally, these tokens will be managed by Tokens Studio.

Example:

```
color.blue.300: #0072EF
color.charcoal.400: #202327
border-radius.10: 10px
```

Becomes:

```
--pine-color-blue-300: #0072EF;
--pine-color-charcoal-400: #202327;
--pine-border-radius-10: 10px;
```

### Tier 2

Tier 2 tokens, alias tokens, are semantic and reference Tier 1 tokens in their values. What makes these tokens useful is that they are applied to specific contexts and can used to service different needs such as system themes or dark mode. The advantage to using these tokens is that they can be updated once and propagate throughout the app. Ideally, these will be maintained and updated by the team.

Example:

```
color.text.link: {color.blue.300}
border-radius.default: {border-radius.10}
color.background.action.primary.default: {color.charcoal.400}
```

Becomes:

```
--pine-color-text-link: #0072EF
--pine-border-radius-default: 10px;
--pine-color-background-action-primary-default: #202327
```

### Tier 3

Tier 3 tokens, component level tokens, are tokens created specifically within components. They are meant to be used sparingly to call out specific contexts.

Example:

```
button.color.background.primary.default: {color.background.action.primary.default}
```

Becomes:

```
--pine-button-color-background-primary-default: #202327
```

## Final Notes

This is proposed as Step 1 in the Pine tokens refactor. There will be a list of follow-up tickets for the following:
- Add Tokens Studio integration
- Removing existing tokens
- Adding/using new tokens in components
- Updating internal documentation
- Adding tokens to Storybook

## Review

Some questions to keep in mind while reviewing that can help facilitate discussion:
- Does using Style Dictionary here make sense for our goals in Pine?
- Is the tiering system and purpose clear?
- Does the naming convention of tier 2 tokens clear?


ju-Skinner:
Great work Ashley.

One thing you didn't call out or maybe it's implied. Are we still using Token Studio to maintain the various tokens? If so, then Style Dictionary is only responsible for the transformation of the data. Is that correct?

libs/core/tokens/tokens.json
ju-Skinner, line 1: tl;dr
replace content of`build.tokens` in scripts with `npx style-dictionary build`. Remove `build-tokens.ts`

You'll need to update the `package.json` to remove the `build.tokens` script. Unless it needs to be updated to run `style-dictionary build` in replacement of. I assume you will be working that into this process, should it be the path forward. 

This should get you a green build

package.json
```diff
@@ -29,6 +29,7 @@
     "nx": "^16.8.1",
     "nx-cloud": "latest",
     "semver": "^7.3.8",
+    "style-dictionary": "^3.9.1",
```
ju-Skinner: This should not be in the root `package.json`. The way you want to add this package to `core` is by running `npm install style-dictionary -w @pine-ds/core`

Since we are in a mono repo we need to pass the `-w` option. 

`-w` is an option that tells npm we want to install this to the workspace.

anechol: Thanks! I'll make that update.

anechol:
> Great work Ashley.
> 
> One thing you didn't call out or maybe it's implied. Are we still using Token Studio to maintain the various tokens? If so, then Style Dictionary is only responsible for the transformation of the data. Is that correct?

@ju-Skinner Since we have mentioned wanting to shift source of truth from Figma to code, the idea was to make all token updates within the appropriate JSON and those can be propagated to Tokens Studio. There is some [existing interaction](https://github.com/tokens-studio/sd-transforms) between Style Dictionary and Tokens Studio, but I haven't gotten around to looking into that too deeply just yet.

ju-Skinner (approved):

QuintonJason (approved):
This is great work!

pixelflips (approved):
Great work! Thanks so much for all the research and communication along the way! LGTM! 👍🏼

PR #140, opened 2024-02-01
chore: remove console log functions from components

ju-Skinner:
# Description

Removes developer `console.log` calls from components


pixelflips (approved):
:shipit:

PR #158, opened 2024-03-25
[DSS-536] Copy Text + Image - Fix failing tests

anechol:
# Description

Addresses failing tests in the Copy Text and Image components. 

Copy Text:

The issue seemed to be two-fold: 
1. The Clipboard API was never used so the test defaulted to the fallback. This issue was discovered when the fallback was removed.
2. Because the Clipboard API is async, it's possible the promise never had time to resolve outside of a local testing environment. It's also possible this point may not be true because of point 1.

Image:

The issue was a flaky test. Using `img.complete` isn't a consistent way to test lazy loading as lazy loaded images can still return `complete: true` [when they're not visible in the viewport](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/complete). 

[DSS-536](https://kajabi.atlassian.net/browse/DSS-536)

## Type of change

# How Has This Been Tested?

As Clipboard API is supported in Safari as of version [13.1](https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/), the old fallback using `document.execCommand` has been removed. Clipboard functionality was tested in multiple browsers to ensure it works. Because Firefox requires permissions for Clipboard, the old versions were installed and tested directly instead of through Browserstack. Browserstack login info is in 1P.

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers: Firefox (121, 122); Safari (15.6, 16.5), Chrome (120, 121)
- Screen readers:
- Misc:

# Checklist:

[DSS-536]: https://kajabi.atlassian.net/browse/DSS-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/components/pds-copytext/pds-copytext.tsx
+++ b/libs/core/src/components/pds-copytext/pds-copytext.tsx
@@ -43,8 +43,10 @@ export class PdsCopytext {
 
   private copyToClipboard = async (value: string) => {
     try {
-      await navigator.clipboard.writeText(value);
-      this.pdsCopyTextClick.emit('Copied to clipboard');
+      if (typeof navigator.clipboard !== 'undefined') {
+        await navigator.clipboard.writeText(value);
+        this.pdsCopyTextClick.emit('Copied to clipboard');
+      }
     } catch (err) {
       console.error(err);
       this.pdsCopyTextClick.emit(`Error writing text to clipboard: ${err}`);
--- a/libs/core/src/components/pds-copytext/pds-copytext.tsx
+++ b/libs/core/src/components/pds-copytext/pds-copytext.tsx
@@ -48,7 +48,6 @@ export class PdsCopytext {
         this.pdsCopyTextClick.emit('Copied to clipboard');
       }
     } catch (err) {
-      console.error(err);
       this.pdsCopyTextClick.emit(`Error writing text to clipboard: ${err}`);
     }
   };
--- a/libs/core/src/components/pds-copytext/pds-copytext.tsx
+++ b/libs/core/src/components/pds-copytext/pds-copytext.tsx
@@ -43,24 +43,10 @@ export class PdsCopytext {
 
   private copyToClipboard = async (value: string) => {
     try {
-      if (navigator.clipboard) {
-        await navigator.clipboard.writeText(value);
-        this.pdsCopyTextClick.emit('Copied to clipboard');
-      } else {
-        // falback for safari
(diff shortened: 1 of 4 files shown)


libs/core/src/components/pds-copytext/pds-copytext.tsx
```diff
@@ -41,24 +41,12 @@ export class PdsCopytext {
...
-        el.value = value;
-        el.setAttribute('readonly', '');
-        el.style.position = 'absolute';
-        el.style.left = '-9999px';
-        document.body.appendChild(el);
-        el.focus();
-        el.setSelectionRange(0, el.value.length);
-        document.execCommand('copy');
-        document.body.removeChild(el);
-        this.pdsCopyTextClick.emit('Copied to clipboard');
```
ju-Skinner, line 61: Did you check to make sure the component still works in Safari? I didn't want to make any assumptions, hence the question.

anechol: Yes, all the browsers tested for this are mentioned in the config section.

ju-Skinner (approved):

pixelflips (approved):
Passing locally for me with no problems. Thanks for taking this on. I did notice some odd styling on the copy text button but that's for another day. Nice work! 👍🏼

QuintonJason (approved):
Great work!

libs/core/src/components/pds-copytext/pds-copytext.tsx
```diff
@@ -43,24 +43,12 @@ export class PdsCopytext {
...
       }
     } catch (err) {
+      console.error(err);
```
ju-Skinner: Should this error only be outputted if we are running locally? If so, when can wrap this in a `Build.isDev`

You will need to add Build to the import statement for `@stencil/core`

ju-Skinner (approved):
Comment, not blocking but I do believe we should not use console.error

QuintonJason (approved):
Looks and works great

libs/core/src/components/pds-copytext/pds-copytext.tsx
anechol: All good! I can just remove this.

PR #163, opened 2024-04-05
Tokens: Tokens Build Script

anechol:
## Description
Adds a script file for transforming and building tokens exported from Tokens Studio when the PR created from TS is merged in.

Because the tokens file that comes from Tokens Studio will be one JSON file, part of this script accounts for splitting the sets up into multiple JSON files and building separate SCSS files from those sets.

This script uses a [package](https://github.com/tokens-studio/sd-transforms/tree/v0.15.2?tab=readme-ov-file) called `sd-transforms`. This package allows for custom transforms to properly format tokens that are exported from the Tokens Studio plugin in Figma.

[DSS-513](https://kajabi.atlassian.net/browse/DSS-513)

## Testing
1. Grab the tokens set from [this PR](https://github.com/Kajabi/pine/pull/162/files). This would be the JSON file that would normally get merged into the repo via Tokens Studio.
2. Create a `tokens.json` file within the `libs/core/src/global/styles/tokens` folder.
3. In the project's root folder, run `npm install`
4. Then run `npx nx run core:build.tokens`
5. Two folders, `core` and `semantic`, should be created within the tokens folder
6. Check that each folder has a generated JSON file of tokens and a generated SCSS file
7. Check that each generated JSON file only contains tokens from the parent set (core and semantic)
8. Check that each generated SCSS file only contains CSS custom properties built from the JSON file in the same folder

[DSS-513]: https://kajabi.atlassian.net/browse/DSS-513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/scripts/sd-transforms.mjs
ju-Skinner, line 1: You don't want to commit this file. This file is generated from the command `npm run build.ts`.  I would look to add `scripts/**/*.mjs` to `libs/core/.gitignore`

To be honest, you really don't need to have the original file. You could have the `.mjs` file and above the ignore statement. I'll reach out via DM to discuss this further, in case there is confusion.

pixelflips (approved):
LGTM! I followed the provided testing instructions and all seems to work as expected! Really really nice work here!

One note: there are a couple of typos in the [tokens file](https://github.com/Kajabi/pine/blob/d03f49519578971285760c74488b96735074f388/libs/core/src/global/styles/tokens/tokens.json) that you linked out to in the testing instructions, but those don't relate to this PR.

QuintonJason (approved):
Tokens look good

PR #170, opened 2024-04-23
chore(tokens): updates font related tokens and adds typography semantic tokens

pixelflips:
# Description

To future-proof the token values, we will increase the scale of the current font sizes and line height values to provide more flexibility. Some values may not be used immediately but will be available if and when necessary.

These new values will also be used to set up the typography semantic tokens.

CORE:
Updates font weight
Updates line height
Updates font sizes
Fixes letter spacing

Semantic:
Add typography tokens based on core updates above

Fixes [DSS-586](https://kajabi.atlassian.net/browse/DSS-586)

## Type of change
- [X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [x] other: Token Studio in Figma

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-586]: https://kajabi.atlassian.net/browse/DSS-586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/global/styles/tokens/core/_core.scss
+++ b/libs/core/src/global/styles/tokens/core/_core.scss
@@ -1,13 +1,21 @@
 /**
  * Do not edit directly
- * Generated on Fri, 12 Apr 2024 15:37:39 GMT
+ * Generated on Wed, 24 Apr 2024 16:03:41 GMT
  */
 
 :root {
+  --pine-border-radius-0: 0;
   --pine-border-radius-100: 8px;
-  --pine-border-radius-160: 16px;
-  --pine-border-radius-040: 4px;
-  --pine-border-radius-060: 6px;
+  --pine-border-radius-125: 10px;
+  --pine-border-radius-150: 12px;
+  --pine-border-radius-175: 14px;
+  --pine-border-radius-200: 16px;
+  --pine-border-radius-225: 18px;
+  --pine-border-radius-250: 20px;
+  --pine-border-radius-275: 22px;
+  --pine-border-radius-300: 24px;
+  --pine-border-radius-050: 4px;
+  --pine-border-radius-075: 6px;
   --pine-border-radius-round: 9999px;
   --pine-border-width-none: 0;
   --pine-border-width-thin: 1px;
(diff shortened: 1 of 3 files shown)


libs/core/src/global/styles/tokens/tokens.json
```diff
@@ -751,6 +819,154 @@
...
       }
+    },
+    "heading-1": {
+      "value": {
+        "fontWeight": "{font-weight.bold}",
+        "fontSize": "{font-size.250}",
+        "lineHeight": "{line-height.275}",
+        "letterSpacing": "{letter-spacing.md}"
+      },
+      "type": "typography"
```
anechol: We should be able to add the font family in here as well in TS.

anechol (approved):
One small thing to change, but otherwise, awesome work getting these updated.

libs/core/src/global/styles/tokens/tokens.json
pixelflips: @anechol all should be set to `{font-family.circular}` now. Great catch!

QuintonJason (approved):
Good to go

ju-Skinner:
Shouldn't the `core.scss` and `semantic.scss` files be updated in the PR as well?

pixelflips:
> Shouldn't the `core.scss` and `semantic.scss` files be updated in the PR as well?

My understanding is that happens when the build tokens is run, which right now we are doing locally/manually. @anechol can confirm as I've only been updating from within token studio right now.

ju-Skinner:
Last Friday we talked about this. I mentioned how the build would potentially have conflicts due to files changing and it would put us in a weird scenario. I proposed that we should make those changes and have them be a part of the PR (this). Again we can reconvene as a team and discuss further. 

cc: @monicawheeler @QuintonJason @anechol

pixelflips:
Ok cool, got it. I can run here for now so we can get these in.

anechol (approved):
Nice work!

ju-Skinner (approved):

libs/core/src/global/styles/tokens/core/_core.scss
```diff
@@ -1,13 +1,21 @@
 /**
  * Do not edit directly
- * Generated on Fri, 12 Apr 2024 15:37:39 GMT
+ * Generated on Wed, 24 Apr 2024 16:03:41 GMT
  */
```
ju-Skinner, line 4: Note for @anechol. We need to figure out how to prevent this from being added.

anechol: Yes this is something I will look into.

PR #172, opened 2024-04-24
fix(accordion): animation for safari

ju-Skinner:
# Description

While investigating an issue reported, it was discovered that the animations were not working in Safari. It has been noted that transitions don't always work best in Safari. So I elected to use keyframes which seem to work well across all browsers. 

# Demo of before and after
https://www.loom.com/share/e32096c5bf4f4ae2b0f42f1748212eb5?sid=1244e615-466c-4a46-bcd2-a8b11bb88e80

# Notes
See the [slack message](https://kajabi.slack.com/archives/C03GKQJHCLX/p1714058690532659) about how to configure the local SSL Certs.

Fixes #(issue)
https://kajabi.atlassian.net/browse/DSS-582

## Type of change

# How Has This Been Tested?
Clicking the various panels to ensure they animate while opening.

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-accordion/pds-accordion.scss
+++ b/libs/core/src/components/pds-accordion/pds-accordion.scss
@@ -26,6 +26,16 @@ details {
   }
 }
 
+@keyframes slide-down {
+  from {
+    transform: translateY(-100%)
+  }
+
+  to {
+    transform: translateY(0%);
+  }
+}
+
 /* stylelint-disable-next-line */
 details[open] {
   background-color: var(--color-content-background);
@@ -89,6 +99,7 @@ summary {
 }
 
 .pds-accordion__body {
+  animation: slide-down var(--animation-transform-timing);
   max-height: 0;
   overflow: hidden;
 }
--- a/libs/core/.gitignore
+++ b/libs/core/.gitignore
@@ -10,6 +10,7 @@ scripts/**/*.js
 *.sw[mnpcod]
 *.log
 *.lock
+*.pem
 *.tmp
 *.tmp.*
 log.txt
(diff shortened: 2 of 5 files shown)


libs/core/src/components/pds-accordion/pds-accordion.scss
```diff
@@ -89,6 +97,7 @@ summary {
 }
 
 .pds-accordion__body {
+  animation: slide-down var(--animation-transform-timing);
```
pixelflips, line 102: Seeing 2 things and am not 100% sure why there's a difference.

1. Chrome doesn't seem to animate. Not sure if it's not animating or if 200ms is just super fast.

2. Safari now animates (and feels much slower than Chrome), but I see an overlap with the trigger and the sub-items.
Screenshot 2024-04-25 at 9 14 40 AM

If I move the `.pds-accordian__body` styles from line 99 into line 57 they both behave the same, but the overlap issue happens. This is in storybook, so not 100% sure if shows up in KP properly.

ju-Skinner: I added the original transitions back to the file. This seems to have fixed the Chrome issue you were referring to.

pixelflips (approved):

monicawheeler (approved):
Looks great! A small note, there are a couple of non-scss changes in this PR. Is that intentional or necessary? Would it be worth separating those into a separate PR? Thanks!

QuintonJason (approved):
Good to go

ju-Skinner:
> Looks great! A small note, there are a couple of non-scss changes in this PR. Is that intentional or necessary? Would it be worth separating those into a separate PR? Thanks!

@monicawheeler We wouldn't be able to test the validity of these changes without the `non-scss` changes. So, no we should not separate them out. Great question though.

PR #174, opened 2024-04-25
Tokens: update component tokens

anechol:
# Description

Updates the following components' tokens (Jira tickets linked):
- [Accordion](https://kajabi.atlassian.net/browse/DSS-560)
- [Avatar](https://kajabi.atlassian.net/browse/DSS-561)
- [Button](https://kajabi.atlassian.net/browse/DSS-562)
- [Copy Text](https://kajabi.atlassian.net/browse/DSS-565)
- [Image](https://kajabi.atlassian.net/browse/DSS-568)
- [Tabs](https://kajabi.atlassian.net/browse/DSS-576)
- [Label (internal)](https://kajabi.atlassian.net/browse/DSS-580)

## Type of change

- [ ] Style changes (updates tokens)

# How Has This Been Tested?

- [x] tested manually

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-accordion/pds-accordion.scss
+++ b/libs/core/src/components/pds-accordion/pds-accordion.scss
@@ -4,20 +4,25 @@
 
 details {
   --animation-transform-timing: 200ms;
-  --border-radius: var(--pine-border-radius-default);
+
+  --border-radius-default: var(--pine-border-radius-100);
+
   --box-shadow-focus: inset 0 0 0 2px var(--pine-color-blue-200);
-  --color-content-background: var(--pine-color-background-container-default);
+
+  --color-content-background-default: var(--pine-color-white);
   --color-background-hover: var(--pine-color-grey-300);
-  --color-font: var(--pine-color-charcoal-200);
-  --color-font-active: var(--pine-color-charcoal-500);
-  --color-font-hover: var(--pine-color-charcoal-300);
-  --font: var(--pine-body-semi);
+  --color-text-default: var(--pine-color-charcoal-200);
+  --color-text-active: var(--pine-color-charcoal-500);
+  --color-text-hover: var(--pine-color-charcoal-300);
+  
   --spacing-details-padding-inline: var(--pine-spacing-150);
   --spacing-summary-padding-block: var(--pine-spacing-050);
   --spacing-summary-padding-inline-start: var(--pine-spacing-100);
   --spacing-summary-padding-inline-end: var(--pine-spacing-050);
+  
+  --typography-default: var(--pine-font-weight-semi-bold) var(--pine-font-size-100)/var(--pine-line-height-150) var(--pine-font-family-circular);
 
-  border-radius: var(--border-radius);
+  border-radius: var(--border-radius-default);
   padding-inline: var(--spacing-details-padding-inline);
 
   pds-icon {
@@ -42,8 +47,8 @@ details[open] {
   transition: transform var(--animation-transform-timing);
 
   summary {
-    color: var(--color-font-active);
-    font: var(--font);
+    color: var(--color-text-active);
+    font: var(--typography-default);
 
     pds-icon {
       transform: scaleY(-1);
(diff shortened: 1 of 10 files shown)


libs/core/src/components/pds-accordion/pds-accordion.scss
```diff
@@ -4,20 +4,25 @@
 
 details {
   --animation-transform-timing: 200ms;
```
ju-Skinner, line 6: should we prefix this with `--number-animation-transform-timing`?

ju-Skinner (approved):
There is one comment left, but it can go either way. So non-blocking.

libs/core/src/components/pds-accordion/pds-accordion.scss
```diff
@@ -4,20 +4,25 @@
...
+  --color-content-background-default: var(--pine-color-white);
+  --color-background-hover: var(--pine-color-grey-300);
+  --color-text-default: var(--pine-color-charcoal-200);
+  --color-text-active: var(--pine-color-charcoal-500);
+  --color-text-hover: var(--pine-color-charcoal-300);
+  
+  --spacing-details-padding-inline: var(--pine-spacing-150);
+  --spacing-summary-padding-block: var(--pine-spacing-050);
+  --spacing-summary-padding-inline-start: var(--pine-spacing-100);
+  --spacing-summary-padding-inline-end: var(--pine-spacing-050);
```
pixelflips, line 21: Could these be adjusted to prevent using duplicate values?

libs/core/src/components/pds-button/pds-button.scss
```diff
@@ -1,140 +1,151 @@
...
+  --color-text-destructive-disabled: var(--pine-color-text-action-destructive-disabled);
+  --color-outline-destructive: var(--pine-color-red-200);
+
+  --spacing-disclosure-icon-margin-inline-end: var(--pine-spacing-0);
+  --spacing-disclosure-margin-inline-start: var(--pine-spacing-100);
+  --spacing-icon-margin-inline-end: var(--pine-spacing-100);
+  --spacing-padding-block: var(--pine-spacing-100);
+  --spacing-padding-inline: var(--pine-spacing-200);
+  --spacing-unstyled-padding: var(--pine-spacing-0);
+  --spacing-unstyled-margin: var(--pine-spacing-0);
```
pixelflips: I may be misunderstanding the spacing naming, but wouldn't we want to try and prevent duplicates?

libs/core/src/components/pds-tabs/pds-tab/pds-tab.scss
```diff
@@ -30,35 +38,35 @@ pds-tab {
 @mixin pds-button-style-reset {
   appearance: none;
   background-color: transparent;
-  border: 0;
+  border: var(--border-radius-none);
```
pixelflips: Should a radius token be used for the border? Not a show-stopper.

pixelflips (commented):
A couple of non-blocking comments/questions (mostly around spacing). Let me know on those and I will approve. I just wasn't 100%.

QuintonJason (approved):
Good to go

libs/core/src/components/pds-button/pds-button.scss
anechol: @pixelflips I understand your point. It feels like a case by case basis for me. I'm thinking in some cases repetition may be fine if it makes sense in the context it's being used. What do you think?

pixelflips: That makes sense, but those to me would be the outliers. For instance the `--spacing-unstyled-margin` and `--spacing-unstyled-padding` could be combined into something like `--spacing-unstyled-default` and then if they ever have different values we could separate them. That just seems more DRY to me and is one less variable to maintain over time as an example. 

Definitely not a blocker this time around.

pixelflips (approved):

libs/core/src/components/pds-button/pds-button.scss
anechol: I like this explanation. Thanks for this. Just updated the tokens.

libs/core/src/components/pds-tabs/pds-tab/pds-tab.scss
anechol: This is an oversight lol. Updated this as well. Good catch!

PR #175, opened 2024-04-25
feat(token): add token table

QuintonJason:
# Description
This PR is used to display Pine tokens. The proposed solution is to add those values to a table and use pages based on subcategories.

- [x] add token table

Fixes #(issue)
[DSS-517](https://kajabi.atlassian.net/browse/DSS-517)

Border
Screenshot 2024-05-02 at 2 19 22 PM

Box Shadow
Screenshot 2024-05-02 at 2 19 16 PM

Color
Screenshot 2024-05-02 at 2 19 10 PM
kens

Typography
Screenshot 2024-05-02 at 2 18 54 PM

Spacing
Screenshot 2024-05-02 at 2 19 02 PM

## Type of change

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-517]: https://kajabi.atlassian.net/browse/DSS-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/stories/typography.docs.mdx
+++ b/libs/core/src/stories/typography.docs.mdx
@@ -17,6 +17,3 @@ import {DocTokenTable} from '@pine-ds/doc-components';
 
 <h1>Core Line Height</h1>
 <DocTokenTable type="core" category="line-height" />
-
-<h1>Core Spacing</h1>
-<DocTokenTable type="core" category="spacing" />
--- a/libs/core/src/stories/typography.docs.mdx
+++ b/libs/core/src/stories/typography.docs.mdx
@@ -3,17 +3,22 @@ import {DocTokenTable} from '@pine-ds/doc-components';
 
 <Meta title="Foundations/Design Tokens/Typography"/>
 
-<h1>Core Font Family</h1>
-<DocTokenTable type="core" category="font-family" />
+# Font Family
 
-<h1>Core Font Size</h1>
-<DocTokenTable type="core" category="font-size" />
+<DocTokenTable category="font-family" />
 
-<h1>Core Font Weight</h1>
-<DocTokenTable type="core" category="font-weight" />
+# Font Size
 
-<h1>Core Letter Spacing</h1>
-<DocTokenTable type="core" category="letter-spacing" />
+<DocTokenTable category="font-size" />
 
-<h1>Core Line Height</h1>
-<DocTokenTable type="core" category="line-height" />
+# Font Weight
+
+<DocTokenTable category="font-weight" />
+
+# Letter Spacing
+
+<DocTokenTable category="letter-spacing" />
+
+
+# Line Height
+<DocTokenTable category="line-height" />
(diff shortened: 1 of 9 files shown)


libs/core/src/stories/typography.docs.mdx
```diff
@@ -1,8 +1,24 @@
...
+
+<DocTokenTable category="font-family" />
+
+# Font Size
+
+<DocTokenTable category="font-size" />
+
+# Font Weight
+
+<DocTokenTable category="font-weight" />
```
anechol, line 16: For this and font family, the font family in storybook doesn't reflect what is expected of the design system because we never set it up that way. I feel like we should change this, but that feels like a team decision. What do you think?

anechol (approved):
I wish I had the words to describe how awesome it is seeing this work come together 🔥 🔥  

Three very small, non-blocking things: 
1. Since we will be using semantic tokens eventually, can we put these into a folder called "Core" under Design Tokens?
2. Could we update the border colors to match the boxes in the spacing tokens for visibility?
3. If we can't find a way to reorder the tokens that are out of place for having a "0" in front of them, I wouldn't be opposed to dropping the "0" if it makes them easier to display.

pixelflips (approved):
This is 🔥! I'd love to see the table incorporate the styles from the props table, but that could be handled in a follow-up. Great work!

ju-Skinner:
Can you rebase against `main` to see if it resolves the `copy-text` test failures?

QuintonJason:
@anechol 
> 1. Since we will be using semantic tokens eventually, can we put these into a folder called "Core" under Design Tokens?

Is it ok if we make that change when the semantic tokens are added? Until semantic tokens exist there will be that extra nesting layer that may confuse users 

> 2. Could we update the border colors to match the boxes in the spacing tokens for visibility?

Can do

> 3. If we can't find a way to reorder the tokens that are out of place for having a "0" in front of them, I wouldn't be opposed to dropping the "0" if it makes them easier to display.

Currently the tables are reflecting the order of the keys that exist in the core.json and core.scss files. This may be a situation where we have to make that change in the source, which may be Token Studio

anechol:
> @anechol
> 
> > 1. Since we will be using semantic tokens eventually, can we put these into a folder called "Core" under Design Tokens?
> 
> Is it ok if we make that change when the semantic tokens are added? Until semantic tokens exist there will be that extra nesting layer that may confuse users

Yep, that's fine with me.

> > 2. Could we update the border colors to match the boxes in the spacing tokens for visibility?
> 
> Can do
> 
> > 3. If we can't find a way to reorder the tokens that are out of place for having a "0" in front of them, I wouldn't be opposed to dropping the "0" if it makes them easier to display.
> 
> Currently the tables are reflecting the order of the keys that exist in the core.json and core.scss files. This may be a situation where we have to make that change in the source, which may be Token Studio

Yep, this is what I was thinking. That should probably be a discussion because we would have to update components using those affected tokens.

ju-Skinner (approved):

PR #177, opened 2024-04-29
style(chip): update token variables

pixelflips:
# Description

Update chip tokens to use new values where applicable.

Fixes [#(DSS-564)](https://kajabi.atlassian.net/browse/DSS-564)

## Type of change
- [x] Token adjustments

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-chip/pds-chip.scss
+++ b/libs/core/src/components/pds-chip/pds-chip.scss
@@ -34,9 +34,9 @@
   --spacing-xl: var(--pine-spacing-150);
   --spacing-xxl: calc(var(--pine-spacing-350) / 2);
 
-  --dot-size: var(--spacing-xs);
-  --close-size: var(--spacing-lg);
-  --close-size-lg: calc(var(--spacing-xl) * 2);
+  --close-size: 10px;
+  --close-size-lg: 24px;
+  --dot-size: 4px;
 
   align-items: center;
   border-radius: var(--border-radius);
@@ -114,7 +114,7 @@ $pds-chip-sentiment-interactive: (
   padding: 0;
 
   .pds-chip__dot {
-    margin-block-end: 1px;
+    margin-block-end: calc(var(--spacing-xxs) / 4);
     margin-block-start: var(--spacing-xxs);
   }
 }
--- a/libs/core/src/components/pds-chip/pds-chip.scss
+++ b/libs/core/src/components/pds-chip/pds-chip.scss
@@ -34,9 +34,9 @@
   --spacing-xl: var(--pine-spacing-150);
   --spacing-xxl: calc(var(--pine-spacing-350) / 2);
 
-  --close-size: 10px;
-  --close-size-lg: 24px;
-  --dot-size: 4px;
+  --sizing-close: 10px;
+  --sizing-close-lg: 24px;
+  --sizing-dot: 4px;
 
   align-items: center;
   border-radius: var(--border-radius);
@@ -102,10 +102,10 @@ $pds-chip-sentiment-interactive: (
   border: 1px solid transparent;
   border-radius: 50%;
   display: inline-block;
-  height: var(--dot-size);
+  height: var(--sizing-dot);
   margin-block-end: var(--spacing-xxs);
   margin-inline-end: var(--spacing-xs);
-  width: var(--dot-size);
+  width: var(--sizing-dot);
 }
 
 // dropdown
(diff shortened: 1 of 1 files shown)


libs/core/src/components/pds-chip/pds-chip.scss
```diff
@@ -1,82 +1,78 @@
...
+  --spacing-xs: var(--pine-spacing-050);
+  --spacing-sm: calc(var(--pine-spacing-150) / 2);
+  --spacing-md: var(--pine-spacing-100);
+  --spacing-lg: calc(var(--pine-spacing-250) / 2);
+  --spacing-xl: var(--pine-spacing-150);
+  --spacing-xxl: calc(var(--pine-spacing-350) / 2);
 
-  --dot-size: 4px;
   --close-size: 10px;
   --close-size-lg: 24px;
```
ju-Skinner: These "sizing" vars should be prefixed with `--sizing`.

ju-Skinner (changes requested):
Small fix.

libs/core/src/components/pds-chip/pds-chip.scss
```diff
@@ -1,82 +1,78 @@
...
+  --color-info-interactive: var(--pine-color-blue-200);
+  --color-neutral: var(--pine-color-grey-300);
+  --color-neutral-dot: var(--pine-color-charcoal-100);
+  --color-neutral-interactive: var(--pine-color-grey-400);
+  --color-success: var(--pine-color-green-100);
+  --color-success-dot: var(--pine-color-green-300);
+  --color-success-interactive: var(--pine-color-green-200);
+  --color-warning: var(--pine-color-yellow-100);
+  --color-warning-dot: var(--pine-color-yellow-300);
+  --color-warning-interactive: var(--pine-color-yellow-200);
```
anechol: Could we add the context of these colors here, e.g. if they're background or text?

libs/core/src/components/pds-chip/pds-chip.scss
```diff
@@ -1,82 +1,78 @@
...
-  --margin-inline-xl: calc(var(--pine-spacing-md) / 2);
+  --line-height: var(--pine-line-height-100);
+
+  --spacing-xxs: calc(var(--pine-spacing-050) / 2);
+  --spacing-xs: var(--pine-spacing-050);
+  --spacing-sm: calc(var(--pine-spacing-150) / 2);
+  --spacing-md: var(--pine-spacing-100);
+  --spacing-lg: calc(var(--pine-spacing-250) / 2);
+  --spacing-xl: var(--pine-spacing-150);
+  --spacing-xxl: calc(var(--pine-spacing-350) / 2);
```
anechol, line 35: Could we add context to these tokens as well? How they're being applied? Margin, padding, etc?

anechol (changes requested):
Nice work getting these updated. Just a couple things I think could use a second look.

libs/core/src/components/pds-chip/pds-chip.scss
pixelflips: these are used for both padding/margins. what else could i use?

ju-Skinner (approved):

libs/core/src/components/pds-chip/pds-chip.scss
anechol: I was thinking making a distinction in usage between block and inline if it applies.

pixelflips: @anechol This is the area where I am most confused I guess. If I have to break them out into each possible spacing type inline, block, start, end, etc... we end up with a lot of repeating values and one-off variables. To me that would defeat the purpose of using the variables and would be better to just add them separately to each selector instead.

I may be misunderstanding but let me know.

anechol: I commented on another PR about this, but I do see your point and it makes total sense. For some cases it _could_ be fine, but the reasoning you highlighted works for me. I think this process gets a lot clearer making a component and variables from scratch instead of updating existing ones. Thanks for the insight and perspective here. Let's get this merged in.

anechol (approved):

PR #220, opened 2024-07-01
Tokens: Updates Config File

anechol:
# Description

This PR initially served as a testing ground for discovery regarding the themes and multi-file sync features within Tokens Studio Pro. The final update for this PR is updating the Style Dictionary config file to use multi-file sync. One step of the config file, separating tokens into their own separate files, has been removed in favor of programmatically reading token folders so that token files can be built without the config file needing to be updated to account for new token sets.

The [Confluence page](https://kajabi.atlassian.net/wiki/spaces/DSS/pages/1743028324/Using+Tokens+Studio) for Tokens Studio has been updated with a new image to reflect the file path in the settings.

## Type of change

- [x] Refactor

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:


libs/core/scripts/sd-transforms.mjs
```diff
@@ -6,6 +6,17 @@ registerTransforms(StyleDictionary);
 
 const basePath = `src/global/styles/tokens`;
 
+const getTokenSetNamesFromFolders = (path) => {
+	let tokenSets = [];
+	fs.readdirSync(path, { withFileTypes: true })
+		.filter(dirent => dirent.isDirectory())
+		.map(dirent => tokenSets.push(dirent.name));
+
+	console.log('tokenSets: ', tokenSets);
```
pixelflips: Is the console.log needed?

pixelflips (approved):
one small comment, otherwise lgtm. 👍🏼

QuintonJason (approved):

libs/core/src/global/styles/tokens/$themes.json
```diff
@@ -0,0 +1 @@
+[]
```
ju-Skinner, line 1: Would you like us to introduce a theme here? I believe this will also give us the greatest flexibility with the Rebrand work.

ju-Skinner (approved):

libs/core/src/global/styles/tokens/$themes.json
anechol: I don't think that'll be necessary since we'll be getting a mapping of old -> new tokens. With the way Tokens Studio works with theming, it allows you to toggle on certain sets of tokens to achieve that particular theme. I feel like theming will come more into play once we establish a set of semantic tokens.

PR #223, opened 2024-07-11
Tokens: Adds Letter Spacing

anechol:
# Description

Adds new letter spacing tokens and removes leading zero from existing font size and letter spacing tokens.

[DSS-682](https://kajabi.atlassian.net/browse/DSS-682)

## Type of change

- [x] Token change (adds, updates, or removes design tokens)

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

[DSS-682]: https://kajabi.atlassian.net/browse/DSS-682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/src/global/styles/tokens/core/_core.scss
```diff
@@ -106,6 +106,10 @@
   --pine-color-purple-900: #372d82;
   --pine-color-purple-950: #221b4b;
   --pine-font-family-greet: "Greet";
+  --pine-font-size-50: 8px;
+  --pine-font-size-62: 10px;
+  --pine-font-size-75: 12px;
+  --pine-font-size-87: 14px;
   --pine-font-size-100: 16px;
```
pixelflips: I think we'll have to adjust the scale if the base font size (100) is going to be changing to 14px.

pixelflips (changes requested):
Left a comment on the scale changing. If that does not apply here, just let me know.

libs/core/src/global/styles/tokens/core/_core.scss
anechol: No you're correct, I missed that. Working too quickly here 😅

Will get these updated shortly.

libs/core/src/global/styles/tokens/semantic/semantic.json
```diff
@@ -372,7 +372,7 @@
   "body-xs": {
     "value": {
       "fontWeight": "{font-weight.normal}",
-      "fontSize": "{font-size.075}",
+      "fontSize": "{font-size.75}",
       "lineHeight": "{line-height.100}",
       "fontFamily": "{font-family.greet}"
```
pixelflips: Also may not be related to this PR, but I think greet is only applied to headings.

anechol: I'm actually just going to remove these semantic tokens since they aren't in use at the moment and I will be redoing them soon.

pixelflips (approved):
LGTM! Nice work!

ju-Skinner (approved):

PR #237, opened 2024-08-09
style: add expanded color palette tokens

QuintonJason:
# Description
This PR ensure that all new colors are added to the color token set:
- `sky`
- `sage`
- `teal`
- `lime`
- `magenta`
- `pink`
- `rose`

- [x] add expanded color palette

## Type of change


pixelflips (approved):

anechol (changes requested):
Thanks for adding these. Let's change the `50` values to `050` as well here once we get a rebase in

PR #241, opened 2024-08-14
fix: update box shadow tokens

QuintonJason:
# Description
- [x] update and apply box shadow tokens to match the figma design tokens
- [x] update sort function for token table 

## Type of change

# How Has This Been Tested?
1. In Token Studio
2. [Box shadow tokens page](http://localhost:6006/?path=/docs/foundations-design-tokens-box-shadow--docs)


pixelflips (commented):
There are a few usages in various locations, should we update those as well?

Screenshot 2024-08-14 at 2 34 45 PM

libs/core/src/components/pds-box/pds-box.tsx
```diff
@@ -163,7 +163,7 @@ export class PdsBox {
     ${this.border ? 'pds-box--border' : ''}
     ${this.borderRadius !== undefined && this.borderRadius.trim() !== '' ? `pds-border-radius-${this.borderRadius}` : ''}
     ${this.direction !== undefined && this.direction.trim() !== '' ? `pds-box-direction-${this.direction}` : ''}
-    ${this.display !== undefined && this.display.trim() !== '' ? `pds-box--display-${this.display}` : ''}
+    ${this.display !== undefined && this.display.trim() !== '' ? `pds---display-${this.display}` : ''}
```
anechol: We're missing "box" here.

libs/core/src/global/styles/tokens/core/_core.scss
```diff
@@ -15,10 +15,13 @@
...
-  --pine-box-shadow-sm: 0 2px 4px 0 rgba(0,0,0,0.12), 0 0 2px 0 rgba(0,0,0,0.08);
-  --pine-box-shadow-md: 0 8px 14px 0 rgba(0,0,0,0.16), 0 0 4px 0 rgba(0,0,0,0.08);
-  --pine-box-shadow-lg: 0 8px 40px 0 rgba(0,0,0,0.24);
+  --pine-box-shadow-100: 0 1px 3px 0 rgba(0,0,0,0.1), 0 1px 2px 0 rgba(0,0,0,0.06);
+  --pine-box-shadow-150: 0 4px 8px -2px rgba(0,0,0,0.1), 0 2px 4px -2px rgba(0,0,0,0.06);
+  --pine-box-shadow-200: 0 12px 16px -4px rgba(0,0,0,0.08), 0 4px 6px -2px rgba(0,0,0,0.03);
+  --pine-box-shadow-300: 0 20px 24px -4px rgba(0,0,0,0.08), 0 8px 8px -4px rgba(0,0,0,0.03);
+  --pine-box-shadow-400: 0 24px 48px -12px rgba(0,0,0,0.18);
+  --pine-box-shadow-500: 0 32px 64px -12px rgba(0,0,0,0.14);
+  --pine-box-shadow-050: 0 1px 2px 0 rgba(0,0,0,0.05);
```
anechol, line 24: 🔥 🔥

libs/core/src/components/pds-box/pds-box.tsx
pixelflips: are the triple dashes here intentional?

pixelflips (commented):
Left a comment for clarification. Let me know and I'll approve.

libs/core/src/components/pds-box/pds-box.tsx
QuintonJason: Great catch. Updated

anechol (changes requested):
Thanks for getting these updated. Only one minor issue below that needs addressing.

pixelflips (approved):

libs/core/src/components/pds-box/pds-box.tsx
anechol: Ah looks like that was addressed right when I commented. This can be ignored. LGTM!

QuintonJason: Great catch! Updated

anechol (approved):

PR #232, opened 2024-08-02
ci: change to nx release

ju-Skinner:
# Description
Change over from bare metal NPM publish to NX Release.

PR #255, opened 2024-09-09
ci: change to nx release

ju-Skinner:
# Description

This is a re-do for PR #232


.github/workflows/schedule-release.yml
```diff
@@ -0,0 +1,216 @@
+name: "Pine Continuous Deployment"
+
+on:
+  ## We will uncomment this when we settle on a deployment schedule
```
monicawheeler, line 4: Do we want to set this schedule prior to merging?

.prettierignore
```diff
@@ -0,0 +1,5 @@
+# Add files here to ignore them from prettier formatting
+/dist
+/coverage
+/.nx/cache
+/.nx/workspace-data
```
monicawheeler: Minor note, eol missing

monicawheeler (approved):
Minor comments, otherwise looks good!

.github/workflows/schedule-release.yml
ju-Skinner: If that's what we want. This could also be run manually like it is currently today.

QuintonJason (approved):

pixelflips (approved):

anechol (approved):

PR #257, opened 2024-09-12
style: converts tokens to W3C DTCG format

anechol:
# Description

Tokens Studio and Style Dictionary have released new major versions that now use the W3C Design Tokens Community Group's (DTCG). This PR updates Pine tokens to follow that format instead of remaining on the legacy format.

There should be no significant changes in tokens, values, or UI.

[DSS-891](https://kajabi.atlassian.net/browse/DSS-891)

## Type of change

- [x] Tokens (adds, updates, or removes design tokens)

[DSS-891]: https://kajabi.atlassian.net/browse/DSS-891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/global/styles/tokens/core/_core.scss
+++ b/libs/core/src/global/styles/tokens/core/_core.scss
@@ -12,7 +12,7 @@
   --pine-border-radius-050: 4px;
   --pine-border-radius-075: 6px;
   --pine-border-radius-round: 9999px;
-  --pine-border-width-none: 0;
+  --pine-border-width-none: 0rem;
   --pine-border-width-thin: 1px;
   --pine-border-width-thick: 2px;
   --pine-box-shadow-100: 0 1px 3px 0 rgba(0,0,0,0.1), 0 1px 2px 0 rgba(0,0,0,0.06);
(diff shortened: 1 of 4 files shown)


libs/core/src/global/styles/tokens/core/_core.scss
```diff
@@ -12,7 +12,7 @@
   --pine-border-radius-050: 4px;
   --pine-border-radius-075: 6px;
   --pine-border-radius-round: 9999px;
-  --pine-border-width-none: 0;
+  --pine-border-width-none: 0rem;
```
pixelflips: Curious to understand why 0 needs a unit?

pixelflips (approved):

libs/core/src/global/styles/tokens/core/_core.scss
anechol: @pixelflips This is part of the built-in CSS transform group in Style Dictionary that will add units to some 0 values. I just pushed a change that will make sure these values are pixels instead.

pixelflips: @anechol gotcha. I was just curious since 0 is typically unitless, but thanks for clarifying it's needed for Style Dictionary. Seems like an odd thing on their part but 🤷🏼.

QuintonJason (approved):

anechol:
Closing as this has already been implemented in other recent Pine PRs.

PR #306, opened 2024-11-13
style(tabs): updates component with semantic tokens

anechol:
# Description

Updates component with new semantic tokens.

## Type of change

- [x] Styles (adds, updates, or removes styling)

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

### DIFF
--- a/libs/core/src/components/pds-tabs/pds-tab/pds-tab.scss
+++ b/libs/core/src/components/pds-tabs/pds-tab/pds-tab.scss
@@ -146,7 +146,7 @@ pds-tab {
 }
 
 .pds-tabs--filter {
-  --color-background-tab: var(--pine-color-secondary);
+  --color-background-tab: var(--pine-color-background-container);
 
   .pds-tab {
     background-color: var(--color-background-tab);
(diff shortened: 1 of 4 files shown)


libs/core/src/components/pds-tabs/pds-tab/pds-tab.scss
```diff
@@ -100,12 +82,12 @@ pds-tab {
 
   .pds-tab {
     background-color: var(--color-background-tab);
-    border-radius: var(--border-radius-availability) var(--border-radius-availability) var(--border-radius-none) var(--border-radius-none);
+    border-radius: calc(var(--pine-dimension-xs) * 1.25) calc(var(--pine-dimension-xs) * 1.25) var(--pine-dimension-none) var(--pine-dimension-none);
```
ju-Skinner, line 85: I know we discussed this, but let's revisit once we are done. For the calculations, I think we should hard code the expected value, just because if values change it could have a larger impact.

ju-Skinner (approved):

PR #310, opened 2024-11-14
feat(select): adds new component

FuturaExtraBold:
# Description

Adds new `<pds-select />` web component and React component to the codebase. 

No new dependencies are needed.

### Jira Ticket

https://kajabi.atlassian.net/browse/DSS-1147

image

## Type of change

# How Has This Been Tested?

- [x] unit tests (Not full coverage ATM)
- [x] e2e tests
- [x] accessibility tests (Not full coverage ATM)
- [x] tested manually

**Test Configuration**:

- OS: macOS 14.6.1 (23G93)
- Browsers: Chrome 124, Safari 17.6
- Screen readers: N/A

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works (Not full coverage ATM)
- [x] New and existing tests pass locally with my changes (Not full coverage ATM)


libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+</p>
+
+## Guidelines
+
+### Usage
+
+#### When to use
+
+- When you want to provide a simple list of options for the user to choose from.
+- Use a select component when there are fewer options (typically 5–10) and only a single selection is required.
```
ju-Skinner: >Use a select component when there are fewer options (typically 5–10) and only a single selection is required.

This is an inaccurate statement. This was called out in the document you retrieved it from. Nor is this a guideline we want to have for this component.

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+- **Example**: Use a select component to choose a state from a short list of U.S. regions.
+
+#### When **NOT** to use
+
+- If your list of options is greater than 10, consider using a custom dropdown component. (WIP)
+- If you need to allow multiple selections, search, icons, nested items, or other complex choices, use a custom dropdown component. (WIP)
```
ju-Skinner: This is also inaccurate

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+- The select component comes with a built-in label through the `label` property, it is strongly encouraged that you include it.
+- Please provide clear and descriptive error, and helper-message text.
+- Selects can have a default option that is not selectable.
+
+## Properties
+
+<DocArgsTable componentName="pds-select" docSource={components} />
+
+## Variants
```
ju-Skinner: We do not need this title here, there is only one.

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+### Default
+
+<p>This example shows the default state of the select component with a label and a list of options. The user can choose one of the options from the dropdown menu.</p>
+
+<DocCanvas
+  client:only
+  mdxSource={{
+    react: '<PdsSelect componentId="pds-select-default-example" label="Name"></PdsSelect>',
+    webComponent: '<pds-select component-id="pds-select-default-example" label="Name"></pds-select>',
```
ju-Skinner: Your examples do not match the code you have below. They should include the `options`.

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+### Disabled
+
+<p>This example demonstrates the select component in a disabled state. The user cannot interact with the dropdown menu or select any options.</p>
+
+<DocCanvas
+  client:only
+  mdxSource={{
+    react: '<PdsSelect componentId="pds-select-disabled-example" disabled="true" label="Name"></PdsSelect>',
+    webComponent: '<pds-select component-id="pds-select-disabled-example" disabled="true" label="Name"></pds-select>',
```
ju-Skinner: Same comment as before, regarding `options` missing from the examples

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+### With Message
+
+<p>This example includes a helper message that provides additional context or instructions for the user. The helper message appears below the select component.</p>
+
+<DocCanvas
+  client:only
+  mdxSource={{
+    react: '<PdsSelect componentId="pds-select-with-message-example" label="Name" helper-message="This is a helper message to convey extra context for the select"></PdsSelect>',
+    webComponent:
+      '<pds-select component-id="pds-select-with-message-example" label="Name" helper-message="This is a helper message to convey extra context for the select"></pds-select>',
```
ju-Skinner: Same comment as before, regarding `options` missing from the examples

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+### Invalid
+
+<p>This example shows the select component with an error message. The error message provides feedback to the user when there is an issue with their selection.</p>
+
+<DocCanvas
+  client:only
+  mdxSource={{
+    react: '<PdsSelect componentId="pds-select-invalid-example" label="Name" error-message="This is an error message to convey extra context for the select"></PdsSelect>',
+    webComponent:
+      '<pds-select component-id="pds-select-invalid-example" label="Name" error-message="This is an error message to convey extra context for the select"></pds-select>',
```
ju-Skinner: Same comment as before, regarding `options` missing from the examples

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+### Autocomplete
+
+<p>This example demonstrates the select component with the autocomplete attribute enabled. The browser may provide autocomplete suggestions based on the user's input.</p>
+
+<DocCanvas
+  client:only
+  mdxSource={{
+    react: '<PdsSelect autocomplete="on" componentId="pds-select-autocomplete-example" label="Name"></PdsSelect>',
+    webComponent: '<pds-select autocomplete="on" component-id="pds-select-autocomplete-example" label="Name"></pds-select>',
```
ju-Skinner: Same comment as before, regarding `options` missing from the examples

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,149 @@
...
+  <div style={{ padding: '20px', width: '100%' }}>
+    <pds-select autocomplete="on" component-id="pds-select-autocomplete-example" label="Name">
+      <option value="paul">Paul McCartney</option>
+      <option value="john">John Lennon</option>
+      <option value="george">George Harrison</option>
+      <option value="ringo">Ringo Starr</option>
+    </pds-select>
+  </div>
+</DocCanvas>
+
```
ju-Skinner, line 191: There are a few missing examples that I don't see.

* `value` set, with the correct option selected
* `optgroup`  - grouping of options
* `multiple` - allow multiple values to be selected

libs/core/src/components/pds-select/pds-select.scss
```diff
@@ -0,0 +1,127 @@
...
+select {
+  appearance: none;
+  border: 1px solid var(--color-border-default);
+  border-radius: 10px;
+  font-size: var(--font-size-select-field);
+  font-weight: var(--font-weight-select-field);
+  grid-area: field;
+  line-height: var(--line-height-select-field);
+  padding: var(--spacing-field-padding-block) var(--spacing-select-field-padding-inline);
+  padding-right: var(--pine-spacing-450);
```
ju-Skinner: What is your reasoning for not making this a `CSS Custom Prop`? I see other instances where it's only used in one place, but it was defined above.

libs/core/src/components/pds-select/pds-select.scss
```diff
@@ -0,0 +1,127 @@
...
+  line-height: var(--line-height-helper-message);
+  margin-block-end: 0;
+  margin-block-start: var(--spacing-margin-top-helper-message);
+}
+
+.pds-select__error-message {
+  align-items: center;
+  color: var(--color-text-error);
+  display: flex;
+  gap: var(--pine-spacing-050);
```
ju-Skinner: Same response as the other `pine-spacing` token used above

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+   * Indicates whether or not the select field is required.
+   */
+  @Prop() required: boolean;
+
+  /**
+   * The value of the select field.
+   */
+  @Prop({ mutable: true }) value?: string;
```
ju-Skinner: We should have logic that adds `selected` to the corresponding `option` if the developer sets the `value`. In the example below, I would expect to see that the option george would have `selected` 

e.g 
```
<pds-select component-id="sample" label="sample names" value="george">
    <option value="paul">Paul McCartney</option>
    <option value="john">John Lennon</option>
    <option value="george">George Harrison</option>
    <option value="ringo">Ringo Starr</option>
</pds-select>
```

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,146 @@
...
+      if (option.tagName === 'OPTION') {
+        this.selectEl.appendChild(option.cloneNode(true));
+      }
+    });
+  };
+
+  render() {
+    return (
+      <Host aria-disabled={this.disabled ? 'true' : null}>
+        <div class="pds-select">
```
ju-Skinner, line 202: do we need this `div`? The `Host` element acts as a div when rendered.

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,146 @@
...
+              )}
+              {this.errorMessage && (
+                <p class="pds-select__error-message" id={messageId(this.componentId, 'error')} aria-live="assertive">
+                  <pds-icon icon={danger} size="small" />
+                  {this.errorMessage}
+                </p>
+              )}
+            </div>
+          )}
+          <pds-icon class="pds-select__select-icon" icon={enlarge} />
```
ju-Skinner: Is there a reason you elected to import the icons and not use `name="enlarge"`?

libs/core/src/components/pds-select/stories/pds-select.stories.js
```diff
@@ -0,0 +1,76 @@
...
+  html`<pds-select
+    autocomplete="${args.autocomplete}"
+    component-id="${args.componentId}"
+    disabled="${args.disabled}"
+    error-message="${args.errorMessage}"
+    helper-message="${args.helperMessage}"
+    label="${args.label}"
+    name="${args.name}"
+    required="${args.required}"
+    type="${args.type}"
```
ju-Skinner, line 78: Where is `value`?

libs/core/src/components/pds-select/stories/pds-select.stories.js
```diff
@@ -0,0 +1,76 @@
...
+    <option value="ringo">Ringo Starr</option>
+  </pds-select>`;
+
+export const Default = BaseTemplate.bind({});
+Default.args = {
+  componentId: 'pds-select-default-example',
+  disabled: false,
+  label: 'Select your favorite Beatle',
+  name: 'beatles',
+  required: true,
```
ju-Skinner, line 92: `value` is missing

libs/core/src/components/pds-select/test/pds-select.e2e.ts
```diff
@@ -0,0 +1,42 @@
...
+    expect(component).toHaveClass('hydrated');
+
+    const element = await page.find('pds-select >>> select');
+    let value = await element.getProperty('disabled');
+    expect(value).toBe(false);
+
+    component.setProperty('disabled', true);
+    await page.waitForChanges();
+    value = await element.getProperty('disabled');
+    expect(value).toBe(true);
```
ju-Skinner: There should also be some other checks on the element. e.g classes, attributes etc.

libs/core/src/components/pds-select/test/pds-select.e2e.ts
```diff
@@ -0,0 +1,42 @@
...
+    const page = await newE2EPage();
+
+    await page.setContent('<pds-select><option value="0">Option 1</option><option value="1">Option 2</option></pds-select>');
+    const select = await page.find('pds-select >>> select');
+
+    const options = await select.findAll('option');
+    expect(options.length).toBe(2);
+    expect(options[0].textContent).toBe('Option 1');
+    expect(options[1].textContent).toBe('Option 2');
+  });
```
ju-Skinner, line 48: You are missing a test for `value` that would show the behavior of selecting an option on render. 

These E2E tests should be written as if an end-user is using them in the browser. We should `click` functions and check the expectation of the interaction. I do not see any of that here.

libs/core/src/components/pds-select/test/pds-select.spec.tsx
```diff
@@ -0,0 +1,229 @@
...
+    // Trigger the handleSlotChange logic
+    const slot = page.root?.shadowRoot?.querySelector('slot');
+    const select = page.root?.shadowRoot?.querySelector('select');
+    const slottedOptions = slot?.assignedNodes?.({ flatten: true });
+
+    slottedOptions?.forEach((node) => {
+      if (node.nodeType === Node.ELEMENT_NODE) {
+        select?.appendChild(node.cloneNode(true));
+      }
+    });
```
ju-Skinner, line 45: Do you need this here? Is there another way that the internal function `handleSlotChange` could be called?

libs/core/src/components/pds-select/test/pds-select.spec.tsx
```diff
@@ -0,0 +1,229 @@
...
+    if (!pdsSelect || !pdsSelect.shadowRoot) {
+      throw new Error('pdsSelect or pdsSelect.shadowRoot is not available');
+    }
+    const select = pdsSelect.shadowRoot.querySelector('select');
+
+    // Change value to 'paul'
+    if (select) {
+      select.value = 'paul';
+      select.dispatchEvent(new Event('change'));
+      await page.waitForChanges();
```
ju-Skinner: You should also be checking to see that the `pdsSelectChange` event has been emitted.

I noticed we don't have that event, this is something that should be added. As a developer, they would like the ability to hook into that event, to perform or add additional side-effects when the value changes.

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,146 @@
...
+   * The value of the select field.
+   */
+  @Prop({ mutable: true }) value?: string;
+
+  /**
+   * Emitted when a keyboard input occurred.
+   */
+  @Event() pdsSelect: EventEmitter<InputEvent>;
+
+  /**
```
ju-Skinner, line 93: We need an event `pdsSelectChange`

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,146 @@
...
+   * Indicates whether or not the select field is required.
+   */
+  @Prop() required: boolean;
+
+  /**
+   * The value of the select field.
+   */
+  @Prop({ mutable: true }) value?: string;
+
```
ju-Skinner, line 73: Missing a prop that would allow `multiple` selections

libs/core/src/components/pds-select/test/pds-select.spec.tsx
```diff
@@ -0,0 +1,229 @@
...
+            <div aria-hidden="true" class="hidden">
+              <slot></slot>
+            </div>
+            <pds-icon class="pds-select__select-icon" icon="${enlarge}"></pds-icon>
+          </div>
+        </mock:shadow-root>
+      </pds-select>
+    `);
+  });
+});
```
ju-Skinner, line 584: Overall good job on the `.spec` test. These are meant to only test a piece of functionality (functions) and you did that.

libs/core/src/components/pds-select/test/pds-select.spec.tsx
```diff
@@ -0,0 +1,229 @@
...
+    if (select) {
+      select.value = 'paul';
+      select.dispatchEvent(new Event('change'));
+      await page.waitForChanges();
+
+      // Check if the event is emitted
+      expect(pdsSelectEvent).toHaveBeenCalled();
+    } else {
+      throw new Error('select element is not available');
+    }
```
ju-Skinner: I am not sure why you have the guard. The specs would fail should any thing be undefined.

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const assignedElements = slot.assignedElements({ flatten: true }) as HTMLOptionElement[];
+
+    assignedElements.map((item: HTMLOptionElement) => {
+      const option = item;
+
+      if (option.tagName === 'OPTION') {
+        this.selectEl.appendChild(option.cloneNode(true));
+      }
+    });
+  };
```
ju-Skinner, line 160: Will this work with `optgroup`? This is something that we need to allow.

libs/core/src/components/pds-select/pds-select.tsx
ju-Skinner: How will we handle a mult-select with `value` allowing `string`?

ju-Skinner (changes requested):
For your first PR this is a good start. I'm sorry, but basic functionality and behaviors are missing, so changes are needed.

QuintonJason (changes requested):
Great start! It's coming along but the select will need some additional considerations:
- add `multiple` prop to ensure multiple `option`s can be selected
- there currently is no `option` or `optiongroup` component. This children components maintain state through a `selected` and we'd need events for those things

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: I will remove.

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: I will remove.

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: I will add options to the examples. Same goes for the comments below.

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: I will add these. Good callout.

libs/core/src/components/pds-select/pds-select.scss
FuturaExtraBold: Will add a custom var above for this.

libs/core/src/components/pds-select/pds-select.scss
FuturaExtraBold: Will add a custom var above for this.

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: I will dig into this.

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: I will dig into this.

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: Saw this pattern other places in the repo. I figured this is the preferred way. I will opt to change to name.

libs/core/src/components/pds-select/stories/pds-select.stories.js
FuturaExtraBold: Oversight. Good callout.

libs/core/src/components/pds-select/test/pds-select.e2e.ts
FuturaExtraBold: I will dig into this.

libs/core/src/components/pds-select/test/pds-select.spec.tsx
FuturaExtraBold: I will add this hook. Good callout.

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: I will add the `multiple` prop

libs/core/src/components/pds-select/test/pds-select.spec.tsx
FuturaExtraBold: I can remove this.

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: I will add `optgroup`.

FuturaExtraBold:
@ju-Skinner @QuintonJason Thanks for the feedback! I will make the appropriate changes. I built the component based on what was in the design, but I agree completely, we need the full select functionality for our devs. Working on the changes now.

FuturaExtraBold:
Any previous examples of the callout in the description of the Accessibility violation in Storybook? Hopefully, it's something small I'm missing.

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,146 @@
+import { Component, Event, EventEmitter, Host, h, Prop } from '@stencil/core';
+import { messageId } from '../../utils/form';
+import { PdsLabel } from '../_internal/pds-label/pds-label';
+import { danger, enlarge } from '@pine-ds/icons/icons';
+
+@Component({
+  tag: 'pds-select',
+  styleUrls: ['../../global/styles/base.scss', 'pds-select.scss'],
```
anechol: We are removing the base css file from all components in a WIP PR, so this reference can be removed.

libs/core/src/components/pds-select/pds-select.tsx
anechol: ^ Second the name here.

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,146 @@
...
+                </p>
+              )}
+              {this.errorMessage && (
+                <p class="pds-select__error-message" id={messageId(this.componentId, 'error')} aria-live="assertive">
+                  <pds-icon icon={danger} size="small" />
+                  {this.errorMessage}
+                </p>
+              )}
+            </div>
+          )}
```
anechol: Is it intended to have both a helper message and error message display when both are set?

libs/core/src/components/pds-select/pds-select.scss
```diff
@@ -0,0 +1,127 @@
+:host {
+  --box-shadow-focus: 0 0 0 1px #ffffff, 0 0 0 3px var(--pine-color-purple-300);
+  --box-shadow-focus-error: 0 0 0 1px #ffffff, 0 0 0 3px var(--pine-color-red-300);
```
anechol, line 3: Thanks for adding all these custom props. This will make them much easier to update when semantic tokens are fully addressed.

libs/core/src/components/pds-select/test/pds-select.spec.tsx
```diff
@@ -0,0 +1,229 @@
...
+  });
+
+  it('renders autocomplete attribute when property is passed', async () => {
```
anechol: Can we add to test to make sure the autocomplete works?

anechol (changes requested):
Good start here! Got a few things below that weren't mentioned by others

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: Good question, Ashley! I would imagine we would want both helper and error messages for better context, but I would be open to changing if that's what we all deem the best option.

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: Done.

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: Done

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: Done

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: Done

libs/core/src/components/pds-select/docs/pds-select.mdx
FuturaExtraBold: Done

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: Done

libs/core/src/components/pds-select/pds-select.scss
FuturaExtraBold: To callout, this is specifically for the padding right in the `grid` component, so the select 'icon' (in our case 'enlarge') renders in the right spot, and doesn't block select field text. I **will** add the custom var in the vars section for this, though.

libs/core/src/components/pds-select/pds-select.scss
FuturaExtraBold: This is an oversight. I will add a custom var to the vars section.

libs/core/src/components/pds-select/test/pds-select.spec.tsx
FuturaExtraBold: This is from your refactor to fix a failing spec. I can def look into this once all the fixes/updates are applied.

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: Here's the relevant Slack thread for this, in case we want an action item: https://kajabi.slack.com/archives/C03GKQJHCLX/p1732319349764399

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: I reverted this change. I need the `.pds-select` for the element to render properly. This is the pattern I mimicked from `pds-input`. This allows for the grid layout and the icon to display properly. I reverted the tests, too, and all is well ATM. AFAICT, putting the `.pds-select` class on the `Host` element doesn't apply the styling to the shadow, hence the extra div.

libs/core/src/components/pds-select/pds-select.tsx
FuturaExtraBold: Done.

libs/core/src/components/pds-select/test/pds-select.spec.tsx
FuturaExtraBold: Added in [7202024](https://github.com/Kajabi/pine/pull/310/commits/7202024affc1c421ffc9ace802914bc3d144ac88)

FuturaExtraBold:
**Callout**: The multiple-select was not accounted for in design. Neither the [old](https://sage.kajabi.com/pages/component/form_select?tab=preview) or [new](https://www.figma.com/design/CC1YmaGKHnsvB28yLY9mEH/%E2%9D%96-Mercury-components?node-id=3459-23122) designs have multiple select added. I will be happy to add once a design decision is made. Perhaps a follow-up is best in this case.

FuturaExtraBold:
@QuintonJason I am adding the `multiple` prop as I write this message. I am curious as to why we would want new components for `option` and `optgroup` for the slotted contents rather than using the HTML elements?

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  /**
+   * Text to be displayed as the select label.
+   */
+  @Prop() label: string;
+
+  /**
+   * Indicates whether multiple options can be selected.
+   */
+  @Prop() multiple: boolean;
```
ju-Skinner: We should probably define a default value for this. e.g `false`

libs/core/src/components/pds-select/docs/pds-select.mdx
```diff
@@ -0,0 +1,287 @@
...
+#### When to use
+
+- When you want to provide a simple list of options for the user to choose from.
+- **Example**: Use a select component to choose a state from a short list of U.S. regions.
+
+#### When **NOT** to use
+
+- If your list of options is greater than 10, consider using a custom dropdown component. (WIP)
+- **Example**: Use a custom dropdown for selecting multiple filters or choosing a specific product from a comprehensive list. (WIP)
```
ju-Skinner: We can remove this.

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+@Component({
+  tag: 'pds-select',
+  styleUrl: 'pds-select.scss',
+  shadow: true,
+})
+export class PdsSelect {
+  /**
+   * A reference to the HTMLSelectElement used in the component.
+   */
```
ju-Skinner: This is truly a nitpick, but I don't think we need these comments (Lines 16-19 as well)

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,207 @@
...
+   * the assigned <option> elements from the slot.
+   */
+  private handleSlotChange = () => {
+    const slot = this.slotContainer.querySelector('slot') as HTMLSlotElement;
+
+    this.selectEl.innerHTML = '';
+    const assignedElements = slot.assignedElements({ flatten: true }) as (HTMLOptionElement | HTMLOptGroupElement)[];
+
+    assignedElements.forEach((item) => {
+      if (item.tagName === 'OPTION' || item.tagName === 'OPTGROUP') {
```
ju-Skinner: it looks like there was a change that was lost.
```suggestion
      if ( ['OPTION', 'OPTGROUP'].includes(item.tagName)) {
```

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,207 @@
...
+                </p>
+              )}
+              {this.errorMessage && (
+                <p class="pds-select__error-message" id={messageId(this.componentId, 'error')} aria-live="assertive">
+                  <pds-icon name="danger" size="small" />
+                  {this.errorMessage}
+                </p>
+              )}
+            </div>
+          )}
```
ju-Skinner: Can we clean this up?

Few options
1. Extract it out to a helper method `renderMessages()`
2. Would having an empty div cause problems? If not, then we could simply remove the double checks. L187 and (L189 or L194)

ju-Skinner (changes requested):
I have a few small requests for changes. Nothing too major.

libs/core/src/components/pds-select/pds-select.tsx
pixelflips: Done. Add to a couple of other booleans as well. Good call.

libs/core/src/components/pds-select/docs/pds-select.mdx
pixelflips: Done

libs/core/src/components/pds-select/pds-select.tsx
pixelflips: Done

libs/core/src/components/pds-select/pds-select.tsx
pixelflips: Done. Thaks for the suggestion. Not sure where it went.

libs/core/src/components/pds-select/pds-select.tsx
pixelflips: Gave this a shot, let me know.

QuintonJason (approved):
Great job with the component and tests

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,231 @@
...
+   * It updates the selected option accordingly.
+   */
+  valueChanged() {
+    this.updateSelectedOption();
+  }
+
+  /**
+   * Lifecycle method that is called just before the component is loaded.
+   * This method is used to perform any necessary updates to the selected option
+   * before the component is rendered.
```
anechol: I feel like we don't need this lifecycle method description.

libs/core/src/components/pds-select/pds-select.tsx
```diff
@@ -0,0 +1,231 @@
...
+  private renderMessages() {
+    if (!this.helperMessage && !this.errorMessage) return null;
+
+    return (
+      <div class="pds-select__message">
+        {this.getHelperMessage()}
+        {this.getErrorMessage()}
+      </div>
+    );
+  }
```
anechol, line 188: 🔥 🔥 🔥

libs/core/src/components/pds-select/test/pds-select.spec.tsx
```diff
@@ -0,0 +1,584 @@
...
+
+    // Mock the selectEl and its options property
+    const mockOption1 = { value: 'option1', selected: false } as HTMLOptionElement;
+    const mockOption2 = { value: 'option2', selected: false } as HTMLOptionElement;
+    const mockOption3 = { value: 'option3', selected: false } as HTMLOptionElement;
+    const mockOptions = [mockOption1, mockOption2, mockOption3];
+    const mockSelectEl = {
+      options: mockOptions,
+    } as unknown as HTMLSelectElement;
+    component.selectEl = mockSelectEl;
```
anechol, line 246: Non-blocker but a thought: a lot of these tests require setting up mock options. I wonder if there's a way we can set these up outside of the test and pull them in as necessary so we don't have to repeat them.

anechol (approved):
Nice work on this, just two non-blocking comments.

libs/core/src/components/pds-select/test/pds-select.spec.tsx
pixelflips: This came up in a convo with Julian as well. I think these tests will need a refactor but when I tried updating them I kept hitting other issues so will save for a follow up.

libs/core/src/components/pds-select/pds-select.tsx
pixelflips: Will remove.

pixelflips: Done.

ju-Skinner (approved):

PR #315, opened 2024-11-15
fix: updates checked property on switch update

pixelflips:
# Description

The <pds-switch> component’s visual state correctly reflects the toggle (i.e., it appears to turn on or off), but the checked property is not properly updated when the switch is clicked.

Follow-up: These updates also remove the radio variant. The previous underlying code was not working as expected and we decided as a team that its use case was not a desired UX either.

We may revisit if needed in the future, but we will need to follow up with the design team. 

Fixes https://kajabi.atlassian.net/browse/DSS-1170

## Type of change
Updates .checked property on interaction with the `pds-switch`

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-switch/docs/pds-switch.mdx
+++ b/libs/core/src/components/pds-switch/docs/pds-switch.mdx
@@ -29,15 +29,14 @@ The default switch style uses a checkbox input.
 
 <DocCanvas client:only
   mdxSource={{
-    react: '<PdsSwitch componentId="pds-switch-default-example" label="Default Switch" name="pds-switch-default" type="checkbox" />',
-    webComponent: '<pds-switch component-id="pds-switch-default-example" label="Default Switch" name="pds-switch-default" type="checkbox"></pds-switch>'
+    react: '<PdsSwitch componentId="pds-switch-default-example" label="Default Switch" name="pds-switch-default" />',
+    webComponent: '<pds-switch component-id="pds-switch-default-example" label="Default Switch" name="pds-switch-default"></pds-switch>'
   }}>
   <div>
     <pds-switch
       component-id="pds-switch-default-example"
       label="Default Switch"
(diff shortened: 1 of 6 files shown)


monicawheeler (approved):

monicawheeler:
Looks good! One non-blocking request, could you add the "Why" to the description to explain why we are removing radio? This way the history is tracked with the code change. Thanks!

monicawheeler (approved):

libs/core/src/components/pds-switch/docs/pds-switch.mdx
```diff
@@ -102,7 +76,7 @@ Displays additional descriptive text underneath the main label.
       label="Remember me"
       name="pds-switch-helper"
       type="checkbox"
```
ju-Skinner: Need to remove `type` from the examples, since the property has been removed

ju-Skinner (changes requested):
see comment about removing `type` from the examples, since the property is no longer valid.

libs/core/src/components/pds-switch/docs/pds-switch.mdx
pixelflips: @ju-Skinner i think the type attribute is still needed on the underlying input though, no? How else would it be a checkbox?

As in:

`<input type="checkbox"`...

pixelflips:
> see comment about removing `type` from the examples, since the property is no longer valid.

@ju-Skinner Should be updated now, good catch!

ju-Skinner (approved):
great job :shipit:

PR #326, opened 2024-12-05
feat(popover): add popover component

QuintonJason:
# Description
- [x] add popover component
- [x] add popover spec tests
- [x] add popover e2e tests
- [x] add popover docs

Fixes [DSS-1146](https://kajabi.atlassian.net/browse/DSS-1146)

### Preview
popover-api

## Type of change

# How Has This Been Tested?

Visit [Popover docs page](http://localhost:6006/?path=/docs/components-popover--docs)

- [x] unit tests
- [x] e2e tests
- [x] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-1146]: https://kajabi.atlassian.net/browse/DSS-1146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/src/components/pds-popover/docs/pds-popover.mdx
```diff
@@ -0,0 +1,94 @@
+import { DocArgsTable, DocCanvas } from '@pine-ds/doc-components';
+import { components } from '../../../../dist/docs.json';
+
+# Popover
+Popovers display contextual information when users click or fires a keyboard event on an element. They typically contain more information than tooltips and can include interactive elements.
```
ju-Skinner: `... users click or initiate a keyboard event ...`

libs/core/src/components/pds-popover/pds-popover.tsx
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Text that appears on the trigger element
+   */
+  @Prop() text: string;
+
+  /**
+   * Sets the maximum width of the popover content
+   * @defaultValue "352px"
+   */
+  @Prop() maxWidth?: string = '352px';
```
ju-Skinner: Should we make this a number to assist with the validation? We can append the `unit type`

libs/core/src/components/pds-popover/pds-popover.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Determines the preferred position of the popover
+   * @defaultValue "right"
+   */
+  @Prop({ mutable: true, reflect: true }) placement: TooltipPlacementType = 'right';
```
ju-Skinner: `TooltipPlacementType` should probably be refactored to be called `PlacementType`.

libs/core/src/components/pds-popover/pds-popover.tsx
```diff
@@ -0,0 +1,220 @@
...
+   * Determines when the popover is active
+   * @defaultValue false
+   */
+  @State() active = false;
+
+  /**
+   * Determines the action that triggers the popover
+   * @defaultValue "show"
+   */
+  @Prop() popoverTargetAction: 'show' | 'hide' = 'show';
```
ju-Skinner, line 25: I believe we should add some additional documentation around this property.

ju-Skinner: I believe we should add some additional documentation around this property.

libs/core/src/components/pds-popover/pds-popover.scss
```diff
@@ -0,0 +1,63 @@
...
+    border: var(--border-width-default) solid var(--color-border-default);
+    border-radius: var(--pine-border-radius-circle);
+    color: var(--color-text-default);
+    display: flex;
+    font: var(--typography-default);
+    min-height: 40px;
+    padding: var(--spacing-padding-block) var(--spacing-padding-inline);
+  }
+
+}
```
monicawheeler: Missing eol

libs/core/src/components/pds-popover/pds-popover.scss
```diff
@@ -0,0 +1,63 @@
...
+      background-color: var(--color-background-secondary-hover);
+    }
+
+    align-items: center;
+    background-color: var(--color-background-default);
+    border: var(--border-width-default) solid var(--color-border-default);
+    border-radius: var(--pine-border-radius-circle);
+    color: var(--color-text-default);
+    display: flex;
+    font: var(--typography-default);
```
ju-Skinner: This is a nitpick, due to what I know of semantic tokens. Can you expand this to be 

```
font-family:
font-weight:
line-height: 
...
```

monicawheeler:
Do the shadow styles matching the [Figma](https://www.figma.com/design/CC1YmaGKHnsvB28yLY9mEH/%E2%9D%96-Mercury-components?node-id=3686-23689&node-type=frame&t=QIqVrtRunqUVCCwf-0)? It looks like we could use a bit more depth at the top of the popover panel.

libs/core/src/components/pds-popover/docs/pds-popover.mdx
```diff
@@ -0,0 +1,94 @@
...
+    webComponent: `
+      <pds-popover component-id="manual" text="Manual Popover" popover-type="manual">
+        <p>This popover requires explicit closing</p>
+        <button>Close Popover</button>
+      </pds-popover>
+    `
+  }}>
+  <pds-popover component-id="manual" text="Manual Popover" popover-type="manual">
+    <p>This popover requires explicit closing</p>
+    <button>Close Popover</button>
```
pixelflips: Clicking this button does not close the popover in the demo

libs/core/src/components/pds-popover/pds-popover.scss
pixelflips: nitpick. file has no new line at end

libs/core/src/components/pds-popover/pds-popover.tsx
```diff
@@ -0,0 +1,220 @@
...
+      popoverElement.hidePopover();
+    }
+
+    this.hidePdsPopover.emit();
+  }
+
+  render() {
+    return (
+      <Host>
+        <button
```
pixelflips, line 115: Do we plan to eventually allow different types of triggers? It looks like we can only pass text, not an icon, for example. Maybe I am missing something.

libs/core/src/components/pds-popover/stories/pds-popover.stories.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const Manual = BaseTemplate.bind({});
+Manual.args = {
+	componentId: 'popover-1',
+  placement: "right",
+	popoverType: "manual",
+  text: "Show popover",
+	slot: html`<p>Popover content</p>
+    <p>Popover content</p>`
+};
```
pixelflips: nitpick. file is missing new line at end of file

libs/core/src/components/pds-popover/pds-popover.tsx
pixelflips, line 115: The trigger seems to only show the popover vs toggling. Meaning I can open it with the trigger, but clicking the trigger again does not close it. This may be intended so just let me know.

pixelflips (commented):
Left a few comments, let me know if any may not apply at this time.

libs/core/src/components/pds-popover/pds-popover.scss
pixelflips: Looks like you fixed before I submitted my review.

libs/core/src/components/pds-popover/pds-popover.tsx
```diff
@@ -0,0 +1,220 @@
...
+  handleKeyDown(event: KeyboardEvent) {
+    if (event.key === 'Enter' || event.key === ' ' && !this.active) {
+      if (this.active) {
+        event.stopPropagation();
+        return;
+      }
+      const closestTarget = (event.target as HTMLElement).shadowRoot.querySelector(".pds-popover__trigger");
+      if (closestTarget) {
+        event.preventDefault();
+        event.stopPropagation();
```
ju-Skinner: I think we need a `return` here

QuintonJason:
> Do the shadow styles matching the [Figma](https://www.figma.com/design/CC1YmaGKHnsvB28yLY9mEH/%E2%9D%96-Mercury-components?node-id=3686-23689&node-type=frame&t=QIqVrtRunqUVCCwf-0)? It looks like we could use a bit more depth at the top of the popover panel.

@monicawheeler  I reached out to design as the shadows have been update since I made this change. Awaiting design feedback

libs/core/src/components/pds-popover/docs/pds-popover.mdx
QuintonJason: Great callout. This is a known issue that I'm currently working through.

ju-Skinner (changes requested):

libs/core/src/components/pds-popover/pds-popover.tsx
QuintonJason: Based on the Popover API, the trigger has 3 available props: `show`, `toggle`, and `hide`. Currently, the shadow is configured to `show`, but not `toggle`. Adding `toggle` into the mix allows that expected behavior, but I intentionally left it off because `show` is the default trigger behavior. Thoughts?

QuintonJason: The original attempt at the popover allowed for anything, but there were underlying complexities. This implementation was created based on the popover api and requires full control over the trigger. Thoughts?

pixelflips: That's super helpful to know and now I am not sure. It felt odd to me that I couldn't toggle it and had to explicitly move my mouse away and then click, but maybe that's just me. Maybe it's something we can bring up with the group if it pops up in the future.

QuintonJason: Added

pixelflips (approved):
Outside of the linting error and the trigger follow-up, LGTM! Nice work! 👍🏼

anechol (changes requested):
Thanks for putting in all the discovery work to get this working. I have a point of feedback that may evolve into a larger discussion but wanted to get your thoughts on it:

According to MDN and the HTML spec, a popover is any element that has the `popover` attribute attached to it that can be triggered by a popover action (typically a button). This can be used for any number of common components like tooltips, dropdowns, and modals (in the case of modals, the element should be a dialog). With the way this component is built, is there any way we can use this component to cover all of those cases if a button is already attached? The reason I bring this up is two-fold:

1) I can foresee engineers wanting to roll their own trigger button, whether it's the `pds-button` component or something else.
2) It would be great for us to reuse this component for the aforementioned cases of tooltips, dropdowns, and modals, but I don't think it's possible with this setup without making extra props, e.g. a modal for the popover container to become a dialog tag.

These points are both related and bring up some concerns: there is no way to trigger a vanilla popover with the `pds-button`. This example:

```html
<pds-button popovertarget="my-popover">Click Me</pds-button>
<div popover="auto" id="my-popover">Popover Content</div>
```

...does not work because the popover element (div) and the popover trigger (the actual button within pds-button) are not in the same tree (hello, Shadow DOM!). The only way this can work would be to remove the shadow DOM on the `pds-button` element. This point matters because if our button component can't trigger the API without such a drastic change, then we aren't able to decouple the popover trigger button from the popover element. And if we can't decouple those two, then the component isn't as reusable in the way I'm thinking it could be.

It would be awesome for us to leverage as much native browser technology as possible given the advantages it gives us in simplicity and accessibility, but I genuinely don't see a workaround here in the effort of reusability. Do you have any thoughts here? Always open to pairing more if need be.

anechol (approved):
Once again, thanks for the large effort in discovery and research work for this component. 

As discussed, we can move forward with this and get it over the finish line. The only thing left for me here is the popover not closing when the button is clicked. Other than that small thing, awesome work!

libs/core/src/components/pds-popover/docs/pds-popover.mdx
ju-Skinner: @QuintonJason Have you resolved this concern?

libs/core/src/components/pds-popover/pds-popover.tsx
```diff
@@ -0,0 +1,134 @@
...
+   * Determines when the popover is active
+   * @defaultValue false
+   */
+  @State() active = false;
+
+  /**
+   * Determines the action that triggers the popover
+   * @defaultValue "show"
+   */
+  @Prop() popoverTargetAction: 'show' | 'toggle' = 'show';
```
anechol: I think we can add a bit of context here to explain that this only works if the popover type is set to manual.

libs/core/src/components/pds-popover/docs/pds-popover.mdx
```diff
@@ -0,0 +1,157 @@
...
+    - Important: Ensure the id (`manual` in this example) matches the componentId prop of the `PdsPopover` component.
+
+2. Attach Function to onClick:
+
+    - Assign the `handleManualClose` function to the `onClick` event of the button within the popover.
+    - This ensures that when the user clicks the "Close Popover" button, the `handleManualClose` function is executed, and the popover is hidden. */}
+
+export const handleManualClose = () => {
+  const popover = document.querySelector('pds-popover[component-id="manual"]');
+  popover.hide();
```
ju-Skinner: is this still valid? I don't see a `hide` method on the `PdsPopover` class.

libs/core/src/components/pds-popover/docs/pds-popover.mdx
```diff
@@ -0,0 +1,157 @@
...
+    - Define a JavaScript function (e.g., `handleManualClose`) that will be responsible for closing the popover.
+    - Inside this function, you'll typically use `document.getElementById('manual').hide()` to call the `hide()` method on the popover element.
+    - Important: Ensure the id (`manual` in this example) matches the componentId prop of the `PdsPopover` component.
+
+2. Attach Function to onClick:
+
+    - Assign the `handleManualClose` function to the `onClick` event of the button within the popover.
+    - This ensures that when the user clicks the "Close Popover" button, the `handleManualClose` function is executed, and the popover is hidden. */}
+
+export const handleManualClose = () => {
```
ju-Skinner: should this function be referenced in the example below `<pds-popover onClick="handleManualClose" ...>`

ju-Skinner (commented):
I left a few comments related to the documentation. There may be some code changes, depending on your response.

libs/core/src/components/pds-popover/docs/pds-popover.mdx
QuintonJason: removed

libs/core/src/components/pds-popover/docs/pds-popover.mdx
QuintonJason: removed

ju-Skinner (approved):

PR #331, opened 2024-12-12
docs(button): updates Button docs to match other components

anechol:
# Description

Updates Button documentation to match other components.

[DSS-1177](https://kajabi.atlassian.net/browse/DSS-1177)

## Type of change

- [x] Documentation (updates component documentation)

[DSS-1177]: https://kajabi.atlassian.net/browse/DSS-1177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/components/pds-button/docs/pds-button.mdx
+++ b/libs/core/src/components/pds-button/docs/pds-button.mdx
@@ -26,6 +26,10 @@ While accessible out of the box, there are some best practices to keep in mind w
 - Buttons should not disable tabbing or keyboard navigation.
 - Buttons should always pass [color contrast best practices](https://www.w3.org/WAI/WCAG22/Understanding/contrast-minimum.html).
 
+### Buttons vs Links
+
+In many current web design patterns, buttons and links have evolved to look visually identical. However, for best practices, the fundamental differences between the two should be considered before using. Links are used to navigate the user to another page. Buttons are used to perform an action on the current page or submit a form.
+
 ## Properties
 
 <DocArgsTable componentName="pds-button" docSource={components}/>
--- a/libs/core/src/components/pds-button/docs/pds-button.mdx
+++ b/libs/core/src/components/pds-button/docs/pds-button.mdx
@@ -3,15 +3,15 @@ import { components } from '../../../../dist/docs.json';
 
 # Button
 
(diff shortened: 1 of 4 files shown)


libs/core/src/components/pds-button/pds-button.tsx
```diff
@@ -4,9 +4,9 @@ import { hasShadowDom } from '../../utils/utils';
 import { caretDown } from '@pine-ds/icons/icons';
 
 /**
- * @part button - The main button element that represents the button component.
- * @part caret - The caret icon element that appears when the button variant is 'disclosure'.
- * @part icon - The icon element that appears before the text in the button, if provided.
+ * @part button - Button element styles.
+ * @part caret - Caret icon styles. Appears only on the disclosure variant.
+ * @part icon - Icon element styles.
```
ju-Skinner: Is `styles` the correct term here? 

mdn reference: https://developer.mozilla.org/en-US/docs/Web/API/Element/part

ju-Skinner (changes requested):
the verbiage for `@part` is incorrect

pixelflips (approved):

libs/core/src/components/pds-button/pds-button.tsx
anechol: My thought process here is that this is used to change the part's styles, as in style declarations, since they are intended to be targeted with CSS. If that isn't clear, I can reword all instances of part explanations.

libs/core/src/components/pds-button/docs/pds-button.mdx
```diff
@@ -32,9 +32,9 @@ When following our usage guidelines, pds-button should be accessible right out o
 
 ## Variant
```
QuintonJason: remove title for consistency

ju-Skinner: +1 It was removed in the Link PR #343

libs/core/src/components/pds-button/pds-button.tsx
ju-Skinner: While yes that is also a use case, however, this is intended for the code.

QuintonJason (approved):

libs/core/src/components/pds-button/docs/pds-button.mdx
anechol: This has been removed.

libs/core/src/components/pds-button/pds-button.tsx
anechol: I've updated the description, let me know what you think.

ju-Skinner (approved):

PR #386, opened 2025-02-28
ci: getDX Deployments and Service identification

voodooGQ:
https://github.com/Kajabi/sage-lib/pull/2073

Refer to that PR above as it goes over most of the details. This just ensures we report deployments and services via PRs to GetDX

### DIFF
--- a/libs/doc-components/catalog-info.yaml
+++ b/libs/doc-components/catalog-info.yaml
@@ -1,12 +1,12 @@
 apiVersion: backstage.io/v1alpha1
 kind: Component
 metadata:
-  name: pine-react
-  description: The React library for the Pine design system
+  name: pine-doc-components
+  description: The Documentation Components library for the Pine design system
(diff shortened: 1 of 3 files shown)


libs/doc-components/catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: pine-react
```
ju-Skinner: Should this read "pine-doc-components?"

ju-Skinner (approved):
If the catalog info is correct, then please feel free to merge.

libs/doc-components/catalog-info.yaml
voodooGQ: I must not have hit save before pushing 😆 

Addressing. Thank you!

PR #387, opened 2025-02-28
style: add 125 token

QuintonJason:
# Description
- [x] update pine tokens to include dimenion-125

### DIFF
--- a/libs/core/src/global/styles/tokens/base/_core.scss
+++ b/libs/core/src/global/styles/tokens/base/_core.scss
@@ -143,6 +143,7 @@
   --pine-line-height-075: 0.9375;
   --pine-dimension-0: 0px;
   --pine-dimension-100: 8px;
+  --pine-dimension-125: 10px;
   --pine-dimension-150: 12px;
   --pine-dimension-200: 16px;
   --pine-dimension-250: 20px;
--- a/libs/core/src/components/pds-chip/pds-chip.scss
+++ b/libs/core/src/components/pds-chip/pds-chip.scss
@@ -1,7 +1,7 @@
 :host {
   --box-shadow-focus: 0 0 0 1px #ffffff, 0 0 0 3px var(--pine-color-purple-300);
 
-  --sizing-close: 10px;
+  --sizing-close: var(--pine-dimension-125);
 
   align-items: center;
   border-radius: var(--pine-dimension-sm);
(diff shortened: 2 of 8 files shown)


libs/core/src/global/styles/tokens/base/_core.scss
```diff
@@ -143,6 +143,7 @@
   --pine-line-height-075: 0.9375;
   --pine-dimension-0: 0px;
   --pine-dimension-100: 8px;
+  --pine-dimension-125: 10px;
```
pixelflips, line 146: Should components where 10px is hard coded be updated? Also, we'll need to ensure this gets updated in Token Studio.  (I think it may just be a pull in Token Studio once merged, but please confirm with @anechol )

anechol: Tokens Studio is all good, we paired on getting that updated. The main branch will just need to be pulled in as normal when this is merged in.

monicawheeler (changes requested):
My understanding is that our dimension scale is representative of 4px increments, how does 10px fit into this scale?

pixelflips:
> My understanding is that our dimension scale is representative of 4px increments, how does 10px fit into this scale?

@monicawheeler Some additional info:

https://kajabi.slack.com/archives/C03GKQJHCLX/p1740428421688799
https://kajabi.slack.com/archives/C03GKQJHCLX/p1740763001181419

QuintonJason:
> My understanding is that our dimension scale is representative of 4px increments, how does 10px fit into this scale?

@monicawheeler  Yes, it doesn't fit into our 4 pt scale, but it exists in the system and `kp`. This PR was created to resolve an issue with devs not being able to find a token for that value. 

If we don't support `10px`, I see the remedy as one of 2 things:

1. Design updates all references from `10px` to a 4X value
2. We add `10px`

I leaned in favor of choosing 125 because one of the reasons we initially created this numbered scale was to have a better way to handle "in-between" values, since t-shirt sizes were too strict. This was a case where our number scale had an appropriate "in-between" value to account for that value. We currently have `025` to represent `2px`, so we already have a case in our 4X where that is not the case.

monicawheeler (approved):
Thank you everyone for the clarification.

pixelflips (approved):

PR #391, opened 2025-03-03
style: add conditional label for input and select

QuintonJason:
# Description
There are uses of the input in which the label is not visible. In those cases the spacing properties assigned to the label remain as an artifact

- [x] Input - update conditional label element when prop not preset
- [x] Select - update conditional label element when prop not preset

Fixes #(issue)

#### Input 
|  Before  |  After  |
|--------|--------|
|Screenshot 2025-03-04 at 1 18 01 PM|Screenshot 2025-03-04 at 1 16 59 PM|

#### Select
|  Before  |  After  |
|--------|--------|
|Screenshot 2025-03-04 at 1 10 50 PM|Screenshot 2025-03-04 at 1 13 00 PM|

## Type of change

# How Has This Been Tested?

- Visit either the Input or Select page
- In the `Playground`, remove the `label`
- Inspect the element and verify that the empty `<label>` element is no longer present

`Textarea` already has this structure

- [x] unit tests
- [x] tested manually

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-input/pds-input.scss
+++ b/libs/core/src/components/pds-input/pds-input.scss
@@ -17,7 +17,7 @@
   flex-direction: column;
 }
 
-label {
+.pds-input__label {
   margin-block-end: var(--pine-dimension-2xs)
 }
 
--- a/libs/core/src/components/pds-input/pds-input.scss
+++ b/libs/core/src/components/pds-input/pds-input.scss
@@ -15,7 +15,10 @@
 .pds-input {
   display: flex;
   flex-direction: column;
-  gap: var(--pine-dimension-2xs);
+}
+
+label {
+  margin-block-end: var(--pine-dimension-2xs)
 }
 
 .pds-input__field {
@@ -97,7 +100,7 @@
 .pds-input__helper-message {
   font: var(--pine-typography-body-sm-medium);
   margin-block-end: 0;
-  margin-block-start: 0;
+  margin-block-start: var(--pine-dimension-2xs)
 }
 
 .pds-input__error-message {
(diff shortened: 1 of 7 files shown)


libs/core/src/components/pds-input/pds-input.tsx
```diff
@@ -256,7 +256,9 @@ export class PdsInput {
         aria-readonly={this.readonly ? 'true' : null}
       >
         <div class="pds-input">
-          <label htmlFor={this.componentId}>{this.label}</label>
+          {this.label &&
+            <label htmlFor={this.componentId}>{this.label}</label>
+          }
```
ju-Skinner: I'm curious if we still want his approach, with how we just changed the checkbox to wrap the input with `label.`. Thoughts?

QuintonJason: Yes, this approach is still valid. I don't believe I've ever seen a label wrap an input not of type `checkbox`, `radio`. I believe those are the only ones that get that treatment because the text is typically inline with the input on either side. The nesting would cause the need for additional `div`s to have control over layout

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -17,6 +17,10 @@
   flex-direction: column;
 }
 
+label {
```
pixelflips: It is not a blocker, but this component uses classes for most other element styling. Should we do the same for the label?

pixelflips (approved):
One non-blocker, otherwise LGTM! 👍🏼

ju-Skinner (approved):

PR #426, opened 2025-04-02
feat(text): expands text color prop

anechol:
# Description

Adds support for more color possibilities for the color prop.

[DSS-1348](https://kajabi.atlassian.net/browse/DSS-1348)

## Type of change
Before:
Screenshot 2025-04-02 at 5 00 48 PM

After:
Screenshot 2025-04-02 at 5 00 05 PM

# How Has This Been Tested?

- [x] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-1348]: https://kajabi.atlassian.net/browse/DSS-1348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/src/components/pds-text/docs/pds-text.mdx
```diff
@@ -61,39 +61,42 @@ Text can be aligned to the start, center, end, or justified.
...
     react: `
-      <PdsText style="background-color: #33; padding-inline: 8px" color="primary" tag="p">Zebstrika</PdsText>
-      <PdsText color="secondary" tag="p">Pangoro</PdsText>
-      <PdsText color="neutral" tag="p">Urshifu</PdsText>
-      <PdsText color="accent" tag="p">Sableye</PdsText>
-      <PdsText color="danger" tag="p">Arcanine</PdsText>
-      <PdsText color="info" tag="p">Garchomp</PdsText>
-      <PdsText color="success" tag="p">Sceptile</PdsText>
-      <PdsText color="warning" tag="p">Pikachu</PdsText>
+      <PdsText style="background-color: #33; padding-inline: 8px" color="primary" tag="h5">Zebstrika</PdsText>
```
pixelflips: I think the inline styling syntax here may be backwards with the web component. Meaning react uses the camelCase.

libs/core/src/components/pds-text/pds-text.scss
```diff
@@ -81,29 +76,15 @@ h1, h2, h3, h4, h5, h6, p, code, pre, em, strong {
...
 
-@mixin generate-type-colors($type-colors) {
-  @each $key, $value in $type-colors {
-    .pds-text--color-#{$key} {
-      color: $value;
-    }
-  }
+// Variable is set in TSX
+// If color prop value isn't valid, will default to body text color
+.pds-text {
```
pixelflips, line 83: Not a showstopper, but could we just default the prop to `primary` and reuse that styling?

pixelflips (approved):
A couple of small comments to consider; otherwise, nice update. 👍🏼

libs/core/src/components/pds-text/pds-text.scss
anechol: Do you mean here in the CSS or in the tsx file?

pixelflips: The tsx file.

anechol: We definitely could and that would just mean the style prop would always be set as opposed to the CSS fallback. So to me it really feels like a preference thing. The only concern I have would be the rare instance where the custom prop is undefined, but that would be an issue on either side unless we decide to fall back to a non custom prop value.

libs/core/src/components/pds-text/docs/pds-text.mdx
anechol: Fixed

QuintonJason (approved):

libs/core/src/components/pds-text/pds-text.tsx
```diff
@@ -23,7 +23,8 @@ export class PdsText {
   | 'danger'
   | 'info'
   | 'success'
-  | 'warning';
+  | 'warning'
+  | string;
```
ju-Skinner: This could be written as the suggestion below. This is possible because, in your setColor function, you are predefining a list of look values. 

```suggestion
@Prop() color?: string
```

ju-Skinner (approved):
Per our convo via Tuple, apply the diff and this is good to go.

libs/core/src/components/pds-text/pds-text.tsx
anechol: Thanks for the help on refactoring this

PR #428, opened 2025-04-10
fix(input): add prefix, suffix, prepend, and append

QuintonJason:
# Description
- [x] add a slots the to input for interactive and non-interactive elements

Fixes [DSS-1311](https://kajabi.atlassian.net/browse/DSS-1311))

Screenshot 2025-04-14 at 10 05 33 AM

## Type of change

- [x} This change requires a documentation update

# How Has This Been Tested?
Visit the Input page -> Prefix section and below

- [x] unit tests

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-1311]: https://kajabi.atlassian.net/browse/DSS-1311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/components/pds-button/pds-button.scss
+++ b/libs/core/src/components/pds-button/pds-button.scss
@@ -1,8 +1,11 @@
 :host {
   --pds-button-background: var(--color-background-default);
   --pds-button-border: var(--pine-border);
-  --pds-button-border-radius-start: var(--pine-border-radius-full);
-  --pds-button-border-radius-end: var(--pine-border-radius-full);
+  --pds-button-border-radius: var(--pine-border-radius-full);
+  --pds-button-border-radius-start-end: var(--pine-border-radius-full);
+  --pds-button-border-radius-start-start: var(--pine-border-radius-full);
+  --pds-button-border-radius-end-end: var(--pine-border-radius-full);
+  --pds-button-border-radius-end-start: var(--pine-border-radius-full);
   --pds-button-min-height: 40px;
   --color-border-default: transparent;
   --color-border-disabled: transparent;
@@ -43,11 +46,12 @@
   align-items: center;
   background-color: var(--pds-button-background);
   border: var(--pds-button-border);
-  border-bottom-left-radius: var(--pds-button-border-radius-start);
-  border-bottom-right-radius: var(--pds-button-border-radius-end);
-  border-top-left-radius: var(--pds-button-border-radius-start);
-  border-top-right-radius: var(--pds-button-border-radius-end);
+  border-radius: var(--pds-button-border-radius);
+  border-bottom-left-radius: var(--pds-button-border-radius-end-start, var(--pds-button-border-radius));
+  border-bottom-right-radius: var(--pds-button-border-radius-end-end, var(--pds-button-border-radius));
   border-color: var(--color-border-default);
+  border-top-left-radius: var(--pds-button-border-radius-start-start, var(--pds-button-border-radius));
+  border-top-right-radius: var(--pds-button-border-radius-start-end, var(--pds-button-border-radius));
   box-sizing: border-box;
   color: var(--color-text-default);
   cursor: pointer;
--- a/libs/core/src/components/pds-button/pds-button.scss
+++ b/libs/core/src/components/pds-button/pds-button.scss
@@ -1,5 +1,9 @@
 :host {
-  --button-dimension: 40px;
+  --pds-button-background: var(--color-background-default);
+  --pds-button-border: var(--pine-border);
+  --pds-button-border-radius-start: var(--pine-border-radius-full);
+  --pds-button-border-radius-end: var(--pine-border-radius-full);
+  --pds-button-min-height: 40px;
   --color-border-default: transparent;
   --color-border-disabled: transparent;
   --color-border-focus: transparent;
@@ -37,18 +41,22 @@
 .pds-button {
   --pds-loader-color: var(--color-text-default);
   align-items: center;
-  background-color: var(--color-background-default);
-  border: var(--pine-border);
+  background-color: var(--pds-button-background);
+  border: var(--pds-button-border);
+  border-bottom-left-radius: var(--pds-button-border-radius-start);
+  border-bottom-right-radius: var(--pds-button-border-radius-end);
+  border-top-left-radius: var(--pds-button-border-radius-start);
+  border-top-right-radius: var(--pds-button-border-radius-end);
   border-color: var(--color-border-default);
-  border-radius: var(--pine-border-radius-full);
   box-sizing: border-box;
-  color: var(--color-text-default); // Set in the variant classes below
+  color: var(--color-text-default);
   cursor: pointer;
   display: flex;
   font: var(--pine-typography-body-brand-label);
   letter-spacing: var(--pine-letter-spacing);
-  min-height: var(--button-dimension);
+  min-height: var(--pds-button-min-height);
   padding: var(--pine-dimension-xs) var(--pine-dimension-sm);
+  position: relative;
   text-decoration: none;
 
   pds-icon {
--- a/libs/core/src/components/pds-button/pds-button.scss
+++ b/libs/core/src/components/pds-button/pds-button.scss
@@ -44,14 +44,14 @@
 .pds-button {
   --pds-loader-color: var(--color-text-default);
   align-items: center;
-  background-color: var(--pds-button-background);
+  background-color: var(--pds-button-background, var(--color-background-default));
   border: var(--pds-button-border);
-  border-radius: var(--pds-button-border-radius);
-  border-bottom-left-radius: var(--pds-button-border-radius-end-start, var(--pds-button-border-radius));
-  border-bottom-right-radius: var(--pds-button-border-radius-end-end, var(--pds-button-border-radius));
   border-color: var(--color-border-default);
-  border-top-left-radius: var(--pds-button-border-radius-start-start, var(--pds-button-border-radius));
-  border-top-right-radius: var(--pds-button-border-radius-start-end, var(--pds-button-border-radius));
+  border-end-end-radius: var(--pds-button-border-radius-end-end, var(--pds-button-border-radius));
+  border-end-start-radius: var(--pds-button-border-radius-end-start, var(--pds-button-border-radius));
+  border-radius: var(--pds-button-border-radius);
+  border-start-end-radius: var(--pds-button-border-radius-start-end, var(--pds-button-border-radius));
+  border-start-start-radius: var(--pds-button-border-radius-start-start, var(--pds-button-border-radius));
   box-sizing: border-box;
   color: var(--color-text-default);
   cursor: pointer;
--- a/libs/core/src/components/pds-input/docs/pds-input.mdx
+++ b/libs/core/src/components/pds-input/docs/pds-input.mdx
@@ -113,6 +113,25 @@ When `invalid` is true, the input displays the error message and provides a visu
   <pds-input component-id="pds-input-error-example" label="Email address" type="email" error-message="Please provide a valid email address" value="Frank Dux" invalid></pds-input>
 </DocCanvas>
 
+### Prefix and Suffix
+A prefix or suffix can be added to the input field to provide additional context or formatting.
+
+<DocCanvas client:only
+  mdxSource={{
+    react: '<PdsInput componentId="pds-input-prefix-suffix-example" label="Name" type="email" value="Frank Dux"><span slot="prefix">@</span><span slot="suffix">.com</span></PdsInput>',
+    webComponent: '<pds-input component-id="pds-input-prefix-suffix-example" label="Name" type="email" value="Frank Dux"><span slot="prefix">@</span><span slot="suffix">.com</span></pds-input>'
+  }}>
+  <pds-input component-id="pds-input-prefix-suffix-example" label="Name" type="email" value="Frank Dux">
+    <span slot="prefix">@</span>
+    <span slot="suffix">.com</span>
+  </pds-input>
+</DocCanvas>
+
+<pds-input component-id="pds-input-prefix-suffix-example" label="Name" type="email" value="Frank Dux">
+  <pds-icon slot="prefix" name="mail"></pds-icon>
+  <pds-button slot="suffix" variant="secondary"><pds-icon name="add-small"></pds-icon></pds-button>
+</pds-input>
+
 ## References
 
 [MDN Web Docs: Input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input)
--- a/libs/core/src/components/pds-input/docs/pds-input.mdx
+++ b/libs/core/src/components/pds-input/docs/pds-input.mdx
@@ -114,7 +114,7 @@ When `invalid` is true, the input displays the error message and provides a visu
 </DocCanvas>
 
 ### Prefix and Suffix
-A prefix or suffix can be added to the input field to provide additional context or formatting.
+A prefix or suffix can be added to the input field to provide additional context or formatting. These elements are placed within the input container, directly adjacent to the input field.
 
 <DocCanvas client:only
   mdxSource={{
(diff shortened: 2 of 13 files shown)


libs/core/src/components/pds-button/pds-button.scss
```diff
@@ -1,5 +1,12 @@
 :host {
-  --button-dimension: 40px;
+  --pds-button-background: var(--color-background-default);
+  --pds-button-border: var(--pine-border);
+  --pds-button-border-radius: var(--pine-border-radius-full);
+  --pds-button-border-radius-start-end: var(--pine-border-radius-full);
+  --pds-button-border-radius-start-start: var(--pine-border-radius-full);
+  --pds-button-border-radius-end-end: var(--pine-border-radius-full);
+  --pds-button-border-radius-end-start: var(--pine-border-radius-full);
+  --pds-button-min-height: 40px;
```
ju-Skinner, line 9: Do we need to define these here? If so, do we want consumers of this component to overwrite them?

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -1,7 +1,63 @@
...
+  --pds-input-icon-color: var(--pine-color-text-secondary);
+  --pds-input-disabled-background: var(--pine-color-background-container-disabled);
+  --pds-input-disabled-text-color: var(--pine-color-text-disabled);
+  --pds-input-error-color: var(--pine-color-text-message-danger);
+  --pds-input-error-background: var(--pine-input-color-background-danger);
+  --pds-input-error-border: var(--pine-color-border-danger);
+  --pds-input-addon-background: var(--pine-color-background-subtle);
+  --pds-input-addon-color: var(--pine-color-text-secondary);
+  --box-shadow-focus: 0 0 0 1px #ffffff, 0 0 0 3px var(--pine-color-focus-ring);
+  --box-shadow-focus-error: 0 0 0 1px #ffffff, 0 0 0 3px var(--pine-color-focus-ring-danger);
```
ju-Skinner, line 25: I have the same question as I asked above on `pds-button.scss`

libs/core/src/components/pds-input/docs/pds-input.mdx
```diff
@@ -113,8 +113,82 @@ When `invalid` is true, the input displays the error message and provides a visu
...
+A prefix or suffix can be added to the input field to provide additional context or formatting. These elements are placed within the input container, directly adjacent to the input field.
+
+<DocCanvas client:only
+  mdxSource={{
+    react: '<PdsInput componentId="pds-input-prefix-suffix-example" label="Name" type="email" value="Frank Dux"><pds-icon name="mail" slot="prefix"></pds-icon><pds-button class="pds-input__suffix" variant="unstyled" slot="suffix"><pds-icon name="remove-circle"></pds-icon></pds-button></PdsInput>',
+    webComponent: '<pds-input component-id="pds-input-prefix-suffix-example" label="Name" type="email" value="Frank Dux"><pds-icon name="mail" slot="prefix"></pds-icon><pds-button class="pds-input__suffix" variant="unstyled" slot="suffix"><pds-icon name="remove-circle"></pds-icon></pds-button></pds-input>'
+  }}>
+  <pds-input component-id="pds-input-prefix-suffix-example" label="Name" type="email" value="Frank Dux">
+    <pds-icon name="mail" slot="prefix"></pds-icon>
+    <pds-button class="pds-input__suffix" variant="unstyled" slot="suffix"><pds-icon name="remove-circle"></pds-icon></pds-button>
```
pixelflips: You could use the `icon-only` and `icon` props instead of the `pds-icon` for the content. This will allow for adding text within the button as well.

```
<pds-button class="pds-input__suffix" variant="unstyled" slot="suffix" icon-only icon="remove-circle">Remove</pds-button>
```

libs/core/src/components/pds-input/docs/pds-input.mdx
```diff
@@ -113,8 +113,82 @@ When `invalid` is true, the input displays the error message and provides a visu
...
+<pds-input component-id="pds-input-prefix-suffix-example-2" label="Name" type="email" value="Frank Dux">
+  <pds-button class="pds-input__suffix" variant="unstyled" slot="suffix"><pds-icon name="remove-circle"></pds-icon></pds-button>
+</pds-input>
+
+**Box in prepend and append slots**
+<pds-input component-id="pds-input-prefix-suffix-example-2" label="Name" type="email" value="Frank Dux">
+  <pds-icon name="mail" slot="prefix"></pds-icon>
+  <pds-button class="pds-input__suffix" variant="unstyled" slot="suffix"><pds-icon name="remove-circle"></pds-icon></pds-button>
+</pds-input>
+
```
pixelflips, line 313: Should we have an example that uses `pds-text` or some type of text as well? I know sometimes we have inputs with like `https://` at the beginning or `mykajabi.com` at the end

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -97,13 +153,62 @@
...
 .pds-input__error-message {
   align-items: center;
-  color: var(--pine-color-text-message-danger);
+  color: var(--pds-input-error-color);
   display: flex;
   gap: var(--pine-dimension-2xs);
 }
+
+.pds-input__prefix,
+.pds-input__suffix {
```
pixelflips: Should the prefix/suffix element styles change when disabled, readonly, invalid, etc?

libs/core/src/components/pds-input/pds-input.tsx
```diff
@@ -137,6 +152,89 @@ export class PdsInput {
...
+  }
+
+  componentWillLoad() {
+    this.inheritedAttributes = {
+      ...inheritAriaAttributes(this.el)
+    };
+    this.hasPrefix = this.el.querySelector('[slot="prefix"]') !== null;
+    this.hasSuffix = this.el.querySelector('[slot="suffix"]') !== null;
+    this.hasPrepend = this.el.querySelector('[slot="prepend"]') !== null;
+    this.hasAppend = this.el.querySelector('[slot="append"]') !== null;
```
ju-Skinner, line 230: Can you elaborate more on these (e.g, the difference between prefix vs prepend and suffix and append)?

libs/core/src/components/pds-input/readme.md
```diff
@@ -49,6 +49,16 @@ Type: `Promise<void>`
...
 
 
+## Shadow Parts
+
+| Part        | Description |
+| ----------- | ----------- |
+| `"append"`  |             |
+| `"prefix"`  |             |
+| `"prepend"` |             |
+| `"suffix"`  |             |
```
ju-Skinner, line 69: These need descriptions, which can be added at the top of the `input.tsx` file using TSDoc syntax.

libs/core/src/components/pds-input/pds-input.tsx
```diff
@@ -137,6 +152,89 @@ export class PdsInput {
    */
   @State() hasFocus = false;
 
+  private updateAddonWidths() {
+    requestAnimationFrame(() => {
```
ju-Skinner, line 162: Do we need this? Is this a request from design?

ju-Skinner (changes requested):
I requested changes due to the missing description for the new slots.  Depending on the answers to the questions, additional changes may be needed.

libs/core/src/components/pds-input/docs/pds-input.mdx
```diff
@@ -113,8 +113,82 @@ When `invalid` is true, the input displays the error message and provides a visu
...
+## References
 
-[MDN Web Docs: HTML autocomplete attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)
+For more information about using inputs, refer to:
+- [WAI-ARIA Input Practices](https://www.w3.org/WAI/ARIA/apg/patterns/textbox/)
```
pixelflips: I'm getting a 404 for this link.

libs/core/src/components/pds-input/pds-input.tsx
QuintonJason: Updated the docs to call out that prefix/suffix items appear visually within the input, while append/prepend appear next to it

libs/core/src/components/pds-button/pds-button.scss
QuintonJason: This was added so that buttons that exist within the slot of an input can receive appropriate styling. Outside of creating these in the host, I don't have a way to update the `border-radius-*` or other props of that button

libs/core/src/components/pds-input/pds-input.tsx
QuintonJason: Yes, the input value needs to account for the size of the prefix of suffix and apply padding to not have an overflow of text situation

Screenshot 2025-04-14 at 11 39 51 AM

libs/core/src/components/pds-input/pds-input.scss
QuintonJason: The need for the `border-radius` update here caused the need in `button.scss` and `select.scss`

libs/core/src/components/pds-input/pds-input.tsx
ju-Skinner: Sorry, when I said "this", I was referring to `requestAnimationFrame`

QuintonJason: Gotcha. There was an issue where the component would load, but the width would immediately be 0. I used that method instead of adding a setTimeout. Is there a preferred way to accomplish this outside of my proposed solution?

libs/core/src/components/pds-input/docs/pds-input.mdx
```diff
@@ -113,8 +113,82 @@ When `invalid` is true, the input displays the error message and provides a visu
...
+  <pds-select class="pds-input__append" slot="append">
+    <option value="1">USD</option>
+    <option value="2">EUR</option>
+    <option value="3">GBP</option>
+  </pds-select>
+</pds-input>
+
+**Unstyled button suffix**
+<pds-input component-id="pds-input-prefix-suffix-example-2" label="Name" type="email" value="Frank Dux">
+  <pds-button class="pds-input__suffix" variant="unstyled" slot="suffix"><pds-icon name="remove-circle"></pds-icon></pds-button>
```
pixelflips: Would the author be expected to apply the `__suffix`/`__append`/`__prepend` classes, or is this something we could do by detecting the slot?

libs/core/src/components/pds-input/docs/pds-input.mdx
```diff
@@ -113,8 +113,82 @@ When `invalid` is true, the input displays the error message and provides a visu
...
+<pds-input component-id="pds-input-prefix-suffix-example" label="Name" type="email" value="Frank Dux">
+  <pds-icon name="dollar" slot="prefix"></pds-icon>
+  <pds-select class="pds-input__append" slot="append">
+    <option value="1">USD</option>
+    <option value="2">EUR</option>
+    <option value="3">GBP</option>
+  </pds-select>
+</pds-input>
+
+**Unstyled button suffix**
```
pixelflips, line 269: How should we handle other button variants?

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -1,7 +1,63 @@
...
+  --pds-button-min-height: 36px;
+}
+
+/* stylelint-disable-next-line */
+:host([has-append]) ::slotted(pds-button[slot="append"]) {
+  --pds-button-background: var(--pds-input-addon-background);
+  --pds-button-border: var(--pine-border-width-none);
+  --pds-button-border-radius-end-start: var(--pine-dimension-none);
+  --pds-button-border-radius-start-start: var(--pine-dimension-none);
+  --pds-button-min-height: 36px;
```
pixelflips: Could all the min-height vars use `--pine-dimension-450`? They seem to be all the same

pixelflips (changes requested):
Off to a great start. I left a few comments, but if they don't apply, mark as resolved or let me know.

libs/core/src/components/pds-button/pds-button.scss
```diff
@@ -37,18 +44,23 @@
 .pds-button {
   --pds-loader-color: var(--color-text-default);
   align-items: center;
-  background-color: var(--color-background-default);
-  border: var(--pine-border);
+  background-color: var(--pds-button-background);
```
pixelflips: I think this might break the background colors applied to all the buttons. You may also need to update the variable naming in the variant styles as well. Think we should keep the previous naming? `--color-background-default` - that could be another solution, but the `--pds` prefix may make more sense.

anechol: Ditto on this.

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -12,81 +68,81 @@
...
+  background: var(--pds-input-background);
+  border: var(--pds-input-border-width) solid var(--pds-input-border-color);
+  border-radius: var(--pds-input-border-radius);
+  color: var(--pds-input-text-color);
+  flex: 1;
+  font: var(--pds-input-font);
+  letter-spacing: var(--pine-letter-spacing);
+  min-width: var(--pine-dimension-none);
+  padding: var(--pds-input-padding-y) var(--pds-input-padding-x);
+  transition: all 0.2s ease;
```
anechol: Bit of a performance nitpick here, but are we okay with transitioning all properties?

libs/core/src/components/pds-button/pds-button.scss
```diff
@@ -37,18 +44,23 @@
...
+  border: var(--pds-button-border);
+  border-radius: var(--pds-button-border-radius);
+  border-bottom-left-radius: var(--pds-button-border-radius-end-start, var(--pds-button-border-radius));
+  border-bottom-right-radius: var(--pds-button-border-radius-end-end, var(--pds-button-border-radius));
```
pixelflips: Should these use logical?

libs/core/src/components/pds-button/pds-button.scss
```diff
@@ -37,18 +44,23 @@
...
   border-color: var(--color-border-default);
-  border-radius: var(--pine-border-radius-full);
+  border-top-left-radius: var(--pds-button-border-radius-start-start, var(--pds-button-border-radius));
+  border-top-right-radius: var(--pds-button-border-radius-start-end, var(--pds-button-border-radius));
```
pixelflips: Should these use logical?

libs/core/src/components/pds-input/pds-input.tsx
anechol: It sounds like there's potentially an issue with the components not being defined yet. Could we possibly leverage the `.whenDefined()` method? I'm not entirely sure what the best practice would be in this case.
https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/whenDefined

anechol (changes requested):
Nice work on this so far! There's a couple of things below to consider, but outside of that, a question: is it possible for us to include the slots in the Playground?

libs/core/src/components/pds-button/pds-button.scss
QuintonJason: added fallback to resolve

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -1,5 +1,123 @@
 :host {
+  --pds-input-background: var(--pine-color-background-container);
+  --pds-input-border-color: var(--pine-color-border);
+  --pds-input-border-color-hover: var(--pine-color-border-hover);
+  --pds-input-border-color-disabled: var(--pine-color-border-disabled);
+
+  --pds-input-border-radius: var(--pine-dimension-125);
+  --pds-input-border-width: 1px;
```
pixelflips: This could use `--pine-border-width`

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -1,5 +1,123 @@
...
+  --pds-input-icon-color: var(--pine-color-text-secondary);
+  --pds-input-disabled-background: var(--pine-color-background-container-disabled);
+  --pds-input-disabled-text-color: var(--pine-color-text-disabled);
+  --pds-input-error-color: var(--pine-color-text-message-danger);
+  --pds-input-error-background: var(--pine-input-color-background-danger);
+  --pds-input-error-border: var(--pine-color-border-danger);
+  --pds-input-error-border-hover: var(--pine-color-border-danger-hover);
+  --pds-input-addon-background: var(--pine-color-background-subtle);
+  --pds-input-addon-color: var(--pine-color-text-secondary);
+  --pds-input-field-min-height: 38px;
```
pixelflips: Think we should do a calc here with two `dimension` tokens? 36 + 2? Not a show stopper.

pixelflips (approved):
A couple of additional comments, but nothing blocking. Great work!

pixelflips (approved):

PR #430, opened 2025-04-16
feature: adds flexible button slots

anechol:
# Description

Adds flexible slots for Button icons. The current `icon` prop is still usable, and the `disclosure` variant will be respected.

This comes with the expectation that the `icon` prop will be deprecated in future work.

[DSS-1356](https://kajabi.atlassian.net/browse/DSS-1356)

## Type of change

Screenshot 2025-04-29 at 4 12 15 PM

# How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-1356]: https://kajabi.atlassian.net/browse/DSS-1356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/core/src/components.d.ts
```diff
@@ -1840,6 +1845,11 @@ declare namespace LocalJSX {
           * Specifies where to open the linked document when href is provided. Only applies when href is set.
          */
         "target"?: '_blank' | '_self' | '_parent' | '_top';
+        /**
+          * Displays a trailing icon in the button. Does not render when iconOnly is true.
+          * @defaultValue null
+         */
+        "trailingIcon"?: string;
```
pixelflips: I didn't dig into it, but could we possibly use something like `flex-direction: row-reverse` since the button uses a flex layout? 

Separate icon and icon-trailing props feel odd to me vs. possibly refactoring to have just start and end slot if we can't use flex styling.

anechol: I had the same thought too, but after speaking with Quinton yesterday, it sounds like the ask is to actually add another icon affordance on the component. Because yes, if the icon needed to flip, there would be nothing to do here because of flex flipping automatically.

ju-Skinner: @QuintonJason, could you elaborate more on the conversation between you and Ashley? This will allow us to understand how to decide if this prop should be added. Thanks

libs/core/src/components/pds-button/pds-button.tsx
```diff
@@ -133,6 +136,26 @@ export class PdsButton {
...
   }
 
+  private renderStartContent() {
+    if (this.icon && this.variant !== 'disclosure') {
+      return (
+        <pds-icon class={this.loading ? 'pds-button__icon--hidden' : ''} name={this.icon} part="icon" aria-hidden="true"></pds-icon>
+      );
+    } else {
+      return <slot name="start" />;
+    }
```
pixelflips: I understand it may be a breaking change, but if we are moving to slots, we should deprecate the icon prop. Maybe we start by removing it from the docs (OR update any examples with icons to use the slot), and then once we've updated KP usage, we could do away with the prop from the code.

Just a thought, but in my opinion, we should only have one or the other to reduce confusion. Definitely check with the others.

pixelflips (commented):
Left a comment on icon prop vs slots, but this is looking good. Just let me know what the group decides on and I'll approve depending.

libs/core/src/components/pds-button/pds-button.tsx
anechol: Yep, deprecation will come in a future update. I've updated the description to reflect that and also updated the docs examples to use slots.

libs/core/src/components/pds-button/docs/pds-button.mdx
```diff
@@ -157,7 +157,7 @@ Full width buttons are intended to be used when the button needs to take up the
   <pds-button variant="primary" full-width="true">Full Width</pds-button>
 </DocCanvas>
 
-## Loading
+### Loading
 
 The loading prop allows a `pds-button` to display a loading state.
```
pixelflips, line 181: I think we have to check the loading when using slots. The icons are still visible.

anechol: Thanks for catching that, should be good now.

libs/core/src/components/pds-button/docs/pds-button.mdx
```diff
@@ -184,26 +208,27 @@ While the text is visually hidden, it is still remains accessible.
...
-    <PdsButton variant="primary" iconOnly="true" icon="favorite">Icon Only</PdsButton>
-    <PdsButton variant="secondary" iconOnly="true" icon="favorite">Icon Only</PdsButton>
-    <PdsButton variant="accent" iconOnly="true" icon="favorite">Icon Only</PdsButton>
-    <PdsButton variant="destructive" iconOnly="true" icon="favorite">Icon Only</PdsButton>
+    <PdsButton variant="primary" iconOnly="true"><pds-icon slot="start" aria-hidden="true" name="favorite" part="icon"></pds-icon>Icon Only
+    </PdsButton>
+    <PdsButton variant="secondary" iconOnly="true"><pds-icon slot="start" aria-hidden="true" name="favorite" part="icon"></pds-icon>Icon Only
+    </PdsButton>
+    <PdsButton variant="accent" iconOnly="true"><pds-icon slot="start" aria-hidden="true" name="favorite" part="icon"></pds-icon>Icon Only</PdsButton>
+    <PdsButton variant="destructive" iconOnly="true"><pds-icon slot="start" aria-hidden="true" name="favorite" part="icon"></pds-icon>Icon Only</PdsButton>
```
ju-Skinner: These should use the React version of `PdsIcon` and their respective prop names `ariaHidden` vs `aria-hidden`.

Was `part="icon"` added intentionally?

anechol: Removed the unnecessary `part` attribute, but kept `aria-hidden` as that is supported in React.

libs/core/src/components/pds-button/docs/pds-button.mdx
```diff
@@ -173,7 +173,31 @@ The loading prop allows a `pds-button` to display a loading state.
   <pds-button variant="primary" loading="true">Loading</pds-button>
```
pixelflips, line 192: The last thing I noticed is that the button size/width changes when loading is true. It seems to only happen when the icon is present. When text only, it holds its size/width when loading.

anechol: Ahhh that's annoying. I have the slots set to not render when they aren't being used otherwise it's just empty markup sitting around 🤔

libs/core/src/components/pds-button/pds-button.tsx
```diff
@@ -43,8 +46,9 @@ export class PdsButton {
   @Prop() href?: string;
 
   /**
-   * Displays an icon in the button.
+   * Displays a leading icon in the button. DEPRECATED.
    * @defaultValue null
+   * @deprecated Use `start` slot instead.
```
ju-Skinner, line 51: While I like the idea of using this, it will only notify that there is a plan to remove it in a future release. Do we have a timeline for when we will remove it?

libs/core/src/components/pds-button/pds-button.tsx
```diff
@@ -185,9 +223,7 @@ export class PdsButton {
           </span>
         )}
 
-        {this.variant === 'disclosure' &&
-          <pds-icon class={this.loading ? 'pds-button__icon--hidden' : ''} icon={caretDown} part="caret" aria-hidden="true"></pds-icon>
-        }
+        {this.renderEndContent()}
```
pixelflips, line 226: When there is an icon placed using the slot, when I set the button to loading ,the icon does not disappear. See screenshot.

Screenshot 2025-04-23 at 2 42 06 PM

pixelflips (commented):
Still seeing a slight issue when loading and the icon slot visibility.

libs/core/src/components/pds-button/docs/pds-button.mdx
```diff
@@ -184,26 +208,27 @@ While the text is visually hidden, it is still remains accessible.
...
-    <PdsButton variant="primary" iconOnly="true" icon="favorite">Icon Only</PdsButton>
-    <PdsButton variant="secondary" iconOnly="true" icon="favorite">Icon Only</PdsButton>
-    <PdsButton variant="accent" iconOnly="true" icon="favorite">Icon Only</PdsButton>
-    <PdsButton variant="destructive" iconOnly="true" icon="favorite">Icon Only</PdsButton>
+    <PdsButton variant="primary" iconOnly="true"><pds-icon slot="start" aria-hidden="true" name="favorite"></pds-icon>Icon Only
+    </PdsButton>
+    <PdsButton variant="secondary" iconOnly="true"><pds-icon slot="start" aria-hidden="true" name="favorite"></pds-icon>Icon Only
+    </PdsButton>
+    <PdsButton variant="accent" iconOnly="true"><pds-icon slot="start" aria-hidden="true" name="favorite"></pds-icon>Icon Only</PdsButton>
+    <PdsButton variant="destructive" iconOnly="true"><pds-icon slot="start" aria-hidden="true" name="favorite"></pds-icon>Icon Only</PdsButton>
```
ju-Skinner: Can you change the web component version of `pds-icon` to use the React version `PdsIcon`?

ju-Skinner (changes requested):

libs/core/src/components/pds-button/docs/pds-button.mdx
anechol: Done

libs/core/src/components/pds-button/pds-button.tsx
anechol: Spoke to Quinton about this, and the idea here is that we didn't want to introduce a breaking change by removing the prop immediately in one fell swoop, but we can create a ticket to start work on this for the next release.

pixelflips (approved):
The loading/icon issue seems to be resolved! Nice work here! LGTM! 👍🏼

ju-Skinner (approved):

QuintonJason (approved):

PR #449, opened 2025-05-05
fix(chip): updates a11y docs

anechol:
# Description

Adds info around Dropdown variant in Accessibility section.

[DSS-1384](https://kajabi.atlassian.net/browse/DSS-1384)

## Type of change

- [x] Documentation (adds, updates, or removes component documentation)

## How Has This Been Tested?

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

[DSS-1384]: https://kajabi.atlassian.net/browse/DSS-1384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/components.d.ts
+++ b/libs/core/src/components.d.ts
@@ -1461,6 +1461,7 @@ declare global {
     };
     interface HTMLPdsChipElementEventMap {
         "pdsTagCloseClick": any;
+        "pdsClick": any;
     }
     interface HTMLPdsChipElement extends Components.PdsChip, HTMLStencilElement {
         addEventListener<K extends keyof HTMLPdsChipElementEventMap>(type: K, listener: (this: HTMLPdsChipElement, ev: PdsChipCustomEvent<HTMLPdsChipElementEventMap[K]>) => any, options?: boolean | AddEventListenerOptions): void;
@@ -2310,6 +2311,10 @@ declare namespace LocalJSX {
           * @defaultValue false
          */
         "large"?: boolean;
+        /**
+          * Event emitted when the dropdown button is clicked on a dropdown variant chip.
+         */
+        "onPdsClick"?: (event: PdsChipCustomEvent<any>) => void;
         /**
           * Event emitted when the close button is clicked on a tag variant chip.
          */
--- a/libs/core/src/components.d.ts
+++ b/libs/core/src/components.d.ts
@@ -1461,7 +1461,6 @@ declare global {
     };
     interface HTMLPdsChipElementEventMap {
         "pdsTagCloseClick": any;
-        "pdsClick": any;
     }
     interface HTMLPdsChipElement extends Components.PdsChip, HTMLStencilElement {
         addEventListener<K extends keyof HTMLPdsChipElementEventMap>(type: K, listener: (this: HTMLPdsChipElement, ev: PdsChipCustomEvent<HTMLPdsChipElementEventMap[K]>) => any, options?: boolean | AddEventListenerOptions): void;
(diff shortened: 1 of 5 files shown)


QuintonJason (approved):
Nice

libs/core/src/components.d.ts
```diff
@@ -1232,6 +1232,7 @@ declare global {
     };
     interface HTMLPdsChipElementEventMap {
         "pdsTagCloseClick": any;
+        "pdsClick": any;
```
pixelflips: I’m not entirely sure since we don’t have a dropdown component implemented yet, but given that this is meant to act as a dropdown trigger, should the dropdown itself handle the click event? Just wondering if there’s a risk of multiple conflicting events being fired.

Just a thought.

anechol: Fair point. I suppose it really depends on how we're building the Dropdown and whether we'd like to still consider the `panel` as a separate component, e.g. something we can use in multiple components like Dropdown, Popover, Modal, etc.

anechol: I'm going to bring this up as a topic of discussion once we have a meeting surrounding pattern work soon.

libs/core/src/components/pds-chip/pds-chip.tsx
```diff
@@ -46,10 +46,19 @@ export class PdsChip {
    */
   @Event() pdsTagCloseClick: EventEmitter;
```
ju-Skinner: I know this has been there, but I'm not sure how this would be used.

ju-Skinner:
I don't think we would need any of the click events here or have the "dropdown" variant.  Take a look at the example below, it achieves the same results.

```
  <pds-dropdown-menu>
    <pds-chip slot="trigger" sentiment="success">My trigger <pds-icon name="caret-down"></pds-icon></pds-chip>
    <pds-dropdown-menu-item>Item 1</pds-dropdown-menu-item>
  </pds-dropdown-menu>
```

image

ju-Skinner (changes requested):
see my comment with an example using dropdown-menu

anechol:
> I don't think we would need any of the click events here or have the "dropdown" variant. Take a look at the example below, it achieves the same results.
> 
> ```
>   <pds-dropdown-menu>
>     <pds-chip slot="trigger" sentiment="success">My trigger <pds-icon name="caret-down"></pds-icon></pds-chip>
>     <pds-dropdown-menu-item>Item 1</pds-dropdown-menu-item>
>   </pds-dropdown-menu>
> ```
> 
> image

@ju-Skinner Agree with you on the event for the Chip now that we have the Dropdown built so I'll get that removed. As far as the Dropdown variant, if we remove that and use the regular Chip, then we'd have to wrap it inside of a button. With the example you provided, the Chip receives no keyboard focus and isn't called out to a screenreader. The Dropdown variant is a button under the hood.

ju-Skinner:
~~It should as it's wrapped in a button.~~

Never mind you're right. It's not I was thinking of the items.

libs/core/src/components/pds-chip/pds-chip.tsx
anechol: Let's make this a future conversation now that the Dropdown is in place.

QuintonJason (approved):

pixelflips (approved):

PR #451, opened 2025-05-08
refactor(pds-tooltip): better positioning when contained in overflow elements

pixelflips:
# Description

This PR refactors the `pds-tooltip` component to address several issues and improve its robustness, primarily by rendering tooltip content within a portal appended to the `document.body`.

**Fixes:** The primary issue fixed is the tooltip content being clipped by parent containers with `overflow: hidden`. This also addresses subsequent challenges related to styling and positioning the portalized tooltip.

**Key Changes:**

1.  **Portal Implementation:**
    *   Tooltip content is now rendered in a portal to prevent clipping by parent elements with `overflow: hidden`.
    *   The portal is dynamically created and removed.
    *   Positioning is handled using `position: fixed` and calculated relative to the trigger element's viewport dimensions via `getBoundingClientRect()`. The `positionTooltip` utility was updated accordingly.

2.  **Styling Strategy:**
    *   To address difficulties in styling portal content (which is outside the component's original scope), the component has been updated to use Light DOM by setting `shadow: false` and `scoped: false` in the `@Component` decorator.
    *   This allows styles defined in `pds-tooltip.scss` to apply to both the host/trigger and the portal content directly, eliminating the need for injecting styles into the portal. Selectors in `pds-tooltip.scss` were updated (e.g., `:host` to `.pds-tooltip`).

3.  **Slotted Content (`<slot name="content">`):**
    *   Logic has been implemented to support HTML content provided via `<slot name="content">`.
    *   The component queries for nodes assigned to this slot and clones them into the portal. It prioritizes slotted content over the `content` prop.
    *   A hidden wrapper around `<slot name="content">` in the component's `render()` method ensures the slot is available for querying.

4.  **Positioning Refinements:**
    *   The logic for determining the positioning anchor (`determinePositioningAnchor()`) has been improved to attempt using the actual slotted trigger element for more precise alignment, especially with elements like buttons.
    *   A `ResizeObserver` is used on the portal's content `div` to automatically call `repositionPortal()` if the content size changes, ensuring the tooltip remains correctly positioned.

5.  **Controlled vs. Uncontrolled State (`opened` prop):**
    *   The component now better distinguishes between being opened via user interaction (hover/focus) versus being programmatically opened via the `opened` prop.
    *   An internal state (`_isInteractiveOpen`) tracks this.
    *   When in a controlled open state (via the `opened` prop), tooltips will not close on scroll or SPA navigation events but will instead reposition. Interactively opened tooltips will still close on these events.

Fixes https://kajabi.atlassian.net/browse/DSS-1280

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-tooltip/docs/pds-tooltip.mdx
+++ b/libs/core/src/components/pds-tooltip/docs/pds-tooltip.mdx
@@ -100,7 +100,7 @@ enable the `htmlContent` property and provide your element in the `content` slot
 Using the placement prop you can adjust where your tooltip will be displayed.
 
 
-### Arrow
+### No Arrow
 By default the arrow is visible but it can be hidden by disabling it when `has-arrow` is `false`.
 
 <DocCanvas client:only
--- a/libs/core/src/components/pds-tooltip/docs/pds-tooltip.mdx
+++ b/libs/core/src/components/pds-tooltip/docs/pds-tooltip.mdx
@@ -69,7 +69,7 @@ enable the `htmlContent` property and provide your element in the `content` slot
 <DocCanvas client:only
   mdxSource={{
     react: `
-      <PdsTooltip hasArrow="true" placement="right">
+      <PdsTooltip htmlContent="true" hasArrow="true" placement="right">
         text
         <div slot="content">
           <p>this is a sentence in a tooltip.</p>
@@ -78,7 +78,7 @@ enable the `htmlContent` property and provide your element in the `content` slot
       </PdsTooltip>
     `,
     webComponent: `
-      <pds-tooltip has-arrow="true" placement="right">
+      <pds-tooltip html-content="true" has-arrow="true" placement="right">
         text
         <div slot="content">
           <p>this is a sentence in a tooltip.</p>
@@ -87,9 +87,9 @@ enable the `htmlContent` property and provide your element in the `content` slot
       </pds-tooltip>
     `
 }}>
-  <pds-tooltip has-arrow="true" placement="right">
+  <pds-tooltip html-content="true" has-arrow="true" placement="right">
     text
-    <div class="sb-unstyled" slot="content">
+    <div slot="content">
       <p>this is a sentence in a tooltip.</p>
       <p>this is a sentence in a tooltip.</p>
     </div>
--- a/libs/core/src/components/pds-tooltip/pds-tooltip.tsx
+++ b/libs/core/src/components/pds-tooltip/pds-tooltip.tsx
@@ -119,6 +119,7 @@ export class PdsTooltip {
 
     return () => {
       window.removeEventListener('pageshow', this.handlePageShow);
+
       if (this.slotMutationObserver !== null) {
         this.slotMutationObserver.disconnect();
       }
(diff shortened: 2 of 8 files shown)


libs/core/src/components/pds-tooltip/pds-tooltip.tsx
```diff
@@ -11,10 +9,22 @@ import {
 @Component({
   tag: 'pds-tooltip',
   styleUrls: ['pds-tooltip.scss'],
-  shadow: true,
+  shadow: false,
```
pixelflips, line 12: Removing the shadow so the portal approach can be supported. This allows for inserting the tooltip before the closing body tag.

libs/core/src/components/pds-tooltip/pds-tooltip.tsx
```diff
@@ -133,54 +146,220 @@ export class PdsTooltip {
...
+            this.contentDiv.style.left = `${currentOverlayLeft + adjustmentX}px`;
+          }
+        }
+      }
+    }
+  }
+
+  private createPortal() {
+    if (this.portalEl !== null) return;
+    this.portalEl = document.createElement('div');
```
ju-Skinner, line 271: Add line break before and/or after the `if`'s

ju-Skinner (changes requested):
See comments regarding the `if` statements

libs/core/src/components/pds-tooltip/docs/pds-tooltip.mdx
```diff
@@ -69,7 +69,7 @@ enable the `htmlContent` property and provide your element in the `content` slot
 <DocCanvas client:only
   mdxSource={{
     react: `
-      <PdsTooltip hasArrow="true" placement="right">
+      <PdsTooltip htmlContent="true" hasArrow="true" placement="right">
```
pixelflips, line 72: This is not a new prop/attr, it was missing from the doc examples.

libs/core/src/components/pds-tooltip/stories/pds-tooltip.stories.js
```diff
@@ -15,10 +15,10 @@ export default {
 }
 
 const BaseTemplate = (args) => html`
-  <pds-tooltip content=${args.content} max-width=${args.maxWidth} has-arrow=${args.hasArrow} placement=${args.placement}>${args.slot}</pds-tooltip>`;
+  <pds-tooltip content=${args.content} max-width=${args.maxWidth} has-arrow=${args.hasArrow} placement=${args.placement} opened=${args.opened}>${args.slot}</pds-tooltip>`;
```
pixelflips, line 18: Updates in this file were made to improve behavior when interacting with the Storybook controls.

anechol (approved):
LGTM!

libs/core/src/components/pds-tooltip/pds-tooltip.tsx
pixelflips: Thanks for the help and clarification on these.

ju-Skinner (approved):

QuintonJason (approved):

PR #458, opened 2025-05-22
feat(link): adds color prop

anechol:
# Description
Adds `color` prop for the Link

[DSS-1434](https://kajabi.atlassian.net/browse/DSS-1434)

## Type of change

Screenshot 2025-05-22 at 4 21 47 PM

# How Has This Been Tested?

- [x] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

[DSS-1434]: https://kajabi.atlassian.net/browse/DSS-1434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libs/core/src/components/pds-link/pds-link.tsx
+++ b/libs/core/src/components/pds-link/pds-link.tsx
@@ -13,6 +13,11 @@ import { launch } from '@pine-ds/icons/icons';
 })
 export class PdsLink {
   /**
+   * Sets the link color.
+   */
+  @Prop() color?: 'secondary' | 'accent' | 'danger';
+
+  /**
    * A unique identifier used for the underlying component `id` attribute.
    */
   @Prop() componentId: string;
--- a/libs/core/src/components/pds-link/pds-link.tsx
+++ b/libs/core/src/components/pds-link/pds-link.tsx
@@ -1,4 +1,5 @@
 import { Component, h, Prop } from '@stencil/core';
+import { setColor } from '../../utils/utils';
 
 import { launch } from '@pine-ds/icons/icons';
 
@@ -15,7 +16,7 @@ export class PdsLink {
   /**
    * Sets the link color.
    */
-  @Prop() color?: 'secondary' | 'accent' | 'danger';
+  @Prop() color?: string;
 
   /**
    * A unique identifier used for the underlying component `id` attribute.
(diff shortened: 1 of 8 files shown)


libs/core/src/components/pds-link/pds-link.tsx
```diff
@@ -12,6 +12,11 @@ import { launch } from '@pine-ds/icons/icons';
   shadow: true,
 })
 export class PdsLink {
+  /**
+   * Sets the link color.
+   */
+  @Prop() color?: 'secondary' | 'accent' | 'danger';
```
ju-Skinner: Do we only want to restrict to these? I know in other components, we kind of opened it up.

ju-Skinner (approved):
Left a question for thought, but it DOES NOT block my approval.

libs/core/src/components/pds-link/pds-link.tsx
anechol: I thought about that, but I don't really see there being a need for someone to have a color outside of this for links. My thought was if we do, we can just refactor later.

pixelflips: I would agree with @ju-Skinner here. Or should we at least match the `pds-text` colors, maybe? I could imagine links being inserted into text content.

We also have others, like mercury, where it would be great to convert to `pds-link`.

Screenshot 2025-05-28 at 2 22 48 PM

anechol: Ahhh thanks for this example. I'll make the change then.

anechol: This has been updated.

libs/core/src/components/pds-link/stories/pds-link.stories.js
```diff
@@ -4,12 +4,29 @@ import { extractArgTypes } from '@pxtrn/storybook-addon-docs-stencil';
 export default {
   argTypes: extractArgTypes('pds-link'),
   args: { href: '#some-anchor' },
+  argTypes: {
+    color: {
+      control: 'radio',
+      options: ['secondary', 'accent', 'danger'],
+    },
```
pixelflips: This doesn't allow custom options to be used in the Storybook controls. Do you think it would be better to document these in the Docs and allow strings here instead? The combination of the two feels confusing.

anechol: Fixed this.

pixelflips (approved):
LGTM!

PR #593, opened 2025-11-20
fix(release): update Slack notification workflow to use environment var

pixelflips:
# Description

Fixes the Slack release notification to display the correct version number. Previously, the notification was showing `**` (empty) for the version because it attempted to read from the root `package.json`, which doesn't contain a version field in this monorepo setup.

**Changes:**
- Extract version from the git tag created by Nx release using `git describe --tags --abbrev=0`
- Simplify environment variable setup by removing unnecessary `package.json` parsing
- Construct release URL directly from the git tag

Next release should show the actual version instead of it missing in the most recent release.

Screenshot 2025-11-20 at 8 28 55 AM

Fixes [DSS-14](https://linear.app/kajabi/issue/DSS-14/fix-empty-version-numbers-in-slack-release-notifications)

## Type of change

# How Has This Been Tested?

- [x] Will be verified on the next Pine release. The fix uses the git tag that Nx release creates, which is a reliable source for the version number.

**Test Configuration**:

- The previous release showed empty versions (`**`) in the Slack notification
- This fix ensures the version is extracted from the git tag instead

# Checklist:

### DIFF
--- a/.github/workflows/schedule-release.yml
+++ b/.github/workflows/schedule-release.yml
@@ -246,6 +246,10 @@ jobs:
         run: git pull
         shell: bash
 
+      - name: Fetch latest tags
+        run: git fetch --tags
+        shell: bash
+
       - name: Setup ENV Vars
         run: |
           # Get the latest git tag (which is the version just released)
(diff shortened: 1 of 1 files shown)


pixelflips:
@coderabbitai review

QuintonJason (approved):

PR #625, opened 2026-01-08
fix(pds-modal) Escape key dismisses modal even when backdropDismiss is false

pixelflips:
# Description

Fixes an issue where the Escape key could still dismiss a `pds-modal` even when `backdropDismiss` was set to `false`.

**Root cause:** The native `<dialog>` element has built-in Escape-to-close behavior. The previous implementation only called `e.preventDefault()` when `backdropDismiss` was `true`, allowing the browser's native behavior to bypass our logic when set to `false`.

**Solution:** Always call `e.preventDefault()` when Escape is pressed on an open modal, then conditionally call `hideModal()` based on the `backdropDismiss` prop.

Fixes DSS-63

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

### DIFF
--- a/libs/core/src/components.d.ts
+++ b/libs/core/src/components.d.ts
@@ -32,7 +32,7 @@ export namespace Components {
      */
     interface MockPdsModal {
         /**
-          * Whether the modal can be dismissed by clicking the backdrop
+          * Whether the modal can be dismissed by clicking the backdrop or pressing Escape
           * @default true
          */
         "backdropDismiss": boolean;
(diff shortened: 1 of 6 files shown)


QuintonJason (approved):
Great fix

pixelflips:
@coderabbitai review

PR #647, opened 2026-01-27
fix: prevent double-initialization crashes with CDN and React

pixelflips:
## Summary

When both CDN and `@pine-ds/react` are loaded on the same page, components can be initialized twice, causing crashes. This PR makes components resilient to double-initialization.

**Changes:**
- Add `!this.internals` guard before calling `attachInternals()` in 7 form-associated components
- Rename `style()` to `getSvgStyle()` in `pds-loader` to avoid conflict with `HTMLElement.style`

Fixes DSS-94

## Type of change

# How Has This Been Tested?

- [x] tested manually

Tested locally in kajabi-products with CDN + @pine-ds/react both loaded. Errors no longer appear and components render correctly.

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-checkbox/pds-checkbox.tsx
+++ b/libs/core/src/components/pds-checkbox/pds-checkbox.tsx
@@ -127,7 +127,8 @@ export class PdsCheckbox {
   }
 
   connectedCallback() {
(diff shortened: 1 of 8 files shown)


QuintonJason (approved):
Visually see this work on a dev release

PR #667, opened 2026-02-09
fix(forms) redefine type property in forms

pixelflips:
## Summary                                                                                                                                                                   
  
  Fixes a `TypeError: Cannot redefine property: type` thrown when form-associated components (`pds-select`, `pds-input`, `pds-textarea`, `pds-switch`, `pds-radio`,            
  `pds-checkbox`) are moved in the DOM by a portal (e.g., `pds-popover`).
                                                                                                                                                                               
  ## Problem                                                                                                                                                                   

  The `exposeTypeProperty` utility in `form.ts` calls `Object.defineProperty(el, 'type', { configurable: false })` during `connectedCallback`. When a component is portaled,
  the element disconnects and reconnects, firing `connectedCallback` a second time. The second `defineProperty` call throws because the property was already defined as
  non-configurable.

  ## Fix

  Added an early return guard in `exposeTypeProperty` that checks if the `type` property already exists on the element instance before attempting to define it. This makes the
  function idempotent without changing any existing behavior.

  ## Affected Components

  - `pds-select`
  - `pds-textarea`
  - `pds-switch`
  - `pds-radio`
  - `pds-checkbox`

  ## Test Plan

  - [x] Added unit tests for `exposeTypeProperty` covering static values, getter functions, and duplicate call safety
  - [ ] Verify `pds-select` inside `pds-popover` no longer throws on open
  - [ ] Verify all form components continue to expose `type` correctly

  ## Linear Issue

  Fixes DSS-146

### DIFF
--- a/libs/core/src/utils/form.ts
+++ b/libs/core/src/utils/form.ts
@@ -37,6 +37,8 @@ export const isRequired = (target, component) => {
  * @param type - The type value (string literal) or a getter function that returns the type
  */
 export function exposeTypeProperty(element: Element, type: string | (() => string)) {
+  if (Object.getOwnPropertyDescriptor(element, 'type')) return;
+
   Object.defineProperty(element, 'type', {
     get: typeof type === 'function' ? type : () => type,
     enumerable: true,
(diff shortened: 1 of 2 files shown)


pixelflips:
@coderabbitai review

QuintonJason (approved):

PR #672, opened 2026-02-13
feat(truncation): add automatic tooltip on truncated text

QuintonJason:
# Description

Add automatic tooltip behavior when text is truncated in Pine components. Previously, truncated text (`text-overflow: ellipsis`) had no way to reveal the hidden content. Now, hovering or focusing a truncated element shows a tooltip with the full text — but only when the content is actually overflowing.

### What changed

- **New shared utility** (`truncation-tooltip.ts`) — handles overflow detection (`scrollWidth > clientWidth`), `ResizeObserver` for dynamic re-evaluation, hover/focus listeners, and a Floating UI-positioned tooltip portal on `document.body`. Styled with Pine design tokens to match `pds-tooltip`.
- **`pds-text`** — automatically shows tooltip on hover/focus when `truncate` is set and text overflows
- **`pds-table-cell`** — same behavior; also makes `max-width` configurable via `--pds-table-cell-truncate-max-width` CSS custom property (default: `100px`)
- **`pds-copytext`** — same behavior, uses `value` prop as tooltip content

No new props required — the tooltip is automatic when `truncate` is set. No breaking changes.

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] tested manually

12 new tests for the shared utility covering: tooltip creation on overflow, no tooltip when text fits, hover/focus interactions, hover+focus overlap, empty text edge case, and cleanup. Component spec files updated with integration tests for all three components.

**Test Configuration**:

- Pine versions: 3.17.0 (current main)
- OS: macOS

# Checklist:


QuintonJason:
@coderabbitai review this pr

pixelflips (commented):
Overall this is a well-implemented feature with solid architecture and good test coverage. A few accessibility and robustness items to address before merge — see inline comments.

libs/core/src/utils/truncation-tooltip.ts
```diff
@@ -0,0 +1,283 @@
...
+    .${TOOLTIP_CLASS}--visible .${TOOLTIP_CLASS}__content {
+      opacity: 1;
+      visibility: visible;
+    }
+  `;
+  document.head.appendChild(style);
+}
+
+/**
+ * Checks if an element's content is overflowing horizontally (i.e. text is truncated).
```
pixelflips, line 75: **Accessibility: Missing `aria-describedby` association**

The tooltip sets `role="tooltip"` and `aria-hidden="false"`, but there is no `aria-describedby` linking the host element to the tooltip. Without this, screen readers won't announce the tooltip content when the truncated element receives focus.

Suggestion:
- Generate a unique `id` for each tooltip portal (e.g., `pds-truncation-tooltip-${counter++}`)
- Set `aria-describedby="<tooltip-id>"` on `hostEl` when the tooltip is shown
- Remove `aria-describedby` from `hostEl` when the tooltip is hidden

This is a standard pattern per [WAI-ARIA tooltip](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/) and would make the feature properly accessible.

libs/core/src/utils/truncation-tooltip.ts
```diff
@@ -0,0 +1,283 @@
...
+        const padding = 5;
+
+        // Ensure tooltip stays within viewport bounds
+        const left = Math.max(padding, Math.min(rect.left, window.innerWidth - portalRect.width - padding));
+        const top = Math.max(padding, rect.top - portalRect.height - 8);
+
+        portalEl.style.left = `${left}px`;
+        portalEl.style.top = `${top}px`;
+      }
+    }
```
pixelflips, line 178: **Accessibility: Missing Escape key dismissal**

Per the [WAI-ARIA tooltip pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/), pressing `Escape` should dismiss the tooltip. This is not currently implemented.

Suggestion: Add a `keydown` listener (on `hostEl` or `document`) that calls `hideTooltip()` when `Escape` is pressed while the tooltip is visible. Something like:

```ts
function handleKeyDown(e: KeyboardEvent): void {
  if (e.key === 'Escape' && portalEl) {
    isHovering = false;
    isFocused = false;
    isHoveringPortal = false;
    cancelHideTimeout();
    removePortal();
  }
}
hostEl.addEventListener('keydown', handleKeyDown);
```

Don't forget to remove the listener in the cleanup function.

libs/core/src/components/pds-table/pds-table-cell/pds-table-cell.tsx
```diff
@@ -74,16 +81,44 @@ export class PdsTableCell {
...
+  handleTruncateChange(newValue: boolean) {
+    if (newValue) {
+      this.initTruncationTooltip();
+    } else {
+      this.destroyTruncationTooltip();
+    }
+  }
+
   /**
    * Determines if the table is currently scrolling.
```
pixelflips, line 100: **Robustness: `textContent` may include unintended child element text**

```ts
getTooltipText: () => this.hostElement.textContent || '',
```

If a table cell contains nested elements (badges, icons with labels, nested spans, etc.), `textContent` will concatenate all of their text content, which could result in garbled or misleading tooltip text.

Consider either:
1. Using `innerText` instead (respects CSS visibility/display, closer to what the user sees), or
2. Targeting a specific text node or slotted content rather than the entire host, or
3. Documenting this as a known limitation — that `truncate` works best on cells with plain text content

pixelflips (approved):
LGTM!

PR #674, opened 2026-02-13
feat(pds-multiselect): add searchPlaceholder, closePanelOnSelect, dis…

QuintonJason:
# Description

Adds Sage `SelectDropdown` parity features to `pds-multiselect` to unblock migration from Sage to Pine in kajabi-products.

Fixes [DSS-163](https://linear.app/kajabi/issue/DSS-163/featpds-multiselect-add-sage-selectdropdown-parity-features)

**New features:**
- **`searchPlaceholder` prop** — customize the search input placeholder text (defaults to "Find...")
- **`closePanelOnSelect` prop** — close the dropdown panel after each selection (defaults to `false`)
- **`pdsMultiselectDismiss` event** — fires when the panel is dismissed via Escape key or click-outside, enabling parent components to react (maps to Sage's `onEscapeHook`)
- **`clear()` method** — programmatic reset of all selections, search query, and form value (maps to Sage's `resetToken` pattern)

**Note:** Single-select use cases (Sage's `allowMultiselect={false}`) should use `pds-combobox` rather than `pds-multiselect`.

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] tested manually

16 new unit tests covering:
- `searchPlaceholder` prop (custom text, default fallback)
- `pdsMultiselectDismiss` event (Escape key fires event, no-op when closed)
- `closePanelOnSelect` prop (default keeps panel open, `true` closes panel)
- `clear()` method (clears values, resets trigger text, emits change event)

Storybook stories added for `CustomSearchPlaceholder` and `ClosePanelOnSelect`. MDX docs updated with examples for all four features including live interactive demos.

# Checklist:


QuintonJason:
@coderabbitai review this pr

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -451,6 +491,10 @@ export class PdsMultiselect {
     return this.options || this.internalOptions;
   }
 
+  private get shouldCloseOnSelect(): boolean {
+    return this.closePanelOnSelect;
+  }
```
pixelflips: **Nit: Unnecessary getter**

This getter is a pure passthrough — it adds indirection without any logic and is only used in one place (`toggleOption`). Consider using `this.closePanelOnSelect` directly and removing the getter.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -795,6 +839,7 @@ export class PdsMultiselect {
       const isOnHost = activeElement === this.el;
 
       if (!isInShadowRoot && !isOnHost) {
+        this.pdsMultiselectDismiss.emit();
```
pixelflips, line 843: **Potential race condition with `closePanelOnSelect`**

When `closePanelOnSelect` is `true`, selecting an option calls `closeDropdown()` and moves focus to the trigger via `triggerEl?.focus()`. Depending on browser timing, `handleContainerFocusOut` could fire *before* focus lands on the trigger, which would incorrectly emit `pdsMultiselectDismiss`.

The existing test ("does not emit dismiss event when panel closes via selection") validates this works in the test environment, but `setTimeout(…, 0)` timing can vary across browsers.

Consider adding a guard flag (e.g., `this.isClosingViaSelection = true` set before `closeDropdown()` in `toggleOption`, cleared in `closeDropdown()`) to explicitly suppress the dismiss event during selection-triggered closes. This would make the behavior deterministic rather than relying on focus timing.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -329,6 +353,21 @@ export class PdsMultiselect {
     this.triggerEl?.focus();
   }
 
+  /**
+   * Clears all selected values and resets the component.
+   */
+  @Method()
+  async clear() {
```
pixelflips, line 363: **Minor: `clear()` emits change even when already empty**

If `value` is already `[]`, calling `clear()` still emits a change event, which could cause unnecessary re-renders or side effects in consumers.

Consider guarding:
```ts
async clear() {
  const hadValues = this.value.length > 0 || this.searchQuery !== '';
  this.value = [];
  this.searchQuery = '';
  this.syncSelectedItems();
  this.updateFormValue();
  if (hadValues) {
    this.pdsMultiselectChange.emit({ values: [], items: [] });
  }
}
```

This is minor since `clear()` is an explicit programmatic action, but worth considering.

libs/core/src/components/pds-multiselect/stories/pds-multiselect.stories.js
```diff
@@ -266,6 +274,53 @@ export const HiddenSelectedItems = {
...
+    placeholder: 'Select...',
+    closePanelOnSelect: true,
+    value: [],
+  },
+  render: (args, { updateArgs } = {}) => html`
+    <pds-multiselect
+      component-id="${args.componentId}"
+      label="${args.label}"
+      placeholder="${args.placeholder}"
+      close-panel-on-select="${args.closePanelOnSelect}"
```
pixelflips: **Nit: Boolean attribute binding**

This renders as `close-panel-on-select="true"` (a string). It works because Stencil coerces it, but for consistency with how other boolean props might be bound, consider using the property binding:
```js
.closePanelOnSelect=${args.closePanelOnSelect}
```
or the boolean attribute syntax:
```js
?close-panel-on-select=${args.closePanelOnSelect}
```

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -54,6 +54,17 @@ export class PdsMultiselect {
    */
   @Prop() placeholder?: string = 'Select...';
 
+  /**
+   * Placeholder text for the search input inside the dropdown panel.
+   */
+  @Prop() searchPlaceholder: string = 'Find...';
```
pixelflips, line 61: **Very minor: Missing `@default` JSDoc tag**

The generated `components.d.ts` shows `@default 'Find...'` for this prop, but the source JSDoc doesn't include it. Adding `@default 'Find...'` would be consistent with props like `closePanelOnSelect` above.

pixelflips (commented):
## Code Review

Overall this is a well-executed PR. The features are well-scoped, documented, and tested. Solid work on the Sage parity additions. A few suggestions inline — the main concern is around potential focus-out timing with `closePanelOnSelect`. The rest are minor nits.

pixelflips (approved):
LGTM!

PR #678, opened 2026-02-17
fix(pds-button): fix empty icon slot wrapper spacing

pixelflips:
# Description

Fixes extra padding/spacing on buttons when no icon is slotted into the start/end slots.

PR #668 fixed a race condition by always rendering the `<slot name="start">` and `<slot name="end">` wrappers, using CSS `:has(::slotted(*))` to hide them when empty. However, `:has(::slotted(*))` has unreliable cross-browser support (particularly Safari). When the selector doesn't match, the empty `.pds-button__icon` spans remain `display: inline-flex`, and the parent's `gap` creates visible spacing where none should exist.

This replaces the CSS `:has()` approach with `slotchange` event handlers that toggle a `--empty` CSS class, following the existing pattern used in `pds-input`.

Fixes [DSS-167](https://linear.app/kajabi/issue/DSS-167/extra-paddingspacing-on-top-bar-menu-buttons)

### Screenshots
|  Before   |  After  |
|--------|--------|
|Screenshot 2026-02-17 at 9 49 41 AM|Screenshot 2026-02-17 at 9 49 04 AM|

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] tested manually

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-button/pds-button.scss
+++ b/libs/core/src/components/pds-button/pds-button.scss
@@ -218,10 +218,7 @@
 .pds-button__icon {
   display: inline-flex;
 
-  // Hide slot wrapper when no content is slotted to prevent empty flex gap space.
-  // Named slots (start/end) are always rendered in the shadow DOM so the browser
-  // can project light DOM content reliably regardless of render timing.
-  &:not(:has(::slotted(*))) {
+  &--empty {
     display: none;
   }
 }
--- a/libs/core/src/components/pds-button/pds-button.tsx
+++ b/libs/core/src/components/pds-button/pds-button.tsx
@@ -158,11 +158,11 @@ export class PdsButton {
 
 
   private handleStartSlotChange = (event: Event) => {
-    this.hasStartContent = (event.target as HTMLSlotElement).assignedNodes().length > 0;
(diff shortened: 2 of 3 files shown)


pixelflips:
@coderabbitai review

pixelflips:
@coderabbitai i addressed the nitpick comments, review again.

pixelflips:
@coderabbitai in regards to your latest nitpick comment:

The shadow DOM <slot> elements don't exist until after Stencil's first render() call, so we can't call slot.assignedElements() in componentWillLoad or connectedCallback — there's nothing to query yet.                                                                                                               
                                           
The alternative (querying light DOM with querySelector('[slot="start"]')) is exactly the synchronous DOM check that PR #668 replaced due to race conditions withStencil's async task queue. Re-adding it would reintroduce the original bug (DSS-147).

The current approach defaults to hidden (--empty) and lets slotchange toggle visibility. This is the safe direction — slotchange fires during slot distribution before paint, so there's no visible flicker. A assignedElements fallback is also unnecessary since it shares the same browser support floor as shadow DOM itself.

QuintonJason (approved):

PR #679, opened 2026-02-17
fix(pds-table-row): fix checkbox double-fire and z-index bleed-through

pixelflips:
# Description

Fixes two issues with `pds-table` selectable tables, reported from Kajabi Products (Media Library index page with selectable `DataTableComponent`).

### 1. Checkbox double-fire (`pds-table-row`)

The `pdsTableRowSelected` event's `detail.isSelected` property is unreliable. When clicking a checkbox, the event sometimes fires twice in the same synchronous tick — first with `isSelected: true`, then immediately with `isSelected: false` — even though the user only checked the box.

**Root cause:** In `pds-table-row.tsx`, the checkbox was wired with `onClick` on the `pds-checkbox` element. Inside `pds-checkbox`'s shadow DOM, a `<label htmlFor="X">` wraps `<input id="X">`. When a click lands on the `<label>` (not directly on the `<input>`), the browser's label activation algorithm dispatches a synthetic click on the input, causing the `onClick` handler to fire **twice**. The handler toggled `isSelected` via `!this.isSelected`, so it toggled to `true` then immediately back to `false`.

This was intermittent because it depended on whether the click hit the native `<input>` element directly (one event, correct) or the `<label>` area (two events, broken).

**Fix:** Switch from `onClick` to `onPdsCheckboxChange` on the internal `pds-checkbox`. The `pdsCheckboxChange` event fires exactly once per interaction (from the native input's `onChange`). Reads `event.detail.checked` directly instead of toggling with `!this.isSelected`.

### 2. Z-index bleed-through (`pds-table`)

Sticky checkbox cells in `pds-table-row` and `pds-table-head` use `z-index: var(--pine-z-index-raised)` which resolves to `1000`. When a consuming application has a fixed header with a lower z-index (e.g. `901`), the sticky checkbox cells and toolbar elements render on top of the header when the page is scrolled.

**Fix:** Add `isolation: isolate` to the `pds-table` `:host` styles. This creates a stacking context that scopes all internal z-indices so they don't compete with page-level elements. Internal stacking (sticky cells above regular cells, fixed columns, scroll shadows) continues to work correctly within the table's own stacking context.

Fixes DSS-168

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [x] tested manually

**Test Configuration**:

- Pine versions: local dev build linked to Kajabi Products
- OS: macOS
- Browsers: Chrome
- Tested on: Kajabi Products Media Library index page with selectable DataTableComponent

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-table/pds-table-row/pds-table-row.tsx
+++ b/libs/core/src/components/pds-table/pds-table-row/pds-table-row.tsx
@@ -1,5 +1,6 @@
 import { Component, Element, Host, h, Event, EventEmitter, Prop, State } from '@stencil/core';
 
+import { CheckboxChangeEventDetail } from '../../pds-checkbox/checkbox-interface';
 import { closest } from '../../../utils/closest';
 
 @Component({
@@ -40,8 +41,8 @@ export class PdsTableRow {
    */
   @Event() pdsTableRowSelected: EventEmitter<{ rowIndex: number; isSelected: boolean; }>;
 
-  private handleClick = () => {
-    this.isSelected = !this.isSelected; // Toggle the selected state
+  private handleCheckboxChange = (event: CustomEvent<CheckboxChangeEventDetail>) => {
+    this.isSelected = event.detail.checked;
     this.handleSelect(this.isSelected);
   }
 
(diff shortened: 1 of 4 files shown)


pixelflips:
@coderabbitai review

pixelflips:
@coderabbitai i have addressed the nitpick comment, please review.

sdhull (approved):
looks promising, thank you! I played with the deployment preview and all seems well

QuintonJason (approved):

PR #691, opened 2026-03-05
feat(pds-modal): expose modal CSS part for custom sizing

QuintonJason:
# Description

Adds `part="modal"` to the inner modal container `<div>` in `pds-modal`, exposing it as a named CSS part so consumers can apply a custom `max-width` without relying on undocumented `size` prop values.

Previously, passing `size="large"` (or any non-standard value) would silently generate a CSS class with no matching styles, causing the modal to fall back to near-fullscreen width and triggering linter errors. This gives consumers a clean, supported escape hatch:

```css
pds-modal::part(modal) {
  max-width: 1100px;
}
```

Also adds `**/components.d.ts` and `**/globals.d.ts` to the root `.prettierignore` to prevent the editor from reformatting Stencil-generated declaration files (these were already listed in `libs/core/.prettierignore` but not the root, which the editor's Prettier extension reads from).

Fixes [DSS-183](https://linear.app/kajabi/issue/DSS-183/pds-modal-expose-dialog-css-part-for-custom-sizing)

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions: local
- OS: macOS
- Browsers:
- Screen readers:
- Misc:

# Checklist:


pixelflips (approved):

sdhull:
This is OK—I should be able to target a particular modal by adding `id="foo"` and then in CSS
```
pds-modal#foo::part(modal) {
  max-width: 1100px;
}
```

Right?

sdhull:
Also, this is _not_ a regression (eg, this is the same in published pine docs), but 2 things I noticed:
- The close `x` in the top right of modals on the docs site does not work
- The "Size" section doesn't include React / Web Component source toggles

QuintonJason:
> This is OK—I should be able to target a particular modal by adding `id="foo"` and then in CSS
> 
> ```
> pds-modal#foo::part(modal) {
>   max-width: 1100px;
> }
> ```
> 
> Right?

That is correct, but it could just be
> ```
> #foo::part(modal) {
>   max-width: 1100px;
> }
> ```

QuintonJason:
> Also, this is _not_ a regression (eg, this is the same in published pine docs), but 2 things I noticed:
> 
> * The close `x` in the top right of modals on the docs site does not work
> * The "Size" section doesn't include React / Web Component source toggles

Got some tickets:
- https://linear.app/kajabi/issue/DSS-184/pds-modal-close-button-in-docs-site-header-does-not-work
- https://linear.app/kajabi/issue/DSS-185/pds-modal-sizes-section-in-docs-missing-react-web-component-source

PR #699, opened 2026-04-08
style(pds-input): add text-overflow ellipsis to pds-input placeholder text

keaton-adams:
# Description

Adds `text-overflow: ellipsis` to `.pds-input__field` using the `:placeholder-shown` pseudo-class so that long placeholder text (e.g. "Search in Very Long Saved View Name") truncates with "..." instead of being clipped. The styles only apply when the placeholder is visible, so typed values continue to scroll normally.

## Type of change

- [x] Style change (non-breaking CSS change)

# How Has This Been Tested?

- [x] tested manually

**Test Configuration**:

- OS: macOS
- Browsers: Chrome

# Checklist:

### DIFF
--- a/libs/core/src/components/pds-input/pds-input.scss
+++ b/libs/core/src/components/pds-input/pds-input.scss
@@ -235,9 +235,7 @@
   letter-spacing: var(--pine-letter-spacing);
   min-height: var(--pds-input-field-min-height);
   min-width: var(--pine-dimension-none);
-  overflow: hidden;
   padding: var(--pds-input-padding-y) var(--pds-input-padding-x);
-  text-overflow: ellipsis;
   transition: border-color 0.2s ease;
   width: 100%;
 
@@ -281,6 +279,10 @@
 
   &::placeholder {
     color: var(--pds-input-placeholder-color);
+  }
+
+  &:placeholder-shown {
+    text-overflow: ellipsis;
   }
 
   .has-error & {
--- a/libs/core/src/components/pds-input/pds-input.scss
+++ b/libs/core/src/components/pds-input/pds-input.scss
@@ -235,7 +235,9 @@
   letter-spacing: var(--pine-letter-spacing);
   min-height: var(--pds-input-field-min-height);
   min-width: var(--pine-dimension-none);
+  overflow: hidden;
   padding: var(--pds-input-padding-y) var(--pds-input-padding-x);
+  text-overflow: ellipsis;
   transition: border-color 0.2s ease;
   width: 100%;
 


keaton-adams:
@coderabbit review

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -235,7 +235,9 @@
   letter-spacing: var(--pine-letter-spacing);
   min-height: var(--pds-input-field-min-height);
   min-width: var(--pine-dimension-none);
+  overflow: hidden;
```
pixelflips: `overflow: hidden` and `text-overflow: ellipsis` on the input field itself will also truncate typed values, not just the placeholder. If someone types a long email or URL and tabs away, they'll see something like "some-long-email-addr…" instead of their full input. The value is still there in the DOM, but it looks like it got cut off.

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -235,7 +235,9 @@
...
   padding: var(--pds-input-padding-y) var(--pds-input-padding-x);
+  text-overflow: ellipsis;
```
pixelflips: I am not sure, but could you check if `:placeholder` supports CSS truncation?

pixelflips (changes requested):
A couple of comments on this change truncating values and not only the placeholder text. I am not sure this can be accomplished with CSS alone, but `:placeholder` might work.

Additionally, please change title of PR to use `style(pds-input)` since it's a CSS-only change. This gets used for generating our changelog.

libs/core/src/components/pds-input/pds-input.scss
keaton-adams: It didn't unfortunately but `:placeholder-shown` does!

libs/core/src/components/pds-input/pds-input.scss
keaton-adams: Ah got it, sorry I didn't realize this. I found what I think is a better solution on `:placeholder-shown` instead.

libs/core/src/components/pds-input/pds-input.scss
```diff
@@ -281,6 +281,11 @@
     color: var(--pds-input-placeholder-color);
   }
 
+  &:placeholder-shown {
+    overflow: hidden;
```
pixelflips: Nitpick, but I don't think you'll need the overflow.

pixelflips (approved):
One small nitpick, but this is a nice approach that seems to have solid browser support.

Please don't merge until @QuintonJason has had a chance to review, but LGTM! Nice work!

libs/core/src/components/pds-input/pds-input.scss
keaton-adams: Good catch, cleaned it up.

keaton-adams:
> One small nitpick, but this is a nice approach that seems to have solid browser support.
> 
> Please don't merge until @QuintonJason has had a chance to review, but LGTM! Nice work!

Thanks so much and will do!

QuintonJason (approved):
Looks good. My concern about browser compatibility was addressed

PR #704, opened 2026-04-10
fix: suppress mock ElementInternals console errors in spec tests

pixelflips:
# Description

Stencil's mock-doc fires `console.error` on every `ElementInternals` property access during spec tests, flooding CI output with noise like:

```
NOTE: Property setFormValue was accessed on ElementInternals, but this property is not implemented.
```

This makes it difficult to identify real test failures in CI logs.

Adds a shared `isSpecTest()` utility to `libs/core/src/utils/form.ts` that checks `__STENCIL_SPEC_TESTS__` (the same env var Stencil's mock-doc uses internally) and applies it as an early-return guard in `updateFormValue()` across all 7 form-associated components. This standardizes three inconsistent patterns (inline jest check, try/catch, no guard) into a single shared approach.

Also extracts `updateFormValue()` in pds-combobox to match the pattern used by all other components, and adds a `setFormValue` existence check that was missing.

**Components updated:** pds-input, pds-textarea, pds-select, pds-multiselect, pds-checkbox, pds-switch, pds-combobox

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] tested manually

Verified all 7 affected component spec suites pass with no `console.error` output related to ElementInternals.

**Test Configuration**:

- Pine versions: 3.22.1
- OS: macOS
- Node: 18, 20, 22 (CI matrix)

# Checklist:

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only add a shared spec-test guard around `ElementInternals` interactions, primarily affecting unit-test environments and leaving runtime form behavior intact.
> 
> **Overview**
> Adds a shared `isSpecTest()` helper (based on `__STENCIL_SPEC_TESTS__`) and uses it as an early-return guard to suppress Stencil mock-doc `ElementInternals` console errors during spec tests.
> 
> Standardizes `updateFormValue()` behavior across form-associated components (`pds-checkbox`, `pds-switch`, `pds-input`, `pds-textarea`, `pds-select`, `pds-multiselect`, `pds-combobox`), including extracting a dedicated `updateFormValue(value)` in `pds-combobox` and removing test-only try/catch patterns. Removes a `pds-textarea` unit test that directly asserted `ElementInternals` interactions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 29bc9e5f3b16fcf78011cde1b0bc41bf5cdd9073. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/core/src/components/pds-combobox/pds-combobox.tsx
+++ b/libs/core/src/components/pds-combobox/pds-combobox.tsx
@@ -364,7 +364,7 @@ export class PdsCombobox implements BasePdsProps {
   private updateFormValue(value: string) {
     if (isSpecTest()) return;
 
-    if (this.internals) {
+    if (this.internals && this.internals.setFormValue) {
       this.internals.setFormValue(value);
     }
   }
--- a/libs/core/src/components/pds-combobox/pds-combobox.tsx
+++ b/libs/core/src/components/pds-combobox/pds-combobox.tsx
@@ -1,5 +1,6 @@
 import { Component, Element, Event, EventEmitter, h, Host, Prop, State, Watch, Method } from '@stencil/core';
 import type { BasePdsProps } from '@utils/interfaces';
+import { isSpecTest } from '../../utils/form';
 import type { ChipSentimentType } from '@utils/types';
 import { computePosition, flip, offset, shift } from '@floating-ui/dom';
 import { debounceEvent } from '@utils/utils';
@@ -285,13 +286,7 @@ export class PdsCombobox implements BasePdsProps {
     }
 
     // Initialize form value with current value
-    if (this.internals) {
-      try {
-        this.internals.setFormValue(this.selectedOption?.value ?? this.value ?? '');
-      } catch (e) {
-        // ElementInternals.setFormValue not available in unit tests
-      }
-    }
+    this.updateFormValue(this.selectedOption?.value ?? this.value ?? '');
   }
 
   disconnectedCallback() {
@@ -319,13 +314,7 @@ export class PdsCombobox implements BasePdsProps {
   handleValueChange() {
     this.filterOptions();
     // Sync with form internals for form association
-    if (this.internals) {
-      try {
-        this.internals.setFormValue(this.value);
-      } catch (e) {
-        // ElementInternals.setFormValue not available in unit tests
-      }
-    }
+    this.updateFormValue(this.value);
 
     // Find and select option that matches the value (for external value changes)
     // Only do this if we're not already updating from a selection
@@ -364,23 +353,19 @@ export class PdsCombobox implements BasePdsProps {
       this.displayText = this.getOptionLabel(this.selectedOption);
       this.value = this.selectedOption.value;
       // Update form internals with the actual option value
-      if (this.internals) {
-        try {
-          this.internals.setFormValue(this.selectedOption.value);
-        } catch (e) {
-          // ElementInternals.setFormValue not available in unit tests
-        }
-      }
+      this.updateFormValue(this.selectedOption.value);
     } else {
       this.displayText = '';
       this.value = '';
-      if (this.internals) {
-        try {
-          this.internals.setFormValue('');
-        } catch (e) {
-          // ElementInternals.setFormValue not available in unit tests
-        }
-      }
+      this.updateFormValue('');
+    }
+  }
+
+  private updateFormValue(value: string) {
+    if (isSpecTest()) return;
+
+    if (this.internals) {
+      this.internals.setFormValue(value);
     }
   }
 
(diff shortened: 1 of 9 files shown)


libs/core/src/utils/form.ts
```diff
@@ -45,3 +45,16 @@ export function exposeTypeProperty(element: Element, type: string | (() => strin
...
+ * Determines if code is running in Stencil's spec test environment.
+ * Stencil's mock-doc logs console.error on any ElementInternals
+ * property access during spec tests. Use this guard to skip
+ * ElementInternals calls in the mock environment.
+ *
+ * Uses __STENCIL_SPEC_TESTS__ (the same env var Stencil's mock-doc checks)
+ * to avoid false positives in downstream consumer test suites.
+ */
+export const isSpecTest = (): boolean => {
+  return typeof process !== 'undefined' && process.env?.__STENCIL_SPEC_TESTS__ === 'true';
```
QuintonJason, line 59: **Consumer / changelog:** `pds-checkbox` and `pds-switch` used to bail out of `updateFormValue()` whenever `jest` was defined or `NODE_ENV === "test"`. This now only skips when `__STENCIL_SPEC_TESTS__` is set (Stencil mock-doc). Apps that render these components in their own Jest suite without that env var may start hitting `setFormValue` / `setValidity` where they did not before — worth a short release note if external consumers matter.

libs/core/src/components/pds-textarea/pds-textarea.tsx
```diff
@@ -440,6 +440,8 @@ export class PdsTextarea {
    * Updates the form value using ElementInternals API
    */
   private updateFormValue() {
```
QuintonJason, line 442: This clears mock-doc noise for `updateFormValue()`. Heads-up: `maxLengthChanged` still calls `this.internals.setValidity(...)` (around lines 216–222) with no `isSpecTest()` guard. If any textarea specs still log ElementInternals warnings, that path may need the same treatment.

libs/core/src/components/pds-combobox/pds-combobox.tsx
```diff
@@ -364,23 +353,19 @@ export class PdsCombobox implements BasePdsProps {
...
-        }
-      }
+      this.updateFormValue('');
+    }
+  }
+
+  private updateFormValue(value: string) {
+    if (isSpecTest()) return;
+
+    if (this.internals && this.internals.setFormValue) {
```
QuintonJason, line 367: Cleaner than the previous try/catch blocks, and the `setFormValue` existence check matches the other form-associated components.

QuintonJason:
**Coverage:** Removing the textarea spec that mocked `internals` makes sense with `isSpecTest()` short-circuiting `updateFormValue()` in spec runs. Just noting that unit tests no longer assert the `setFormValue` + `setValidity` path together in textarea; relying on e2e for real `ElementInternals` is reasonable.

**Title (optional):** Repo convention often uses a scope, e.g. `fix(core):` or `fix(pds-input):` — current `fix:` is fine if you prefer a multi-component title.

**Netlify:** Failing preview here is likely environmental (e.g. lockfile/CI alignment with other open PRs) rather than this change set — worth confirming against `main` if it persists.

QuintonJason (commented):
Left inline notes on `isSpecTest()` (consumer/changelog), textarea (`maxLengthChanged` / `setValidity` follow-up), and combobox refactor. One conversation comment covers coverage, title scope, and Netlify.

libs/core/src/utils/form.ts
pixelflips: Good call — will add a release note when this ships. The behavior change for checkbox/switch is that consumer Jest suites may now see `setFormValue` execute where it previously bailed out. In practice this is a no-op since jsdom doesn't implement ElementInternals, but worth documenting.

libs/core/src/components/pds-textarea/pds-textarea.tsx
pixelflips: Nice catch — added the `isSpecTest()` guard to the `maxLengthChanged` `setValidity` call in 74ad599f.

libs/core/src/components/pds-combobox/pds-combobox.tsx
pixelflips: Thanks — agreed, much cleaner than the scattered try/catch blocks.

pixelflips:
Thanks for the thorough review. Addressed the `maxLengthChanged` gap in 74ad599f — that was the only remaining unguarded `this.internals` access outside of `updateFormValue()`. Will add a release note for the checkbox/switch behavior change when this ships.

QuintonJason (approved):

PR #665, opened 2026-02-06
fix(pds-input): style native search clear button with custom icon

pixelflips:
# Description

Adds custom styling for the native search clear (x) button on `<input type="search">` within `pds-input`. The browser default clear button styling is replaced with a custom X icon using a mask-image SVG, colored with the `--pds-input-placeholder-color` token for a subtle appearance consistent with the input's design language.

Also commits the auto-generated `createComponent.tsx` from `@stencil/react-output-target@0.5.3`. This file was outdated in the repo — the current dependency version generates it with a `customElements.get()` guard to prevent double-registration. Every local build regenerates this file, causing a persistent diff. This commit syncs it with the current build output.

Fixes [DSS-137](https://linear.app/kajabi/issue/DSS-137/datatable-search-clear-icon-styling)

### Screenshots
||  Before   |  After  |
|--------|--------|--------|
|Light|Screenshot 2026-02-06 at 10 59 58 AM|Screenshot 2026-02-06 at 10 58 41 AM|
|Dark|Screenshot 2026-02-06 at 11 00 09 AM|Screenshot 2026-02-06 at 10 58 26 AM|

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] This change requires a documentation update

# How Has This Been Tested?

- [x] tested manually

**Test Configuration**:

- OS: macOS
- Browsers: Chrome, Safari, Firefox (Firefox does not render a native search clear button)

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

PR #705, opened 2026-04-10
feat(pds-multiselect): add grouped options support

QuintonJason:
# Description

Adds grouped option support to `pds-multiselect` so it can replace `Sage::SelectDropdown` in cases where options need to be organized under labeled category headers (e.g., grouping products by type: Evergreen Courses, Cohort Courses, Newsletters, Communities).

A Kajabi team needed grouped product categories in a multiselect and reached for `Sage::SelectDropdown` because `pds-multiselect` lacked grouping. This closes that gap.

Linear: https://linear.app/kajabi/issue/DSS-197/featpds-multiselect-add-grouped-options-support

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] tested manually (Storybook)

**Test Configuration**:

- OS: macOS
- Browsers: Chrome

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Adds grouped rendering to `pds-multiselect` by changing option parsing and listbox markup, which could affect selection/highlighting behavior and accessibility if edge cases slip through. Changes are UI-scoped and covered by new unit tests and docs, keeping overall risk moderate.
> 
> **Overview**
> `pds-multiselect` now supports **grouped options**, rendering non-interactive group headers (from slotted `<optgroup>` elements or a new `group` field on `options`) and hiding empty groups during filtering.
> 
> This updates slot parsing to extract grouped options, refactors option rendering into a shared helper, adds new group styling/ARIA structure, and expands docs + Storybook examples + unit tests to cover grouped behavior and contiguity rules.
> 
> Separately, several components document/expose additional **shadow parts** (e.g. `pds-multiselect` `trigger`, `pds-alert`/`pds-toast` `dismiss`, and button-related parts for avatar/chip/dropdown-menu-item), and the generated typings/READMEs reflect these updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39e860cd507bb79b29f8a6f2c5688c5c720ddf15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


QuintonJason:
### Code review

Found 2 issues:

1. `components.d.ts` and `readme.md` are auto-generated and must not be manually edited. The diff includes JSDoc comments added directly to `components.d.ts` for `pds-popover` (unrelated to this PR's scope) and a Shadow Parts table in `readme.md` with an empty `trigger` description — both are signs of manual edits rather than build output. Per [CLAUDE.md](https://github.com/Kajabi/pine/blob/main/CLAUDE.md): _"Never: edit auto-generated files (`readme.md`, `components.d.ts`) manually."_ The `@part trigger - <description>` annotation should be added to the class JSDoc in `pds-multiselect.tsx` and the files regenerated via `npx nx run @pine-ds/core:build`.

https://github.com/Kajabi/pine/blob/bdd71e51ef86cdbdfd28ac866e03164c5f7d45bd/libs/core/src/components/pds-multiselect/pds-multiselect.tsx#L1148-L1160

2. Group headers are rendered with `role="presentation" aria-hidden="true"`, which makes group names completely invisible to screen readers. Users with assistive technologies will hear individual options with no group context. The WAI-ARIA listbox pattern supports `role="group"` with `aria-label` for this purpose (mirrors native `<optgroup>` semantics):

```html
<li role="group" aria-label={item.group}>
  <!-- option children would need to be nested here -->
</li>
```

Note: implementing `role="group"` properly requires wrapping the group's option `<li>` children inside it, which is a structural change to `getGroupedRenderItems`. Alternatively, `role="presentation"` can be kept for the header element but a visually-hidden `aria-label` added to each option via `aria-describedby` to announce group membership.

https://github.com/Kajabi/pine/blob/bdd71e51ef86cdbdfd28ac866e03164c5f7d45bd/libs/core/src/components/pds-multiselect/pds-multiselect.tsx#L1148-L1160

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -1107,47 +1191,21 @@ export class PdsMultiselect {
...
+            if (item.type === 'group') {
+              return (
+                <li key={`group-${itemIndex}`} role="presentation">
+                  <span class="pds-multiselect__group-header" aria-hidden="true">
+                    {item.group}
+                  </span>
+                  <ul role="group" aria-label={item.group} class="pds-multiselect__group-list">
+                    {item.options.map(({ option, index }) => this.renderOption(option, index, valueArray))}
+                  </ul>
+                </li>
```
pixelflips: The `<li role="presentation">` wrapper between `role="listbox"` and `role="group"` may break the ARIA ownership chain. WAI-ARIA expects `role="group"` to be directly owned by the `role="listbox"` element — the intermediate presentation `<li>` collapses semantically, potentially leaving the group orphaned from the listbox in strict screen readers (JAWS, NVDA).

The intended structure for grouped listbox options is:
```html
<ul role="listbox">
  <li role="group" aria-label="Group Name">
    <span class="group-header" aria-hidden="true">Group Name</span>
    <ul class="group-list">
      <li role="option">...</li>
    </ul>
  </li>
</ul>
```

Or alternatively, flatten the group so `role="option"` elements remain direct children of `role="listbox"` while associating the header via `aria-describedby`.

Worth validating with VoiceOver / NVDA to confirm whether the current nesting causes options to drop out of the listbox context.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -1021,6 +1036,75 @@ export class PdsMultiselect {
...
+        if (!currentGroup || currentGroup.group !== group) {
+          currentGroup = { type: 'group', group, options: [] };
+          items.push(currentGroup);
+        }
+        currentGroup.options.push({ option, index });
+      } else {
+        currentGroup = null;
+        items.push({ type: 'option', option, index });
+      }
+    });
```
pixelflips, line 1065: `getGroupedRenderItems` uses a `currentGroup` pointer that only starts a new group header when `currentGroup.group !== group`. This means non-contiguous items with the same group name (e.g. `[{group:'A'}, {group:'B'}, {group:'A'}]`) will produce **duplicate group headers** for 'A'.

This mirrors native `<optgroup>` behavior (which also doesn't merge non-contiguous groups), but consumers using the `options` prop may not expect it — especially if building arrays from unsorted API data.

Consider either:
1. Documenting the contiguity requirement on the `group` field in `MultiselectOption` and in the `options` prop JSDoc
2. Or consolidating same-named groups regardless of input order (using a `Map<string, GroupItem>`)

At minimum, a JSDoc note on the `group` field like `/** Group label. Options with the same group must be contiguous in the array. */` would prevent silent misuse.

libs/core/src/components/pds-multiselect/pds-multiselect.scss
```diff
@@ -173,6 +173,21 @@
   padding: var(--pine-dimension-xs);
 }
 
+.pds-multiselect__group-header {
+  color: var(--pine-color-text-muted);
+  display: block;
+  font: var(--pine-typography-body-sm-medium);
+  letter-spacing: var(--pine-letter-spacing);
+  padding: var(--pine-dimension-sm) var(--pine-dimension-xs) var(--pine-dimension-2xs);
```
pixelflips, line 181: All group headers now use `padding-block-start: var(--pine-dimension-sm)`. The earlier iteration had a `:first-child` override that gave the first group header smaller top padding (`--pine-dimension-xs`) so it sat closer to the search input. That rule was removed during the `role="group"` refactor without a replacement.

This may be intentional, but if the original tighter spacing on the first header was a design decision, consider restoring it — e.g.:
```scss
li:first-child > .pds-multiselect__group-header {
  padding-block-start: var(--pine-dimension-xs);
}
```

pixelflips (commented):
Code review — 3 inline comments on the grouped options implementation.

libs/react/src/components/react-component-lib/createComponent.tsx
```diff
@@ -26,10 +26,7 @@ export const createReactComponent = <
   defineCustomElement?: () => void
 ) => {
   if (defineCustomElement !== undefined) {
-    // Guard against double-registration when CDN and @pine-ds/react are both loaded
-    if (typeof customElements !== 'undefined' && !customElements.get(tagName)) {
-      defineCustomElement();
-    }
+    defineCustomElement();
```
pixelflips: **Regression: double-registration guard removed**

This commit removed the `customElements.get(tagName)` guard that exists on `main` (added in PR #665). That guard prevents a `NotSupportedError` crash when both the CDN bundle and `@pine-ds/react` are loaded on the same page -- a real production scenario.

`main` has:
```ts
if (typeof customElements !== 'undefined' && !customElements.get(tagName)) {
  defineCustomElement();
}
```

This PR replaced it with a bare `defineCustomElement()` call. The removal is unrelated to grouped options and will regress any consumer that dual-loads CDN + React. This should be reverted to match `main`.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -1107,47 +1193,26 @@ export class PdsMultiselect {
...
+                <li
+                  key={`group-${itemIndex}`}
+                  role="group"
+                  aria-label={item.group}
+                  class="pds-multiselect__group"
+                >
+                  <span class="pds-multiselect__group-header" aria-hidden="true">
+                    {item.group}
+                  </span>
+                  <ul class="pds-multiselect__group-list">
```
pixelflips: **ARIA ownership chain broken -- missing `role="presentation"`**

This `<ul>` has no explicit role, so it defaults to `role="list"`, which breaks the WAI-ARIA ownership chain between `role="group"` and the nested `role="option"` elements. Screen readers (JAWS, NVDA) expect `role="option"` to be directly owned by `role="listbox"` or `role="group"` -- the intermediate implicit `role="list"` severs that link.

This was flagged by both Cursor Bugbot (high severity) and pixelflips on this same PR and remains unaddressed.

As a cascading consequence, `aria-activedescendant` on the combobox input (which points to option IDs inside this `<ul>`) may also be silently ignored by screen readers for grouped options, breaking keyboard navigation announcements.

Fix: add `role="presentation"` to this `<ul>` to make it semantically transparent:
```tsx
<ul class="pds-multiselect__group-list" role="presentation">
```

libs/core/src/components/pds-multiselect/docs/pds-multiselect.mdx
```diff
@@ -740,6 +740,122 @@ Call the `clear()` method to programmatically reset all selections, clear the se
...
+  options={products}
+/>`,
+  }}
+>
+  <pds-multiselect
+    component-id="grouped-options-select"
+    label="Products"
+    placeholder="Select products..."
+  >
+    <optgroup label="Evergreen Courses">
```
pixelflips: **Live preview contradicts prose above it**

The paragraph at line 810 says "For dynamic or async data, add a `group` field to each item in the `options` prop instead" -- but the `DocCanvas` children here (the live preview) use `<optgroup>` slot elements rather than the `options` prop. The `mdxSource` tabs correctly show the `options` prop approach, but the rendered preview demonstrates the slot-based approach.

This mismatch will confuse developers reading the docs, since the live preview doesn't match what's being described. The children should use the `options` prop (set via a `<script>` tag or similar) to match the documented pattern.

pixelflips (commented):
Code review -- 4 issues found (2 critical, 2 moderate). Checked for bugs, ARIA correctness, and CLAUDE.md compliance.

libs/core/src/components/pds-chip/readme.md
```diff
@@ -35,6 +35,13 @@
 | `"(default)"` | The chip's label text. |
 
 
+## Shadow Parts
```
pixelflips, line 38: Not ablocker, but should all these readme.md files be updated?

pixelflips (approved):
Small comment on the additional readme files, otherwise LGTM! 👍🏼

libs/core/src/components/pds-chip/readme.md
QuintonJason: Added `@part` JSDoc annotations to all 5 components whose readme.md sections were manually added without a corresponding source annotation (pds-alert, pds-avatar, pds-chip, pds-dropdown-menu-item, pds-toast) in 39e860cd. All other components with shadow parts already had them documented (either via `@part` JSDoc + build, or prior manual additions).

PR #713, opened 2026-04-20
feat(pds-container): add container component

QuintonJason:
# Description

Introduces the **`pds-container`** web component as the max-width, horizontally centered layout wrapper in the Pine layout system

The component:

- Constrains width via a **`size`** prop using named tokens (`sm`, `md`, `lg`, `xl`, `full`) or any valid CSS length (e.g. `640px`, `50rem`) via the `--pine-container-max-width` custom property.
- Renders a configurable **semantic** inner element with **`tag`** (`div` | `main` | `section` | `article`, default `div`).
- Optional **`centered`** (default `true`) to center with `margin-inline: auto` when a max-width applies.

Documentation and Storybook: MDX, stories under **Components → Layout → Container**, Stencil `readme`, and the **Layout** guide updated to describe `pds-container` and how it works with `pds-row` / `pds-box`. React consumers get the generated proxy. Unit tests cover `size`, `tag`, and `centered`.

Fixes #(issue)

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions: (fill after running locally)
- OS:
- Browsers:
- Screen readers:
- Misc: `pnpm` / Nx targets as usual for core + react packages

# Checklist:


QuintonJason:
### Code review

Found 3 issues:

1. The `NotCentered` story (and the Storybook controls for `centered`) will never render a non-centered container. The `BaseTemplate` uses Lit's boolean directive `?centered="${args.centered}"`, which **removes the attribute entirely** when `args.centered` is `false`. Stencil sees no attribute and falls back to the default `true`. The fix is to use a string binding instead: `centered="${args.centered}"`.

https://github.com/Kajabi/pine/blob/d946a71dccd0ed6d528e8fef217f516ffa66856a/libs/core/src/components/pds-container/stories/pds-container.stories.js#L8-L15

2. The `Default` story renders `size="undefined"` as a literal string attribute because `Default.args` does not include a `size` key and `args.size` is `undefined`. The component receives the string `"undefined"`, which is not in `NAMED_SIZES`, so it is treated as a custom length and sets `--pine-container-max-width: undefined` (invalid CSS) on the host. The template should guard the attribute: `${args.size ? `size="${args.size}"` : ''}`.

https://github.com/Kajabi/pine/blob/d946a71dccd0ed6d528e8fef217f516ffa66856a/libs/core/src/components/pds-container/stories/pds-container.stories.js#L8-L22

3. The "Named sizes" table in the `## Guidelines` section of the component docs only lists `lg`, `xl`, and `full`. The `sm` (340px) and `md` (520px) sizes are omitted, even though both are implemented in the SCSS and documented in the `@Prop` JSDoc. The examples section lower on the same page correctly shows all five.

https://github.com/Kajabi/pine/blob/d946a71dccd0ed6d528e8fef217f516ffa66856a/libs/core/src/components/pds-container/docs/pds-container.mdx#L22-L47

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

pixelflips (approved):
Ran a few reviews and also manually checked via Storybook. LGTM! Great work!

PR #720, opened 2026-04-29
fix: sync package-lock.json with package.json

QuintonJason:
## What

Updates `package-lock.json` to match the current `package.json` dependency tree.

## Why

`npm ci` fails in CI because the lock file is out of sync — several packages are missing (React 19.2.5 transitive deps, @testing-library/dom, etc.). This unblocks the Continuous Deployment workflow so we can cut a new `@pine-ds/core` release picking up `@kajabi-ui/styles@1.1.1`.

## Test plan

- [ ] CI passes (build-core, build-react, build-doc-components all green)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3482,6 +3482,7 @@
       "os": [
         "aix"
       ],
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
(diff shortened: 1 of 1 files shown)


pixelflips (approved):
We still need to figure out what is going on here. Getting to be ridiculous. 😢

PR #731, opened 2026-05-14
chore(ci): self-heal lockfile after release to break npm-ci drift cycle

QuintonJason:
# Description

Every release leaves `main` in an `npm ci`-incompatible state. The release commit
6b773e58 ("chore(release): release 3.25.0") rewrote `package-lock.json` with 262
deletions / 35 additions, removing transitive deps of `@testing-library/jest-dom`
(`@testing-library/dom`, `dom-accessibility-api`, `lz-string`, `pretty-format`,
`react-is`, `@types/aria-query`, plus the optimistic `react@19.x` peer
resolutions). Every PR build that followed then fails:

```
npm ci ... Missing: @testing-library/dom@10.4.1 from lock file
                    react@19.2.6 from lock file
                    ...
```

This is recurring: `git log` shows ~10 manual "sync package-lock.json" commits
over recent history (#728, #720, #719, …). Each release flips the lockfile to a
shape `npm ci` rejects; someone hand-fixes it; the next release does it again.

## Root cause

`nx release --yes` runs `npm install` internally to refresh the lockfile after
version bumps. The release runner's npm resolves peer deps differently than the
build runner's `npm ci` expects, producing a lockfile that **commits cleanly but
fails strict validation**. The release commit is pushed before anything detects
the drift.

The existing `scripts/verify-lockfile-sync.cjs` pre-push guard only inspects
direct `dependencies`/`devDependencies`/`peerDependencies` blocks in each
workspace's `package.json` — it does **not** check the transitive graph that
`npm ci` validates. So it greenlights lockfiles that CI then rejects, giving
false confidence.

## Fix

This PR does three things:

1. **Post-release self-heal step** in `schedule-release.yml`. After
   `nx release` pushes the version-bump commit, run `npm ci`. If it fails,
   regenerate `package-lock.json` with `npm install`, commit, and push a
   `chore: sync package-lock.json after release` fix-up. The commit uses
   `chore:` (not `fix:`) so it doesn't trigger another patch version bump on
   the next `nx release` run — the heal is plumbing, not a user-facing bug
   fix. Guarded against an infinite heal loop (aborts if the regenerated
   lockfile is byte-identical to the broken one).
2. **Remove the false-green local guard**: delete
   `scripts/verify-lockfile-sync.cjs`, its `lefthook.yml` pre-push entry, and
   the `verify-lockfile` npm script. CI's real `npm ci` step is the source of
   truth; the local check was misleading contributors into thinking their
   lockfile was fine.
3. **One-time lockfile regen** (commit `85bc0b74`,
   `fix: sync package-lock.json after release`). The workflow change only
   takes effect on future releases — it can't repair main retroactively.
   Without this commit, every open PR (including this one) would stay red
   waiting for the next release to heal `main`. Running `npm install` locally
   reproduced the canonical `npm ci`-compatible lockfile and added back the
   256 lines the 3.25.0 release removed. Validated locally with
   `rm -rf node_modules && npm ci --ignore-scripts` — passes clean. Kept as
   `fix:` (not `chore:`) because it repairs a consumer-visible bug worth a
   patch release and a changelog entry.

## Type of change

# How Has This Been Tested?

- [x] tested manually:
  - Ran `npm install --ignore-scripts` locally against the broken main
    lockfile; produced a `package-lock.json` that added back the missing
    transitive deps the release commit had stripped.
  - Ran `rm -rf node_modules && npm ci --ignore-scripts` against the
    regenerated lockfile; passes without `Missing: X from lock file` errors.
  - Validated the workflow YAML parses with `js-yaml`, package.json parses
    as JSON, all hooks-and-scripts references to the deleted file are gone
    (grep clean).
- [ ] other: The self-heal step itself will be exercised by the next release
  run. The infinite-loop guard ensures that if the heal can't produce a
  different lockfile, the job goes red rather than churning commits.

# Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches the release GitHub Actions workflow and auto-pushes follow-up commits, so a mistake could break releases or create unexpected commits even though the logic is guarded by `npm ci` checks.
> 
> **Overview**
> Prevents release runs from leaving `main` with an `npm ci`-incompatible `package-lock.json` by adding a post-`nx release` workflow step that runs `npm ci` and, on failure, regenerates the lockfile via `npm install`, validates it with `npm ci`, then commits and pushes a `chore:` fix-up.
> 
> Removes the local lockfile verification guard (`scripts/verify-lockfile-sync.cjs`), its `lefthook.yml` pre-push hook, and the `verify-lockfile` npm script, and updates `package-lock.json` to restore missing transitive/peer dependency entries so CI can install cleanly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 98da1421e4fcd57cf64956c61c0c421668c90d29. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pixelflips (approved):
Ran checks locally and no issues found. LGTM! 🤞🏼

PR #769, opened 2026-06-22
chore: pin @kajabi-ui/styles to ^1.4.0 in published packages

QuintonJason:
# Description

The published packages `@pine-ds/core` (`libs/core`) and `@pine-ds/react` (`libs/react`) depend on `@kajabi-ui/styles` with a wildcard `"*"` range. This means any consumer (kajabi-products and other apps) silently pulls *whatever* the latest published token version is, with no upper bound and no ability to pin a known-good token version through Pine. A breaking or regressing token release can reach every consumer with no version gate.

This pins both to `^1.4.0` — the current published version, and the same range the root `package.json` already uses. Caret keeps automatic minor/patch intake while restoring an explicit, pinnable contract.

No code or runtime behavior changes; tokens are consumed at runtime as CSS custom properties either way.

Note: `@pine-ds/icons` is also still wildcard `"*"` in both packages — same argument applies, but left out to keep this PR tightly scoped. Worth a follow-up.

## Type of change

# How Has This Been Tested?

- [x] other: dependency-range-only change. CI build + existing spec/e2e suites exercise that Pine still builds and renders against `@kajabi-ui/styles@^1.4.0`.

# Checklist:

---

> [!NOTE]
> **Low Risk**
> Manifest-only dependency range change with no code or runtime behavior changes; it reduces unbounded token drift for consumers.
> 
> **Overview**
> Published **`@pine-ds/core`** and **`@pine-ds/react`** no longer declare `@kajabi-ui/styles` as `"*"`. Both now depend on **`^1.4.0`**, matching the root workspace range.
> 
> That gives consumers a bounded, pinnable contract on design tokens while still allowing compatible minor/patch updates. No application or runtime code changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d40c816a1f5b8cf8f7d763d81ce2b826704b7a2d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/core/package.json
+++ b/libs/core/package.json
@@ -72,7 +72,7 @@
   },
   "dependencies": {
     "@floating-ui/dom": "^1.7.0",
-    "@kajabi-ui/styles": "*",
+    "@kajabi-ui/styles": "^1.4.0",
     "@pine-ds/icons": "*",
     "@stencil/core": "4.43.5",
     "@types/dompurify": "^3.0.5",
--- a/libs/react/package.json
+++ b/libs/react/package.json
@@ -37,7 +37,7 @@
     "tsc": "tsc -p ."
   },
   "dependencies": {
-    "@kajabi-ui/styles": "*",
+    "@kajabi-ui/styles": "^1.4.0",
     "@pine-ds/core": "^3.26.4",
     "@pine-ds/icons": "*",
     "tslib": "*"


pixelflips:
@QuintonJason I was poking at this during a local review session with Claude, and I'm not totally clear on why we need this right now, so I figured I'd ask before approving.

The reasoning here is that consumers (kajabi-products + others) "silently pull whatever the latest token version is", but digging into kajabi-products, that doesn't really seem to be happening:

🤖 is saying:

- The tokens that actually render come from a hardcoded CDN <link> pinned to @1.4.0 in app/views/shared/_pine_assets.html.erb — totally outside npm.
- kajabi-products already pins @kajabi-ui/styles directly (1.4.0 devDep) and has it in a resolutions override, which forces the whole tree — including Pine's transitive "*" — down to 1.4.0.
- Its yarn.lock backs this up: "@kajabi-ui/styles@*", "@kajabi-ui/styles@1.4.0": version "1.4.0" — single resolved copy, no drift.

One other thing: in core, it's really only used at build/lint time (the SCSS @use directive gets inlined into the dist CSS), so the runtime range may not do much either way.

Can you double-check these on your end and let me know what's driving the need? If there's a specific consumer or scenario I'm missing, just let me know. Happy to approve, but not clear on why this is needed since we haven't needed it before.

QuintonJason:
You're right, and thanks for pushing on this before approving — I verified your points independently and they hold:

- **Runtime tokens come from the pinned CDN `<link>`, not npm.** `app/views/shared/_pine_assets.html.erb` hardcodes `@kajabi-ui/styles@1.4.0` via jsDelivr, entirely outside the npm graph.
- **kajabi-products already forces the version** — `1.4.0` direct devDep + a `resolutions` override, and `yarn.lock` confirms `"@kajabi-ui/styles@*", "@kajabi-ui/styles@1.4.0"` collapse to a single `1.4.0` copy.
- **In Pine core it's build/lint-time only** (SCSS `@use` inlined into dist CSS), so the runtime range does nothing either way.

So this pin is a genuine no-op for what any consumer actually installs or renders. My original reasoning ("consumers silently pull whatever the latest token version is") was based on reading Pine's manifest in isolation without checking how the tokens actually load or how kajabi-products pins — my mistake. Closing this.

The written contract guidance (pin real ranges, not `*`) lives in the VERSIONING.md doc PR instead, which is where it belongs — and kajabi-products already follows it.

QuintonJason:
Closing — verified no-op; contract is already pinned at the consumer (CDN link + resolutions + lockfile). See prior comment.

PR #164, opened 2024-04-10
chore: update @pine-ds/icons to use latest version

ju-Skinner:
This will reduce the interactions from DSS

# Description

To ensure that we use the latest versions of `@pine-ds/icons`, we changed the package to use `*`. This will reduce the amount of times we have to update on our end. When a release occurs for `@pine-ds/core` or `@pine-ds/react`, we can then just run `npm update @pine-ds/icons` which in turn will get the latest version. 

Fixes #(issue)
https://kajabi.atlassian.net/browse/DSS-558

## Type of change

Please delete options that are not relevant.
If your type of change is not present, add that option.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist:

If not applicable, leave options unchecked.

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing tests pass locally with my changes
- [ ] Design has QA'ed and approved this PR

PR #770, opened 2026-06-22
chore: pin @pine-ds/icons to ^10.0.0 in published packages

QuintonJason:
# Description

Follow-up to #769 (which pinned `@kajabi-ui/styles`). Same issue, same fix, separate PR for clean review.

`@pine-ds/core` (`libs/core`) and `@pine-ds/react` (`libs/react`) depend on `@pine-ds/icons` with a wildcard `"*"` range. Consumers silently pull whatever icon version is latest, with no upper bound and no pinnable contract through Pine — an icon-package release can reach every consumer with no version gate.

Pins both to `^10.0.0` (current published version). Caret keeps automatic minor/patch intake while restoring an explicit range.

No code or runtime behavior changes.

## Type of change

# How Has This Been Tested?

- [x] other: dependency-range-only change; CI build + existing suites exercise that Pine builds against `@pine-ds/icons@^10.0.0`.

# Checklist:

---

> [!NOTE]
> **Low Risk**
> Dependency manifest-only change with no runtime or logic edits; risk is limited to install resolution staying within ^10.x for icons.
> 
> **Overview**
> Pins **`@pine-ds/icons`** from wildcard **`"*"`** to **`^10.0.0`** in **`@pine-ds/core`** and **`@pine-ds/react`** so consumers get an explicit semver contract instead of silently resolving the latest icons release.
> 
> No application or component code changes; only **`libs/core/package.json`** and **`libs/react/package.json`** dependency ranges are updated, mirroring the earlier **`@kajabi-ui/styles`** pin.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08dfef2255fde218ebe8f24efaeb51cb7b5c3afd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/core/package.json
+++ b/libs/core/package.json
@@ -73,7 +73,7 @@
   "dependencies": {
     "@floating-ui/dom": "^1.7.0",
     "@kajabi-ui/styles": "*",
-    "@pine-ds/icons": "*",
+    "@pine-ds/icons": "^10.0.0",
     "@stencil/core": "4.43.5",
     "@types/dompurify": "^3.0.5",
     "dompurify": "^3.2.6",
--- a/libs/react/package.json
+++ b/libs/react/package.json
@@ -39,7 +39,7 @@
   "dependencies": {
     "@kajabi-ui/styles": "*",
     "@pine-ds/core": "^3.26.4",
-    "@pine-ds/icons": "*",
+    "@pine-ds/icons": "^10.0.0",
     "tslib": "*"
   },
   "devDependencies": {


pixelflips:
@QuintonJason same as on #769, I'm trying to understand the motivation here before we merge, since this one looks like it reverses an intentional call.

🤖 is saying (which i seem to recall as well)

> The @pine-ds/icons: "*" wasn't an oversight — it's been the wildcard since the dep was first added (back when it was @sage/icons in 2023), and it was deliberately set back to * in #164 (Apr 2024, "update @pine-ds/icons to use latest version"). The git log right before that even shows us hand-bumping ^3.6.0 → ^3.8.0, which is exactly the manual toil * was chosen to avoid — so Pine always tracks the latest icons without us babysitting the version.
> 
> A couple of things that make me unsure this is needed now:
> - It's effectively a no-op today — npm latest for @pine-ds/icons is 10.0.0, so * and ^10.0.0 resolve to the same thing.
> - It reintroduces the manual-bump burden, and would freeze Pine on 10.x until someone bumps to 11.x.

Is there a specific problem or breakage that's prompting this? Totally open to it if there's a concrete reason, just want to make sure we're not undoing #164 which has seemd to work until now?

Also, while reviewing looks like Sage is way out of date (^9.12.0), we should prob bump that at some point.

QuintonJason:
Agreed — closing this one, and thanks for catching it. You're right that this reverses an intentional call, not an oversight:

- The `*` was deliberately set in #164 (Apr 2024) to avoid hand-bumping the icons version, and the git log right before it (`^3.6.0` → `^3.8.0` manual bumps) shows exactly the toil it was chosen to eliminate.
- It's a no-op today (`*` and `^10.0.0` both resolve to `10.0.0`), and pinning would reintroduce the manual-bump burden and freeze Pine on 10.x until someone bumps to 11.x.

So this is worse than the tokens one — it undoes a working decision for no benefit. My audit flagged the wildcard from the manifest alone without the #164 context. Closing.

(Noted your aside on Sage being on `^9.12.0` — separate hygiene item, worth a ticket.)

PR #785, opened 2026-08-05
feat(pds-tabs): add divider and stretch (fill-height) props

QuintonJason:
# Description

Adds two opt-in props to `pds-tabs` that consumers currently can't get without reaching into the shadow DOM via `::part(tab-list)` / forcing panel heights:

- **`divider`** (boolean) — draws a rule beneath the tab list (`border-block-end` on `.pds-tabs__tablist`), the standard underlined-tabs look. Addresses **[DSS-219](https://linear.app/kajabi/issue/DSS-219)**.
- **`stretch`** (boolean) — makes the component a flex column so the **active tab panel fills the remaining height** (tab list stays fixed at top). Addresses **[DSS-220](https://linear.app/kajabi/issue/DSS-220)** (High). Implemented by passing `stretch` down to `pds-tabpanel` (same mechanism as `variant`/`selected`); the active panel opts into `flex: 1` via a `pds-tabpanel--stretch-active` host class and fills via `block-size: 100%`. No `::slotted`/attribute-reflection churn, so existing snapshots are untouched.

Both are **additive and off by default** — zero visual change for existing consumers.

**Motivation / context:** surfaced by the Clubs app-frame spike ([kajabi-products#55087](https://github.com/Kajabi/kajabi-products/pull/55087)), where the group Chat-tab composer couldn't pin to the panel bottom (the fill-height trap) and the tab strip needed an underline — both hand-worked via `::part()` overrides.

## Scope note — DSS-221 (tab-label padding) intentionally NOT included
The third tabs gap ([DSS-221](https://linear.app/kajabi/issue/DSS-221), tokenise tab-label padding) is **carved out on purpose**: the current default underline padding is a raw `11px` and the spike actually wanted a different padding *model* (padded buttons vs. today's `gap`-spaced tab list). Either change shifts **every tab in every consuming app**, so it needs design sign-off and belongs in its own design-gated PR rather than riding along here. No new dependencies.

## Type of change

- [x] This change requires a documentation update (auto-generated `readme.md` regenerated via build)

# How Has This Been Tested?

- [x] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other: Storybook stories added (`Divider`, `Stretch`); visual/design QA still to come

Added spec coverage: `pds-tabs` gets the `pds-tabs--divider`/`pds-tabs--stretch` classes when the props are set; `pds-tabpanel` gets `pds-tabpanel--stretch-active` only when `stretch` **and** `selected`. Full core suite green locally (`nx run @pine-ds/core:test --skip-nx-cache` — 90 suites / 2920 tests). Build regenerates `readme.md` + `components.d.ts`; lint passes (0 errors).

**Follow-up before ready-for-review:** design QA of the divider colour/weight and the stretch behaviour in a real constrained-height layout.

**Test Configuration**:
- Pine versions: `@pine-ds/core@3.27.0` (this branch)
- Draft — not yet browser/AT-verified.

# Checklist:

- [x] I have added tests that prove my feature works

---

> [!NOTE]
> **Low Risk**
> Additive UI props default to false with targeted layout CSS and unit tests; no changes to tab selection or ARIA behavior when props are unset.
> 
> **Overview**
> Adds two opt-in, default-off props on **`pds-tabs`**: **`divider`** draws a subtle rule under the tab list, and **`stretch`** turns the host into a flex column so the active **`pds-tabpanel`** can consume remaining height inside a bounded parent (replacing common `::part` / manual height hacks).
> 
> **`stretch`** is propagated to child panels; only the selected panel gets **`pds-tabpanel--stretch-active`** and the matching flex/height styles. **`components.d.ts`**, readme, MDX examples, Storybook stories, and unit specs cover the new API and class behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bae11b5252ed87ef9e6e982c5b18ef20f8ff4834. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/core/src/components.d.ts
+++ b/libs/core/src/components.d.ts
@@ -2194,6 +2194,10 @@ export namespace Components {
           * @default false
          */
         "selected": boolean;
+        /**
+          * @default false
+         */
+        "stretch": boolean;
         "variant": string;
     }
     interface PdsTabs {
@@ -2206,6 +2210,16 @@ export namespace Components {
           * A unique identifier used for the underlying component `id` attribute.
          */
         "componentId": string;
+        /**
+          * Adds a divider rule beneath the tab list.
+          * @defaultValue false
+         */
+        "divider"?: boolean;
+        /**
+          * Stretches the component so the active tab panel fills the remaining height. Requires the component to have a constrained height (e.g. a flex parent).
+          * @defaultValue false
+         */
+        "stretch"?: boolean;
         /**
           * Sets the aria-label attached to the tablist element
          */
@@ -5488,6 +5502,10 @@ declare namespace LocalJSX {
           * @default false
          */
         "selected"?: boolean;
+        /**
+          * @default false
+         */
+        "stretch"?: boolean;
         "variant"?: string;
     }
     interface PdsTabs {
@@ -5500,6 +5518,16 @@ declare namespace LocalJSX {
           * A unique identifier used for the underlying component `id` attribute.
          */
         "componentId": string;
+        /**
+          * Adds a divider rule beneath the tab list.
+          * @defaultValue false
+         */
+        "divider"?: boolean;
+        /**
+          * Stretches the component so the active tab panel fills the remaining height. Requires the component to have a constrained height (e.g. a flex parent).
+          * @defaultValue false
+         */
+        "stretch"?: boolean;
         /**
           * Sets the aria-label attached to the tablist element
          */
@@ -6054,6 +6082,14 @@ declare namespace LocalJSX {
         "variant": 'inline' | 'plain';
         "fontSize": 'sm' | 'md' | 'lg';
         "href": string;
+        "turbo": string;
+        "turboFrame": string;
+        "turboAction": 'advance' | 'replace';
+        "turboMethod": 'get' | 'post' | 'put' | 'patch' | 'delete';
+        "turboConfirm": string;
+        "turboStream": string;
+        "turboPrefetch": string;
+        "turboPreload": string;
     }
     interface PdsLoaderAttributes {
         "isLoading": boolean;
@@ -6256,11 +6292,14 @@ declare namespace LocalJSX {
         "parentComponentId": string;
         "variant": string;
         "selected": boolean;
+        "stretch": boolean;
     }
     interface PdsTabsAttributes {
         "tablistLabel": string;
         "componentId": string;
         "variant": 'primary' | 'availability' | 'filter' | 'pill';
+        "divider": boolean;
+        "stretch": boolean;
         "activeTabName": string;
         "activeTabIndex": number;
     }
--- a/libs/core/src/components/pds-tabs/docs/pds-tabs.mdx
+++ b/libs/core/src/components/pds-tabs/docs/pds-tabs.mdx
@@ -233,6 +233,84 @@ Individual tabs can be disabled using the `disabled` prop. Disabled tabs cannot 
   </pds-tabs>
 </DocCanvas>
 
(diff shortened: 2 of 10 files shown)


QuintonJason:
## 🖼️ Visual test finding — runtime `TypeError` on the two new stories

Surfaced while dry-running a new local Storybook visual-test harness against this PR, then confirmed with a before/after baseline against `main`.

### What happens
The **Divider** and **Stretch** stories throw a runtime error on every render (component still renders, so it's non-fatal, but it fires each `componentWillRender`):

```
TypeError: Cannot read properties of undefined (reading 'toString')
  at propGeneration        (pds-tabs)
  at passPropsToChildren
  at componentWillRender
```

### Baseline — this is introduced by the PR, scoped to the new stories
Clean per-navigation console reads (favicon/HMR noise excluded):

| Story | `main` | this PR |
|---|---|---|
| `components-tabs--default` (shared) | ✅ 0 errors | ✅ 0 errors |
| `components-tabs--divider` (new) | — | ❌ 1 error |
| `components-tabs--stretch` (new) | — | ❌ 1 error |

The shared Default render is unaffected on both branches, so this isn't a pre-existing issue — it's specific to the two new stories.

### Root cause
`propGeneration()` calls `.toString()` on `variant` with no guard:

```ts
// libs/core/src/components/pds-tabs/pds-tabs.tsx:124-125
child.parentComponentId = this.componentId.toString();
child.variant = this.variant.toString();   // ← throws when variant is undefined
```

`variant` is declared required (`@Prop() variant!`) with no default. The existing stories all render through `BaseTemplate`, which passes `variant=${args.variant}`, so it's never undefined. The **new** `Divider` (L66) and `Stretch` (L78) stories are hand-written inline and **omit `variant`** → `this.variant` is `undefined` → `.toString()` throws.

### Suggested fix (either/both)
1. **Story** (quick): add `variant="primary"` to the `<pds-tabs>` in the Divider and Stretch stories, matching the others.
2. **Component** (recommended, hardening): guard `propGeneration` — e.g. `this.variant?.toString()` / default to `'primary'`. `classNames()` already guards `if (this.variant && …)`, so the component otherwise tolerates an absent `variant`; only `propGeneration` assumes it's set. Any consumer using `<pds-tabs>` without `variant` will hit this same throw.

_Non-blocking from a rendering standpoint, but worth clearing before merge since it fires on every render and would surface in any consumer that omits `variant`._

pixelflips (commented):
Code looks good, but can we add an example of these to the tabs docs page so the MCP will be able to pick it up as well?

QuintonJason:
Addressed the Cursor Bugbot finding + rebased:

- **Stories omit required `variant`** — added `variant="primary"` to the new `Divider` and `Stretch` stories (fixes the runtime `TypeError` from the missing required prop).
- Rebased onto latest `main` (was ~5 commits behind); `components.d.ts` is now consistent with post-i18n `main`. Both tab specs green.

pixelflips (approved):
Still missing the docs page additions (docs/pds-tabs.mdx), otherwise LGTM! 👍🏼

QuintonJason:
Added the docs-page examples in bae11b52 (`docs/pds-tabs.mdx`): a **Divider** section (rule beneath the tab list) and a **Stretch** section (fill-height panel inside a fixed-height container, with a note that the parent needs a bounded height). Both render live `DocCanvas` blocks with React + web-component source, so the MCP picks them up alongside the existing variant examples. Thanks!

PR #786, opened 2026-08-05
feat(i18n): prop-based localization pass-through (pilot)

QuintonJason:
# Description

Pine ships user-facing strings baked into compiled components — mostly a11y text a screen reader announces (`pds-alert` → `aria-label="Dismiss alert"`). Because the string lives inside the component, a consumer **cannot translate it**.

This makes those strings **props with English defaults**, so the consumer (KP) owns all translation via its own i18n YAML. The design system stays fully string-agnostic — it stores no translations.

**Draft / pilot** — opened for the DS-team naming + versioning call below, not for immediate merge.

> **Reworked from the initial approach** per the 2026-08-05 DS sync: the first cut added a `PineI18n` runtime registry (locale catalog + `set`/`setLocale`); the team chose the simpler **prop-only** model. The registry is gone — just props + a tiny formatting helper for the few dynamic strings.

## What's here

- **Static a11y strings → props with English defaults:** `dismissLabel` (alert, toast), `triggerLabel` (avatar), `searchOptionsLabel` / `selectedItemsLabel` (multiselect); `placeholder` / `searchPlaceholder` keep their English defaults. Consumers pass translated strings.
- **Dynamic strings → template props + a stateless helper** (`libs/core/src/i18n/i18n-format.ts`, no stored strings):
  - `formatMessage(template, vars)` — `{token}` interpolation (textarea `characterCountLabel`, multiselect `itemRemovedLabel`).
  - `pluralize(el, count, forms)` — picks the CLDR form via the platform **`Intl.PluralRules`** (locale read from the DOM `lang`), no ICU dependency (multiselect `selectedCountLabelOne` / `selectedCountLabelOther`).
- No stored strings, no locale registry, **no new runtime dependencies.** New `libs/core/src/i18n/` (the helper) and `libs/core/lint-rules/` (enforcement) directories.

## Type of change

Additive — new optional props (English-defaulted) + a small helper module. Rendered output is unchanged (defaults reproduce today's English), so every pre-existing snapshot still passes.

# How Has This Been Tested?

- [x] unit tests
- [x] accessibility tests
- [x] tested manually

- `stencil build --docs` passes; `components.d.ts` + readmes regenerated.
- **Full suite 2653 / 2653 green**, incl. default + translated prop tests per component and an `i18n-format` unit spec covering interpolation and **Polish `one`/`few`/`many`** pluralization + `{current}/{max}` interpolation.
- Baked-string audit (`npm run i18n.audit`, two rules — JSX attributes + component logic): the **pilot components are clean (0)**; see the honesty note in review round 2.

## Review follow-up (`b7d8304c`)

- **Full CLDR plural reach.** `pds-multiselect` gains `selectedCountLabels` — a plural forms map (object prop *or* JSON-string attribute, keyed by CLDR category) so locales past English's `one`/`other` (Polish `few`/`many`, Arabic `zero`/`two`) can localize the selected-count text. `one`/`other` stay as the ergonomic English defaults; the map layers over them and malformed JSON degrades to those defaults.
- **`pluralize` never returns empty** — falls through `category → other → one → bare {count}`.

## Review round 2 (addressing @pixelflips)

- **Lint rule blind spot fixed.** `no-hardcoded-a11y-string` now recurses into `ConditionalExpression` / `LogicalExpression` operands, so a literal hidden in a branch (`{cond ? 'Label' : undefined}`, `{this.label || 'Options'}`) is caught. This immediately surfaced strings the old rule missed on the flagship — now wired as props.
- **`pds-multiselect` genuinely complete.** New props for every remaining string: `optionsLabel` (listbox), `noOptionsFoundLabel`, `createOptionLabel` + `createOptionAriaLabel`, and `pillRemoveLabel`. The last is threaded into a new **`pds-chip` `dismissLabel`** prop (default `'Remove'`) so pill-mode remove buttons are localizable too.
- **Trigger accessible name.** Reverted the empty-`placeholder` behavior — `getTriggerText()` falls back to a non-empty string so the trigger is never nameless (WCAG 4.1.2), even when `placeholder` is `""`/`undefined` (the latter reaches this path unintentionally via the React wrapper's unconditional prop assignment). Covered by a `placeholder=""` test.
- **Honest audit number.** With the sharper rule, the pilot components report **0**, and the audit now correctly surfaces **1 remaining** string in **`pds-combobox`** (`this.label || 'Options'`) — a non-pilot component. Left as the first rollout-backlog item rather than pulled into this PR; the point is the tool no longer reports clean while strings sit in an audited file.
- **Tooling hygiene.** Renamed the script `lint.i18n` → **`i18n.audit`** so it no longer auto-joins `run-p lint.*` (was running on every contributor lint / CI). Header comments rewritten to describe current behavior (dropped `P3`/"CI-gate once drained"/cross-repo `sage-lint` references; fixed a stale `charCountLabel` → `characterCountLabel`). Noted the runner's ESLint-8 programmatic-API dependency (breaks on an ESLint 9 upgrade).
- **JSDoc consistency.** All translatable-string props now share one phrasing — "Pass a translated string to localize it;" + either "defaults to English." or "`{token}` is interpolated." — so the generated readmes read consistently.
- **Helper limitations documented.** `i18n-format.ts` now spells out that DOM-`lang` resolution is (1) not reactive to a client-side `lang` switch and (2) does not cross shadow boundaries — both arguments for a consumer passing a pre-formatted string when those cases matter.

## Open questions for the team

1. **Prop / string-id naming scheme.** Worth settling before rollout. Concrete input: fully localizing *one* component with prop-per-string took **~11 props on `pds-multiselect`** (and the rule can't see visible children or object-valued attrs like `pds-chip`'s label without bespoke handling). That proliferation is real data for weighing prop-per-string vs. a `(count) => string` formatter / slot, and vs. one plural **forms-map** prop (`selectedCountLabels`) vs. per-category props.
2. **Required a11y props.** Should aria props be `required` (Stencil enforcement) rather than English-defaulted? Precedent exists both ways in-repo — `pds-tabs` `tablistLabel!: string` and `pds-radio-group` `groupLabel` are required with no default; this PR deliberately English-defaults so nothing renders blank / no consumer break. Leaning to record the English-default rationale in the ADR rather than leave open — feedback welcome.
3. Approve the new `i18n/` + `lint-rules/` directories and a **minor version bump** (additive).

Full decision record + the prop-only reconciliation in the initiative ADR / team-sync note.

---

> [!NOTE]
> **Medium Risk**
> Touches accessibility-critical copy across several public components and expands the multiselect API substantially; behavior should stay backward-compatible via defaults, but consumers and wrappers must adopt many new optional props for full localization.
> 
> **Overview**
> Introduces a **prop-based localization model** so consumers can override baked-in English UI and a11y strings without Pine storing translations. A small `libs/core/src/i18n/` helper adds `{token}` interpolation (`formatMessage`) and DOM-`lang`-aware CLDR plural selection (`pluralize` via `Intl.PluralRules`).
> 
> **Pilot components** replace hard-coded strings with English-default props: dismiss labels on alert/toast/chip, avatar `triggerLabel`, textarea `characterCountLabel`, and a large set on **`pds-multiselect`** (placeholders, search/listbox labels, empty state, create-option copy, removal announcements, pill remove labels, and selected-count forms including optional `selectedCountLabels` JSON/object for non-English plural categories).
> 
> **Tooling:** custom ESLint rules (`no-hardcoded-a11y-string`, `no-hardcoded-ui-string-in-logic`) plus `npm run i18n.audit` to scan component `.tsx` sources and report remaining baked strings (optional `--error` for CI). Rule logic now catches literals inside conditionals/logical expressions.
> 
> Defaults preserve current English output; new props and exported `PluralForms` are additive. Specs cover default vs translated props and helper edge cases (e.g. Polish plurals, malformed JSON fallback, non-empty multiselect trigger name when `placeholder` is empty).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21998c4ef9c7e9cee305e99a3f40e9717dc26531. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/core/lint-rules/no-hardcoded-a11y-string.cjs
```diff
@@ -0,0 +1,132 @@
...
+      const e = valueNode.expression;
+      if (e.type !== 'TemplateLiteral' || e.expressions.length === 0) return null;
+      // Preserve the original static text (no injected separators) and look for a
+      // whitespace-delimited word. Sentences have space-bounded words ("of",
+      // "characters"); id fragments are hyphen/underscore-joined ("-listbox") and
+      // are left alone.
+      const staticText = e.quasis.map((q) => q.value.cooked).join('');
+      if (!/(^|\s)[A-Za-z]{2,}(\s|$)/.test(staticText)) return null;
+      return e.quasis.map((q) => q.value.cooked).join('${…}');
+    };
```
pixelflips: The rule has a blind spot that makes the burndown number unreliable.

`literalString()` and `interpolatedWithWords()` both only unwrap a `JSXExpressionContainer` whose `.expression` is *directly* a `Literal` or `TemplateLiteral`. A `ConditionalExpression` or `LogicalExpression` wrapping either is never inspected, so it passes silently.

Running the shipped audit on this branch:

```
scanned 49 component source files - 0 finding(s) across 0 component(s)
  no hard-coded a11y strings - inventory clean.
```

That is not accurate. These are still in `pds-multiselect.tsx`, the component this PR pilots:

[L1201](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/libs/core/src/components/pds-multiselect/pds-multiselect.tsx#L1200-L1202) and [L1284](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/libs/core/src/components/pds-multiselect/pds-multiselect.tsx#L1283-L1285):

```tsx
aria-label={isCreateOption ? `Create new tag: ${option.text}` : undefined}
aria-label={this.label || 'Options'}
```

(The visible strings at [L1209](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/libs/core/src/components/pds-multiselect/pds-multiselect.tsx#L1208-L1210) and [L1305](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/libs/core/src/components/pds-multiselect/pds-multiselect.tsx#L1304-L1306) are caught by neither rule.)

The PR description's "9 findings on main -> 0" and the commit's "i18n burndown still 0" both rest on this metric. Given the stated plan to flip this to a CI gate, a gate that reports clean while four strings sit in the file it just audited is worse than no gate. Recommend recursing into `ConditionalExpression` / `LogicalExpression` operands before the burndown number is used as evidence of completeness.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -53,15 +55,60 @@ export class PdsMultiselect {
...
+   * Accessible label for the search input. Pass a translated string to localize.
+   * @defaultValue 'Search options'
+   */
+  @Prop() searchOptionsLabel = 'Search options';
+
+  /**
+   * Accessible label for the selected-items region. Pass a translated string to localize.
+   * @defaultValue 'Selected items'
+   */
+  @Prop() selectedItemsLabel = 'Selected items';
```
pixelflips, line 79: `searchOptionsLabel` and `selectedItemsLabel` have no test coverage — grep for either prop name or its attribute form (`search-options-label` / `selected-items-label`) in `test/pds-multiselect.spec.tsx` returns zero hits. The new `describe('i18n (selected-count plural forms)')` block covers only `selectedCountLabelOne/Other/Labels`.

[CLAUDE.md](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/CLAUDE.md): "Every new prop, event, or behavior needs corresponding test coverage."

Worth noting this is inconsistent within the PR itself — `pds-alert`, `pds-avatar`, `pds-toast` and `pds-textarea` each got a matching default + translated-value pair. There is also direct precedent: #519 drew the same review comment ("there are no stories or tests covering it... Add unit tests to verify [the prop reaches the rendered aria-label]").

The `selectedCountLabels` *object* form is also untested — only the JSON-string attribute path has a spec, and the object form is what framework consumers will use.

pixelflips (commented):
Review pass on the pilot. Nine inline comments below, roughly in severity order. Most are scoped to the enforcement tooling and to `pds-multiselect` rather than to the pass-through model itself.

One finding has no file to anchor to: commit `b7d8304c` uses a 19-line body and the scope `multiselect`. [CLAUDE.md](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/CLAUDE.md) asks for single-line messages and a `pds-` prefixed scope matching the component name (`feat(pds-multiselect):`).

Separately, on open question 2 (required vs defaulted a11y props): there is existing precedent either way — [`pds-tabs.tsx#L21-L23`](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/libs/core/src/components/pds-tabs/pds-tabs.tsx#L21-L23) uses `@Prop() tablistLabel!: string` and [`pds-radio-group.tsx#L64-L66`](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/libs/core/src/components/pds-radio-group/pds-radio-group.tsx#L64-L66) uses a required `groupLabel`, both with no English default. English defaults here are a deliberate divergence worth recording in the ADR rather than leaving open.

libs/core/lint-rules/i18n-audit.cjs
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env node
+'use strict';
+
+/**
+ * P3 i18n enforcement audit — burndown runner.
+ *
+ * Runs pine-i18n/no-hardcoded-a11y-string over every real component .tsx
+ * (excluding test/docs/stories) and prints the burndown grouped by component.
+ * Audit-only: exits 0 by default. Pass --error to exit non-zero when any
+ * finding remains (the CI-gate mode, once the backlog is drained).
```
pixelflips: These comments carry rollout-plan context that will rot: `P3` is an internal priority/phase label, and "the CI-gate mode, once the backlog is drained" describes a future state rather than what the code does today. A reader six months out has no way to resolve either.

Describing the current behavior is enough here — the flag exists and exits non-zero; why it is not yet wired as a gate belongs in the PR description or the ADR.

libs/core/lint-rules/no-hardcoded-a11y-string.cjs
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * Flags hard-coded user-facing text on Pine components — the attributes that a
+ * screen reader announces or a user reads but that a consumer cannot translate
+ * unless the value is a prop (English-defaulted) they can override. A string
+ * LITERAL on one of the target attributes is a finding; a JSX expression (a prop
+ * ref, or the i18n helpers — e.g. `{this.dismissLabel}` /
+ * `{formatMessage(this.charCountLabel, vars)}`) is assumed localizable and passes.
+ *
+ * Audit-first (mirrors sage-lint): run as a warning + burndown, flip to error in
+ * CI once the backlog is drained. Purely syntactic — no type info required.
```
pixelflips: Same rollout-plan context as the header in `i18n-audit.cjs`: "flip to error in CI once the backlog is drained" is a plan, not behavior.

"mirrors sage-lint" additionally points at another repo — a reader without `sage-lib` context cannot resolve it or check whether it is still true.

Minor, nearby: the usage example at L10-11 cites `this.charCountLabel`, but the prop this PR actually adds is `characterCountLabel` (`pds-textarea.tsx`).

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -1369,9 +1416,34 @@ export class PdsMultiselect {
   private getTriggerText(): string {
     const count = this.selectedItems.length;
     if (count === 0 || (this.selectedDisplay === 'pill' && this.pillPosition === 'below')) {
-      return this.placeholder || 'Select...';
+      // An explicit empty `placeholder` is honored (renders no text); the prop
+      // defaults to 'Select...' so the un-configured case is unchanged.
+      return this.placeholder;
```
pixelflips: Dropping the `|| 'Select...'` guard here (and at L1365) can leave the trigger with no accessible name at all.

The trigger `<button>` has no `aria-label` — its accessible name comes solely from the `.pds-multiselect__trigger-text` span plus an `aria-hidden` icon. When `getTriggerText()` returns an empty string, that span is empty and the button is nameless (WCAG 2.1 SC 4.1.2).

The description frames this as affecting only a consumer who deliberately blanks the placeholder, but `undefined` reaches the same path without intent. [`attachProps.ts#L30-L32`](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/libs/react/src/components/react-component-lib/attachProps.ts#L30-L32) assigns unconditionally:

```ts
(node as any)[name] = newProps[name];
```

So `<PdsMultiselect placeholder={maybeUndefined} />` overwrites the Stencil field initializer, and the default never comes back. The generated types won't warn either — `Components.PdsMultiselect["placeholder"]` narrowed from `string | undefined` to `string` in this PR.

I verified the three contributing facts by reading (guard removed, no `aria-label` on the trigger, unconditional assignment in the wrapper) but did not confirm the rendered output at runtime, so worth a quick check before deciding. Either keep a fallback in `getTriggerText()` or give the trigger an explicit `aria-label` so its name never depends on placeholder text. There is currently no test for `placeholder=""` either way.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -53,15 +55,60 @@ export class PdsMultiselect {
...
-   * @default 'Find...'
+   * Placeholder text for the search input inside the dropdown panel. Pass a
+   * translated string to localize.
+   * @defaultValue 'Find...'
    */
-  @Prop() searchPlaceholder: string = 'Find...';
+  @Prop() searchPlaceholder = 'Find...';
+
+  /**
+   * Accessible label for the search input. Pass a translated string to localize.
```
pixelflips: JSDoc phrasing for the new translatable-string props diverges four ways across the PR, for props that all do the same thing:

- `pds-alert` / `pds-toast` `dismissLabel`, `pds-avatar` `triggerLabel`: "Pass a translated string to localize it; defaults to English."
- this line and `searchPlaceholder` / `selectedItemsLabel` / `placeholder`: "Pass a translated string to localize."
- `pds-textarea` `characterCountLabel`: a third phrasing
- `itemRemovedLabel`, `selectedCountLabelOne` / `Other`: no localize clause at all

[CLAUDE.md](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/CLAUDE.md): "JSDoc descriptions on @Prop, @Event, and @Method must use consistent terminology across components ... match the wording and phrasing of the existing description."

These strings land verbatim in the generated readme tables, so consumers get four explanations of one mechanism. Since this PR is establishing the pattern the rest of the rollout will copy, settling one sentence now is worth it.

Also in this file: the props this PR touches moved to `@defaultValue`, but untouched neighbors (`selectedDisplay` L193, `pillPosition` L201, `maxInlinePills` L208) still use `@default`. `@defaultValue` is the dominant convention repo-wide, so the direction is right — just leaves this file mixed.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
```diff
@@ -1325,7 +1372,7 @@ export class PdsMultiselect {
     return (
       <div
         class="pds-multiselect__pill-list pds-multiselect__pill-list--inline"
-        aria-label="Selected items"
+        aria-label={this.selectedItemsLabel}
```
pixelflips, line 1397: This region label is now translatable, but the `pds-chip` elements rendered just below it are not: [`pds-chip.tsx#L160-L162`](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/libs/core/src/components/pds-chip/pds-chip.tsx#L160-L162) hardcodes `aria-label: 'Remove'` with no override prop.

So in pill mode every selected item still has an untranslatable remove button. Pre-existing rather than introduced here, but it means the pilot does not actually complete `pds-multiselect`, which is the component the pilot is meant to prove out. The new lint rule cannot see it either — it is an object property, not a JSX attribute, so neither rule's visitor reaches it.

This and the ternary blind spot above are the same shape of problem: prop-per-string only works if every string can be enumerated, and both the manual pass and the tool missed some on the flagship component. Worth weighing when deciding the naming scheme in open question 1 — a `(count: number) => string` formatter or a slot has nothing to enumerate and needs no linter to police it.

libs/core/package.json
```diff
@@ -62,6 +62,7 @@
     "lint": "run-p lint.*",
     "lint.eslint": "eslint src",
     "lint.styles": "stylelint \"./src/**/*.*css\"",
+    "lint.i18n": "node lint-rules/i18n-audit.cjs",
```
pixelflips: `"lint": "run-p lint.*"` on L62 globs the `lint.` prefix, so this script auto-joins the shared lint target — it now runs on every `npm run lint`, `npx nx run @pine-ds/core:lint`, and in the CI `test-lint` job, with no separate opt-in.

It exits 0 without `--error`, so nothing fails today; the effect is burndown output on every contributor's lint run. [CLAUDE.md](https://github.com/Kajabi/pine/blob/b7d8304c1a01875c1d39a2dad89cc27e7373570b/CLAUDE.md) lists modifying build config under "Ask first," and open question 3 does raise the directories, so mostly flagging that the wiring is implicit rather than explicit. Naming it `i18n.audit` would keep it opt-in until the gate decision is made.

Separately, `i18n-audit.cjs` L32-40 uses `linter.defineParser()`, `linter.defineRule()`, and eslintrc-shaped `verify()` config. Those work on the pinned `eslint@^8.25.0` but were all removed in ESLint 9, so this script will break on that upgrade.

libs/core/src/i18n/i18n-format.ts
```diff
@@ -0,0 +1,52 @@
...
+  );
+}
+
+/** Nearest `[lang]` ancestor, then `<html lang>`, then `'en'`. */
+export function resolveLocale(el?: Element | null): string {
+  const fromEl =
+    el && typeof el.closest === 'function' ? el.closest('[lang]')?.getAttribute('lang') : null;
+  const fromDoc = typeof document !== 'undefined' ? document.documentElement.lang : '';
+  return fromEl || fromDoc || 'en';
+}
```
pixelflips, line 45: Two limitations worth documenting on this helper, since consumers will reasonably assume `lang` is authoritative.

**Not reactive.** `resolveLocale()` is called during render via `pluralize()`, but Stencil has no reactive dependency on a DOM attribute. A client-side locale switch (`document.documentElement.lang = 'pl'`) will not re-render mounted components — plural forms stay stale until unrelated state forces a render. Fine if consumers full-page-reload on locale change; a silent bug if they do not.

**Cannot cross shadow boundaries.** `el.closest('[lang]')` stops at the shadow root. A Pine component rendered inside another custom element's shadow template, where `lang` sits on that ancestor's host element, falls through to `document.documentElement.lang` and silently renders the wrong plural category. That is the nesting pattern Pine is built for, so it is worth at least a doc callout.

Both are arguments for letting the consumer supply the formatted string rather than having the component infer locale from the DOM.

libs/core/lint-rules/no-hardcoded-a11y-string.cjs
QuintonJason: Fixed in `794cec13`. `no-hardcoded-a11y-string` now recurses into `ConditionalExpression` and `LogicalExpression` operands, so a literal in any branch is reached — RuleTester cases added for both `{cond ? '…' : undefined}` and `{this.label || 'Options'}`.

Running the sharpened rule immediately surfaced exactly the strings you called out. They're now props: `createOptionAriaLabel` (the ternary at L1201) and `optionsLabel` (the `|| 'Options'` at L1284), plus the two visible strings you flagged parenthetically — `createOptionLabel` ("Add …") and `noOptionsFoundLabel`. So the pilot component now reports a genuine 0.

Notably the rule then caught one more the manual pass had missed entirely: the same `this.label || 'Options'` pattern in **`pds-combobox`** (L1590). I've left that one unwired and let the burndown honestly report `1 finding in pds-combobox` rather than pull a non-pilot component into this PR — the point being the tool no longer reports clean while a string sits in a file it audited. It's the first real rollout-backlog item.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
QuintonJason: Added in `794cec13`. New tests in the `i18n` describe block cover `searchOptionsLabel` (default + translated) and `selectedItemsLabel` (default + translated, in pill mode), plus the other newly-added props (`optionsLabel`, `noOptionsFoundLabel`, `createOptionLabel`/`createOptionAriaLabel`, `pillRemoveLabel`) and a `placeholder=""` accessible-name test.

The `selectedCountLabels` **object** form is now tested too — a dedicated case sets the prop as an object (the framework-consumer path) alongside the existing JSON-string-attribute case. Full suite 2653/2653.

libs/core/lint-rules/i18n-audit.cjs
QuintonJason: Rewritten in `794cec13` to describe only current behavior — dropped the `P3` label and the "CI-gate once the backlog is drained" future-state phrasing. Kept the two usage lines and added a note that the runner relies on the ESLint 8 programmatic API (see the reply on your package.json comment).

libs/core/lint-rules/no-hardcoded-a11y-string.cjs
QuintonJason: Fixed in `794cec13`. The header no longer describes the rollout plan or points at `sage-lint`; it now just states what the rule does and that conditional/logical branches are searched. The stale `charCountLabel` reference is corrected to `characterCountLabel`.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
QuintonJason: Good catch — accessibility wins over the niche "blank the placeholder" case, and you're right that `undefined` reaches this path unintentionally via the wrapper's unconditional assignment. Reverted in `794cec13`: `getTriggerText()` (and the inline-pill placeholder span) fall back to a non-empty string, so the trigger's accessible name is never empty. Added a `placeholder=""` test asserting the trigger text stays non-empty.

I kept it as a last-resort fallback rather than giving the trigger a dedicated `aria-label`, to stay minimal — but I agree the deeper fix is that a placeholder shouldn't be the accessible name at all. A dedicated trigger label prop is a reasonable follow-up if we want to sever that dependency entirely.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
QuintonJason: Unified in `794cec13`. Every translatable-string prop across the PR now uses one phrasing: "Pass a translated string to localize it;" followed by either "defaults to English." (static) or "`{token}` is interpolated." (template). That collapses the four variants into one, so the generated readme tables read consistently — which matters since this PR sets the pattern the rollout copies.

On `@default` vs `@defaultValue`: the props this PR touches all use `@defaultValue` now; the untouched neighbors (`selectedDisplay`, `pillPosition`, `maxInlinePills`) still read `@default`. Since normalizing those isn't part of this change, I left them — happy to sweep the file in a separate `docs(pds-multiselect):` commit if you'd prefer it consistent now.

libs/core/src/components/pds-multiselect/pds-multiselect.tsx
QuintonJason: Addressed in `794cec13`. Added a `dismissLabel` prop to `pds-chip` (default `'Remove'`, used in both the button and link close paths), and `pds-multiselect` now threads it through as `pillRemoveLabel` on every pill. So pill mode is genuinely localizable now, and the flagship is actually complete.

Your broader point stands and I've written it into open question 1 as concrete data: fully localizing this one component with prop-per-string took ~11 props, and the rule still can't see visible children or object-valued attributes (the chip's label had to be wired by hand, not caught). That's exactly the enumeration problem — real evidence for weighing a `(count) => string` formatter / slot against prop-per-string when we settle the naming scheme.

libs/core/package.json
QuintonJason: Renamed in `794cec13`: `lint.i18n` → `i18n.audit`, so it's out of the `run-p lint.*` glob and no longer runs on every `npm run lint` / CI `test-lint` — it's opt-in until the gate decision is made.

On the ESLint 9 point — correct, `Linter#defineParser`/`defineRule` and the eslintrc-shaped `verify()` config were all removed in ESLint 9. I've added a note to the runner header flagging that it must move to flat config + `verify()` without the `define*` helpers on that upgrade. Left the implementation on the ESLint 8 API since core is still pinned there.

libs/core/src/i18n/i18n-format.ts
QuintonJason: Documented both in `794cec13` on the `i18n-format.ts` module header:

1. **Not reactive** — `resolveLocale` is read at render and Stencil has no reactive dependency on a DOM attribute, so a client-side `lang` switch won't re-render mounted components; the note tells consumers to pass the pre-formatted string if they switch locale client-side.
2. **Doesn't cross shadow boundaries** — `el.closest('[lang]')` stops at the host's shadow root, so a Pine component nested in another custom element's shadow template falls back to `document.documentElement.lang`; documented, with the same "pass a pre-formatted string" escape hatch.

Both reinforce your underlying point, which I've surfaced for open question 1: inferring locale from the DOM has real limits, and letting the consumer supply the formatted string sidesteps them.

QuintonJason:
Thanks @pixelflips — genuinely useful pass. All nine inline threads are addressed in `794cec13`; highlights: the a11y rule now recurses conditional/logical branches (which surfaced and drove real fixes), `pds-multiselect` is fully localized including a new `pds-chip` `dismissLabel` for pill removal, the empty-`placeholder` a11y regression is reverted, the audit script is renamed out of the `lint.*` glob, JSDoc is unified, and the helper's DOM-`lang` limitations are documented. Full suite 2653/2653; pilot components audit clean, with the sharpened rule honestly reporting 1 remaining string in `pds-combobox` as the first rollout-backlog item.

Two points from the review body:

- **Commit convention.** Fair — the previous commit used a multi-line body and a bare `multiselect` scope. This one is a single-line `feat(pds-multiselect):`. The two earlier commits on the branch (`69fa189c`, `b7d8304c`) still have long bodies; since this pilot will squash-merge, I'll author the final squash message to CLAUDE.md's single-line + `pds-`-scope convention rather than force-push history on a draft. Say the word if you'd rather I clean the branch history now.

- **Required vs. defaulted a11y props (open Q2).** Thanks for the `pds-tabs` / `pds-radio-group` precedent — that's the strongest input yet. I've moved this from "open" toward "decide + record": my recommendation is to keep the English default here (so an un-migrated consumer never ships a blank/nameless control and nothing breaks on upgrade), and document that rationale in the ADR alongside the required-prop precedent, rather than leave it dangling. I've noted both in the PR description and will write it into the ADR. If the team prefers `required` for the pure-aria props specifically, that's a clean follow-up — it's an additive-to-breaking change, so it wants its own call.

pixelflips (approved):
Thanks for addressing the feedback! LGTM! 👍🏼

libs/core/src/i18n/i18n-format.ts
```diff
@@ -0,0 +1,68 @@
...
+): string {
+  let category: Intl.LDMLPluralRule = 'other';
+  try {
+    category = new Intl.PluralRules(resolveLocale(el)).select(count);
+  } catch {
+    /* invalid locale tag → fall back to 'other' */
+  }
+  // Prefer the selected category, then the required `other`, then `one`; never
+  // return empty — a bare `{count}` still shows the number rather than nothing.
+  const template = forms[category] ?? forms.other ?? forms.one ?? '{count}';
```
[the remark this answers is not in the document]
QuintonJason: Good catch — fixed in `21998c4e`. You're right that `??` only falls through on `null`/`undefined`, so an empty-string form (`{ other: '' }`) was treated as present and returned, blanking the trigger. Switched the fallback chain to `||` so empty strings are skipped too:

```ts
const template = forms[category] || forms.other || forms.one || '{count}';
```

Values here are always strings (`Partial<Record<Intl.LDMLPluralRule, string>>`), so `||` is safe — a non-empty form wins, an empty one falls through, and a bare `{count}` is the last resort. Added a spec covering the empty-string-form case (`{ one: '', other: '' }` → `'1'`, and `{ one: '', other: '{count} items' }` → `'1 items'`).

PR #737, opened 2026-05-15
ci: add Chromatic workflow and clarify release cadence

QuintonJason:
# Description

**Two related changes:**

1. **Wire up Chromatic visual regression in CI.** The `@chromatic-com/storybook` addon is already configured in `libs/core/.storybook/main.js`, but no CI workflow ran against it — so visual regressions could ship undetected. This adds `.github/workflows/chromatic.yml` that builds doc-components → Stencil → Storybook, then publishes to Chromatic on PR and push to `main`/`next`. Fork PRs are skipped because repository secrets are unavailable to them.

2. **Document the actual release cadence.** `CONTRIBUTING.md` now explains how Visual regression and Releases work, and the comment block at the top of `schedule-release.yml` is expanded to make clear that releases are **manual / on-demand** via `workflow_dispatch` (and `workflow_call`) — there is no weekly cron yet. Contributors were sometimes asking when the next release would auto-cut; this answers that.

**Required maintainer action**: add a **`CHROMATIC_PROJECT_TOKEN`** repo secret (from the Pine Chromatic project) before merging. Without it the new workflow will fail at the publish step.

Part of the design-system maturity push — closes the Level-4 "Visual regression testing" gap.

Fixes #(no-issue)

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

- Confirmed `@chromatic-com/storybook` is already installed at `libs/core/.storybook/main.js`.
- Verified the workflow only requires a single new secret (`CHROMATIC_PROJECT_TOKEN`).
- Verified the fork-PR guard expression matches the official GitHub Actions pattern.

This PR cannot be fully exercised until the maintainer adds the secret; the first run after merge will surface any environmental issues.

- [ ] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [x] tested manually (workflow syntax review)
- [x] other: pending live CI run after secret is added

**Test Configuration**:

- Pine versions: 3.25.1 (current main)
- OS: macOS 25.3.0
- Browsers: n/a
- Screen readers: n/a
- Misc: GitHub Actions

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing tests pass locally with my changes
[...]

PR #788, opened 2026-08-10
docs(skills): add pine-visual-test-pr visual regression skill

QuintonJason:
## What

Adds a `pine-visual-test-pr` Claude Code skill at `.claude/skills/pine-visual-test-pr/SKILL.md` — a self-iterating, Playwright-MCP visual-test harness for Pine PRs.

It's modeled on kajabi-products' `kp-visual-test-pr`, re-pointed at Pine's actual visual surface (Storybook) instead of Rails worktree servers.

## How it works

1. **Preflight** — `gh` auth, root deps, `libs/core` scripts, Playwright MCP reachable.
2. **PR scope** — reduces the diff to changed `libs/core/src/components/pds-*` dirs.
3. **Storybook up** — `build.stencil` → `start.storybook` (:6006) → enumerate every story from `/index.json` (the Pine analog of `worktree-server list`; never guesses story IDs).
4. **Capture plan** — maps changed components → story IDs via each story file's `title:`, then builds a shot matrix.
5. **Capture + evaluate loop** — screenshots the chrome-free `iframe.html?id=…&globals=theme:…` across **light/dark themes**, viewports, and interactive states (hover/focus/open/disabled/error); self-evaluates each against embedded rules; retries harness failures (max 3), records genuine theme/focus/console/token problems as **findings**.
6. **Report** — posts a markdown report as a PR comment.

Optional **before/after baseline** mode (checkout base → rebuild Stencil → re-shoot the same matrix) labels each delta *intended* vs *regression*, gated behind a clean-tree confirmation.

## Design notes

- **Dark theme is mandatory** — dark-mode/token regressions are Pine's most common and easiest-to-miss visual bug, so light+dark is always captured.
- **Findings ≠ blockers** — a broken theme, missing focus ring, or console error is the *point* of the run and is recorded as a finding, not retried away.
- **Complements Chromatic** — positions itself as the fast, local, pre-CI pass; Chromatic (`@chromatic-com/storybook`) remains the authoritative pixel-diff gate in CI.

## Scope

Docs/tooling only — one new markdown skill file. No component, build, or runtime code touched.

---

> [!NOTE]
> **Low Risk**
> Single new markdown skill file with no changes to components, builds, or production runtime.
> 
> **Overview**
> Adds a new Claude Code skill at `.claude/skills/pine-visual-test-pr/SKILL.md` that documents an end-to-end **Playwright MCP** visual regression workflow for Pine PRs (adapted from kajabi-products’ `kp-visual-test-pr`, but aimed at Storybook instead of app servers).
> 
> The skill defines a **seven-phase runbook**: environment preflight, PR/component scoping from `gh pr diff`, Stencil build + Storybook on `:6006` with story IDs from `/index.json`, a capture matrix (**light/dark**, viewports, hover/focus/open states), optional **base-branch baseline** screenshots (with explicit git/Storybook cleanup around `createComponent.tsx` and port 6006), rule-based evaluation with retries vs **findings**, and a **deduped** `gh pr comment` report under `.claude/visual-test-reports/`.
> 
> It positions the harness as Pine’s current primary **local pixel check** until Chromatic CI merges, and links it to `pine-design-review` and `pine-run-gauntlet`. **Docs/tooling only** — no component, build, or runtime code changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddf7b5da5970ae4390db53b2a012a46d1f5ce116. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/skills/pine-visual-test-pr/SKILL.md
+++ b/.claude/skills/pine-visual-test-pr/SKILL.md
@@ -1,0 +1,417 @@
+---
+name: pine-visual-test-pr
+description: Self-iterating visual test harness for Pine PRs. Builds Stencil + serves Storybook, maps changed pds-* components to their stories, captures Playwright screenshots across light/dark themes, viewports, and interactive states, self-evaluates each against the embedded rules, rejects and retries failures, and posts a markdown report as a PR comment. Optional before/after baseline diff for modified components.
+argument-hint: "[PR number — auto-detects from current branch if omitted]"
+---
+
+# Visual Test PR (Pine)
+
+Automated visual testing for Pine design-system pull requests. Pine ships
+leaf-level web components consumed by every Kajabi app, so a visual
+regression here propagates everywhere — and the most common one is a
+dark-mode / token break that never shows in a light-theme review.
+
+This skill builds Stencil, serves Storybook, discovers the stories for the
+components the PR touched, captures Playwright screenshots across **theme
+(light/dark)**, viewport, and interactive state, evaluates each against the
+rules below, and loops until every shot passes or a structural blocker
+requires human input. It complements — does not replace — Chromatic: this is
+the fast, local, agent-evaluated pass you run before pushing, and before
+Chromatic runs in CI.
+
+## When to Use
+
+- After changing any `libs/core/src/components/pds-*/**` file (`.tsx`, `.scss`/`.tokens.scss`, `.stories.*`)
+- As a complement to `pine-design-review` (token/a11y code review) and `pine-run-gauntlet`
+- Before requesting human review, to pre-validate rendered output in both themes
+- To sanity-check a change locally without waiting for the Chromatic CI job
+
+**Not the right tool when:** the PR only touches docs (`*.mdx`), build config, or non-component TypeScript with no rendered surface — there is nothing to screenshot. Say so and stop.
+
+## Prerequisites
+
+- Playwright MCP configured in this Claude Code session
+- Node deps installed at the Pine repo root (`node_modules/` present)
+- `gh` CLI authenticated
+- Able to run `libs/core` scripts (`build.stencil`, `start.storybook`)
+
+## Progress Tracking
+
+At the start of Phase 0, create tasks for all phases:
+
+```
+TaskCreate: "Phase 0: Environment preflight"
+TaskCreate: "Phase 1: PR resolution + component scope"
+TaskCreate: "Phase 2: Storybook up + story index"
+TaskCreate: "Phase 3: Capture plan"
+TaskCreate: "Phase 4: Baseline decision"
+TaskCreate: "Phase 5: Screenshot capture + evaluation loop"
+TaskCreate: "Phase 6: Report"
+```
+
+Mark each `in_progress` when started, `completed` when done.
+
+### Pivoting Away (Early Exit)
+
+If the skill is abandoned mid-run (user redirects, structural blocker, or the PR has no visual surface), clean up the task list before stopping:
+
+1. Mark the current `in_progress` task `completed` (or `deleted` if no meaningful work was done)
+2. Mark all remaining `pending` tasks `deleted`
+
+Do this immediately when a pivot is decided — before switching to any other work — so stale `in_progress` tasks don't persist across sessions.
+
+---
+
+## Phase 0 — Environment Preflight
+
+Run all checks from the Pine repo root before touching anything. If **any** check fails, print the exact fix command and stop — do not continue to Phase 1.
+
+```bash
+# 1. gh CLI authenticated
+gh auth status
+
+# 2. Repo root has deps installed
+ls node_modules >/dev/null && echo "root deps OK"
+
+# 3. libs/core scripts present
+node -e "const s=require('./libs/core/package.json').scripts; ['build.stencil','start.storybook'].forEach(k=>{if(!s[k]){console.error('missing script '+k);process.exit(1)}}); console.log('core scripts OK')"
+
+# 4. Node version matches
+cat .nvmrc
+```
+
+Also verify Playwright MCP is reachable by calling `mcp__playwright__browser_snapshot` — if it throws a tool-not-found or connection error, the MCP is down.
+
+| Check | Failure Symptom | Fix |
+|---|---|---|
+| `gh auth status` | "not logged in" | `gh auth login` |
+| `node_modules` | "No such file or directory" | `npm install` at repo root |
+| core scripts | "missing script" | Confirm you're in the Pine repo; scripts live in `libs/core/package.json` |
+| Playwright MCP | Tool error / connection refused | Re-open Claude Code — MCP connection dropped |
+
+**Do not proceed until all checks pass.**
+
+---
+
+## Phase 1 — PR Resolution + Component Scope
+
+1. **If PR number argument provided:** use it.
+2. **Otherwise:** auto-detect from current branch:
+   ```bash
+   gh pr view --json number,url,title,headRefName,baseRefName,body
+   ```
+3. **Get changed files:**
+   ```bash
+   gh pr diff --name-only
+   ```
+4. **Determine component scope.** Reduce the diff to the set of changed components:
+   ```bash
+   gh pr diff --name-only | grep -oE 'libs/core/src/components/[^/]+' | sort -u
+   ```
+   Each entry is a component directory (e.g. `pds-button`, `pds-alert`). This is the scope — the skill only screenshots stories for these components (plus their child components, e.g. `pds-tabs` → `pds-tab`, `pds-tabpanel`).
+5. **Classify each changed file** to predict the visual risk (drives the capture plan):
+   | Changed file | Visual risk | Emphasis |
+   |---|---|---|
+   | `*.tokens.scss` / `*.scss` | color, spacing, dark-mode | **both themes**, all variants |
+   | `*.tsx` | markup/DOM, slots, states | interactive states, variant coverage |
+   | `*.stories.*` | the story itself changed | capture the new/changed stories |
+   | shared token/global SCSS | broad | widen scope — may affect many components |
+
+**If no PR found:** stop and tell the user to create one first.
+**If the diff touches no `libs/core/src/components/**` files** and no shared token/global styles: there is no component visual surface — say so and stop (this PR is out of scope for the skill).
+
+---
+
+## Phase 2 — Storybook Up + Story Index
+
+Pine's visual surface is Storybook. Stories render compiled components from `dist/`, so Stencil must be built **before** Storybook starts. **Never guess story IDs — always read them from `/index.json`.**
+
+1. **Build Stencil** (produces/refreshes `libs/core/dist/`):
+   ```bash
+   ( cd libs/core && npm run build.stencil )
+   ```
+2. **Start Storybook** on port 6006 as a background process (it stays up for the whole run):
+   ```bash
+   ( cd libs/core && npm run start.storybook )   # run_in_background: true
+   ```
+3. **Health check** — poll until the story index is served (up to ~60s; Storybook's first Vite optimize pass is slow):
+   ```bash
+   curl -s -o /dev/null -w "%{http_code}" http://localhost:6006/index.json
+   ```
+   Retry until `200`. If it never comes up, read the background process output for the failure (common: port 6006 already taken → the existing server is fine to reuse; or a Stencil build error → fix and rebuild).
+4. **Fetch the story index** and keep it — this is the authoritative story list (the Pine analog of `worktree-server list`):
+   ```bash
+   curl -s http://localhost:6006/index.json > /tmp/pine-stories.json
+   # entries is a map keyed by story id; keep only real stories (not docs)
+   node -e "const e=require('/tmp/pine-stories.json').entries; Object.values(e).filter(x=>x.type==='story').forEach(x=>console.log(x.id+'\t'+x.title+'\t'+x.name))"
+   ```
+
+Store `http://localhost:6006` as `$SB` for all subsequent phases.
+
+### Resolving changed components → story IDs
+
+Story IDs are `slug(title)` + `--` + `slug(exportName)` (e.g. title `components/Button`, story `Primary` → `components-button--primary`). The directory name (`pds-button`) is **not** the title — so map through the story file, don't assume:
+
+1. For each changed component dir, read its story file's `title:` field:
+   ```bash
+   grep -h "title:" libs/core/src/components/<component>/**/stories/*.stories.* 
+   ```
+2. Match that title against the `title` column of the index dump from step 4. Every matching entry's `id` is a story to capture.
+
+---
+
+## Phase 3 — Capture Plan
+
+Build the explicit shot list **before** opening a browser, and output it so the user can see coverage.
+
+### Axes
+
+For every in-scope story, the capture matrix is:
+
+- **Theme (required, both):** `light` and `dark`. This is Pine's highest-value axis — dark-mode token regressions are the most common Pine visual bug. Never skip dark.
+- **Viewport:** `desktop` (1280×800) for every story. Add `mobile` (375×812) only for components with responsive layout behavior (e.g. `pds-nav`, `pds-modal`, `pds-tabs` overflow, `pds-filters`, `pds-table`). Document which components you flagged responsive and why.
+- **Interactive states** (desktop only, only for the states the component actually has, and only when the changed files could affect them):
+  - `hover`, `focus` (keyboard focus ring), `active`/`pressed`
+  - `disabled` (usually already a dedicated story — prefer the story over synthesizing)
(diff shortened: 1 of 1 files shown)


.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+regression here propagates everywhere — and the most common one is a
+dark-mode / token break that never shows in a light-theme review.
+
+This skill builds Stencil, serves Storybook, discovers the stories for the
+components the PR touched, captures Playwright screenshots across **theme
+(light/dark)**, viewport, and interactive state, evaluates each against the
+rules below, and loops until every shot passes or a structural blocker
+requires human input. It complements — does not replace — Chromatic: this is
+the fast, local, agent-evaluated pass you run before pushing, and before
+Chromatic runs in CI.
```
pixelflips, line 20: **Chromatic is described as an existing CI gate, but it isn't merged.**

This and several other spots assert Chromatic in the present tense as the authoritative CI pixel-diff (also L27, L238, L406, L432, L439). But `.github/workflows/chromatic.yml` does not exist on `main` — it only exists on #737, which is still open. What is merged is `@chromatic-com/storybook` as a local Storybook dev panel (`libs/core/.storybook/main.js`), not a CI check.

As of today this skill is the *only* pixel-level regression check Pine has, which changes how a reader should weigh a clean result. Suggest hedging these to "once Chromatic CI lands" / "when configured".

pixelflips (commented):
Code review — 9 inline findings, see comments on the relevant lines.

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+## Phase 1 — PR Resolution + Component Scope
+
+1. **If PR number argument provided:** use it.
+2. **Otherwise:** auto-detect from current branch:
+   ```bash
+   gh pr view --json number,url,title,headRefName,baseRefName,body
+   ```
+3. **Get changed files:**
+   ```bash
+   gh pr diff --name-only
```
pixelflips, line 105: **An explicitly-passed PR number never reaches `gh pr diff`.**

L98 says "If PR number argument provided: use it", and the frontmatter `argument-hint` advertises passing one. But this command and the one on L109 invoke bare `gh pr diff --name-only` with no number and no placeholder — and bare `gh pr diff` resolves against the *current branch's* PR.

Invoking the skill with an explicit PR number while checked out on a different branch (the exact scenario the argument-hint promotes) silently computes the changed-file list and component scope for the wrong PR. Thread `<number>` through these calls.

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+
+- **Theme (required, both):** `light` and `dark`. This is Pine's highest-value axis — dark-mode token regressions are the most common Pine visual bug. Never skip dark.
+- **Viewport:** `desktop` (1280×800) for every story. Add `mobile` (375×812) only for components with responsive layout behavior (e.g. `pds-nav`, `pds-modal`, `pds-tabs` overflow, `pds-filters`, `pds-table`). Document which components you flagged responsive and why.
+- **Interactive states** (desktop only, only for the states the component actually has, and only when the changed files could affect them):
+  - `hover`, `focus` (keyboard focus ring), `active`/`pressed`
+  - `disabled` (usually already a dedicated story — prefer the story over synthesizing)
+  - `error`/`invalid` for form components (`pds-input`, `pds-select`, `pds-textarea`, `pds-radio`, `pds-checkbox`, `pds-combobox`, `pds-multiselect`)
+  - `open`/expanded for overlay components (`pds-modal`, `pds-dropdown-menu`, `pds-tooltip`, `pds-accordion`, `pds-combobox`)
+- **Direction (optional):** `rtl` via `globals=direction:rtl`, only if the PR touches logical-property / layout SCSS or the component is directional.
```
pixelflips, line 185: **The RTL axis is planned here but never executed or graded.**

Phase 5's loop enumerates only `storyId × theme × viewport × state` (L246) and its navigate URL template parameterizes only `theme:<theme>` (L250) — direction is absent. The Visual Test Rules table (L291-302) has no RTL/mirroring rule either, so even a captured RTL shot would have nothing to evaluate it against.

Worth noting the global is genuinely supported (`.storybook/preview.js` defines the `direction` globalType and a `withDirection` decorator), so this is a wiring gap rather than an unsupported feature. Either plumb direction through Phase 5 and add a rule, or drop the axis.

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+mcp__playwright__browser_hover({ selector: "<interactive part>" })
+mcp__playwright__browser_take_screenshot({ filename: "<component>__<story>__desktop__<theme>__hover.png" })
+
+# keyboard focus (focus ring is a frequent dark-mode regression)
+mcp__playwright__browser_press_key({ key: "Tab" })
+mcp__playwright__browser_take_screenshot({ filename: "<component>__<story>__desktop__<theme>__focus.png" })
+
+# open/expanded overlays
+mcp__playwright__browser_click({ selector: "<trigger>" })
+mcp__playwright__browser_take_screenshot({ filename: "<component>__<story>__desktop__<theme>__open.png" })
```
pixelflips, line 278: **These capture calls violate the absolute-path rule this file sets at L201-213.**

L203 is explicit: "A relative `filename` silently lands the PNG outside the repo. Always pass an **absolute** path", with the correct form shown at L210 as `"<repo-abs-path>/.claude/visual-test-reports/pr-<number>/<name>.png"`. All four capture calls here (L263, L270, L274, L278) pass bare relative names instead.

This is the exact failure `1e58d359` ("harden pine-visual-test-pr from dry run against #785") was written to fix — but the fix only landed in the explanatory section. `git show 9660213a` confirms these four lines are byte-identical to the pre-fix version. The same bare-name pattern also persists at L327 and in the Phase 6 report template.

An agent copies the nearest concrete command, not the distant prose rule, so this reproduces the original bug.

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+| 9 — naming | Rename the file — no re-capture needed |
+
+**Max 3 retries per item.** After 3 failures on the same item, classify it as a **structural blocker**, record the reason, and continue. Never stop the loop. Distinguish a *harness* blocker (couldn't capture) from a *real finding* (captured fine, but the render is wrong) — real findings are the point of the skill and belong in the report as issues, not blockers.
+
+### Phase 5b — Baseline capture (if elected in Phase 4)
+
+After all PR-branch shots are approved:
+
+1. Confirm a clean tree; record the current branch.
+2. **`build.stencil` dirties tracked generated files** — `libs/core/src/components.d.ts` and `libs/react/src/components/react-component-lib/createComponent.tsx` are checked in but regenerated on every build, so a branch switch **aborts** ("would be overwritten") after you've built. Reset them before each `git checkout`: `git checkout -- libs/core/src/components.d.ts libs/react/src/components/react-component-lib/createComponent.tsx`. (These are generated, not authored — safe to discard. If a genuine unrelated WIP file is also dirty, `git stash` it and restore after.)
```
pixelflips, line 325: **`createComponent.tsx` is not "generated, not authored — safe to discard".**

It carries a hand-authored double-registration guard (added in `5678d227`) that is not present in the stock `@stencil/react-output-target` template — diffing the repo copy against `node_modules/@stencil/react-output-target/react-component-lib/createComponent.tsx` shows that guard is the only delta. `npm run build.stencil` strips it back to the vendor version.

So Phase 5b step 5's "rebuild Stencil to restore the PR state" (L328) does **not** restore it — following this flow to completion leaves the tree with an authored fix silently reverted, and the skill never surfaces that to the user.

This also contradicts the skill's own error-handling row at L421: "Dirty tree at branch switch | Ask before `git checkout`; never discard the user's work."

`libs/core/src/components.d.ts` is genuinely generated; suggest splitting the two and warning about the `createComponent.tsx` guard explicitly.

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+3. `git checkout <baseRef>` → `( cd libs/core && npm run build.stencil )` → **restart** Storybook (don't reuse the head-branch server — its story index is stale and every open tab will throw `NoStoryMatchError` against the new bundle).
```
pixelflips, line 326: **"restart" has no implementation, and contradicts Phase 2.**

Nothing anywhere in this file stops the process on port 6006 — no `kill`, `lsof`, `pkill`, or KillShell step exists. So "restart Storybook" is not executable as written: starting a second server on 6006 either fails with address-in-use or falls through to Phase 2's rule at L141 ("port 6006 already taken → the existing server is fine to reuse"), which is precisely what this line forbids.

Either outcome silently breaks the baseline comparison, since screenshots get captured against the wrong branch's bundle — the `NoStoryMatchError` this line warns about. Needs an explicit stop step, and Phase 2's blanket "fine to reuse" should carve out the post-branch-switch case.

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+- **Never** loop indefinitely — track attempts per item.
+
+---
+
+## Phase 6 — Report
+
+Generate a markdown report and post it as a PR comment.
+
+```bash
+mkdir -p .claude/visual-test-reports
```
pixelflips, line 344: **The report file is never written, but is read back at L387.**

This creates only the *directory*. The report template at L349-383 is an illustrative fenced block, and no step — prose or command — ever writes the filled-in report to `.claude/visual-test-reports/pr-<number>-visual-test.md`.

L387 then does `cat` on that path, so taken literally every Phase 6 run fails with "No such file or directory" and no comment is posted. Needs an explicit Write step before the post.

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+### <Item>
+- **Rule:** <n> — **Attempts:** 3 — **Observed:** <…> — **Needs:** <human action>
+
+## Rules Applied
+- [x] Component renders  - [x] Assets loaded  - [x] Both themes  - [x] No console errors
+- [x] Focus visible  - [x] Viewport  - [x] No clipping  - [x] Token discipline
+- [x] Naming  - [x] Variant coverage
+```
+
+Post it:
```
pixelflips, line 385: **No once-per-PR dedup guard on posting.**

Posting here is unconditional, so every run adds another full report comment. This skill is explicitly built to be re-run (self-iterating retry loop in Phase 5, baseline mode in Phase 5b), so repeat posts are the normal case, not the edge case.

This was already flagged on #746 for the sibling skill ("Repeated gauntlet re-posts reviews"), which is why `pine-run-gauntlet/SKILL.md` L246-249 now carries an explicit guard: "Before posting, check existing PR reviews/comments for a body containing `Posted by pine-run-gauntlet`. If one exists, **do not post again** unless the user explicitly asks." Same guard is warranted here.

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,439 @@
...
+```bash
+gh pr comment <number> --body "$(cat .claude/visual-test-reports/pr-<number>-visual-test.md)"
```
pixelflips, line 387: **Double-quoted `--body` will break on report content, and `--repo` is missing.**

`--body "$(cat …)"` runs command substitution inside double quotes, so backticks in the report body get re-evaluated by the shell — and visual-test reports routinely contain backticks and quotes (captured console error text, CSS token names, story IDs). Without `--repo Kajabi/pine` this also targets the wrong repo when not run from a pine checkout.

Both were flagged by review on #746 ("Fallback comment breaks body" / "Fallback comment omits repo flag"), and `pine-run-gauntlet/SKILL.md` L318-320 was fixed to say: "use a temp file so markdown quotes/backticks do not break the shell; always pass `--repo Kajabi/pine`" and "do not pass the markdown body as a double-quoted `--body` string."

Suggest `mktemp` + `--body-file` + `--repo Kajabi/pine`.

QuintonJason:
Thanks for the thorough review — all findings addressed in e83a33dd.

**pixelflips (9):**
1. **Chromatic present-tense CI gate** — hedged everywhere (intro, when-to-use, Phase 4, Completion, anti-patterns, related-skills). Now states plainly that only the local `@chromatic-com/storybook` panel is merged and this skill is currently Pine's *only* pixel-level check until the CI job lands.
2. **PR number never reaches `gh pr diff`** — Phase 1 now resolves an explicit `<number>` and threads it through every `gh pr diff`/`view` call, with a note on the wrong-branch trap.
3. **RTL axis half-wired** — dropped it (removed from the axes and the URL template; report template no longer lists `+rtl`), with a note on how to add it as a full axis later.
4. **Capture commands violate the absolute-path rule** — all Phase 5 capture calls (Steps 3 & 4) and the interactive-state calls now use the absolute `<REPORT_DIR>` path; report template updated too.
5. **`createComponent.tsx` is not "safe to discard"** — corrected: it carries the hand-authored double-registration guard; `git checkout --` is kept *because* it restores the committed (guarded) copy, and Phase 5b step 5 now re-runs it **after** the final rebuild (which re-strips the guard), with a `git status` clean-check. Added an anti-pattern against `git add`-ing the stripped version.
6. **"restart" had no implementation** — Phase 5b now has an explicit `lsof -ti:6006 | xargs kill` stop step before re-starting; Phase 2's "reuse the existing server" carves out the post-branch-switch case in both the prose and the error-handling table.
7. **Report file never written but `cat`-read** — Phase 6 step 1 now writes the filled report to `$REPORT` via the Write tool before posting.
8. **No dedup guard on posting** — added the once-per-PR guard (grep for the `Posted by pine-visual-test-pr` marker, skip unless asked), matching the `pine-run-gauntlet` pattern; the marker is now in the report template footer.
9. **Double-quoted `--body` + missing `--repo`** — posting now uses `gh pr comment <number> --repo Kajabi/pine --body-file "$REPORT"`.

**Cursor:**
- *Console check spans whole session (`all: true`)* — already removed in a prior commit; the read is per-navigation.
- *Per-navigation read misses the earlier theme* — fixed: the console read now runs **after each theme's navigation** (inside the light/dark loop in Step 3, and again after interactions in Step 4), so both themes are judged rather than only the last one.

_(This reply posted via `--body-file` — dogfooding fix #9.)_

QuintonJason:
Both re-review findings fixed in the latest commit:

- **Cross-branch PR screenshots (High)** — good catch; threading the number scoped the right files but capture still renders the checked-out tree. Phase 1 now has an explicit branch-match guard (`git branch --show-current` vs the PR's `headRefName`) that **stops on mismatch** and tells the user to check out the head branch first — it does not auto-switch.
- **Stash can restore stripped guard (Medium)** — reordered Phase 5b: reset the two build-generated files (`git checkout --`) **before** stashing any genuine WIP, so the stash can never snapshot the vendor-stripped `createComponent.tsx` and reintroduce it on `pop`.

QuintonJason:
Fixed in the latest commit. The clean check now applies to **skill-induced** changes only (the two generated files matching committed), and runs **before** restoring the step-3 stash. Restoring the user's WIP is done last and legitimately re-dirties the tree — that's expected and no longer treated as a failed clean check.

pixelflips (approved):
Thanks for addressing all the feedback. LGTM! 👍🏼

.claude/skills/pine-visual-test-pr/SKILL.md
```diff
@@ -0,0 +1,472 @@
...
+---
+
+## Phase 1 — PR Resolution + Component Scope
+
+1. **Resolve the PR number into `<number>` and thread it through every `gh` call below.** A bare `gh pr diff` / `gh pr view` resolves against the *current branch's* PR — so if the skill is invoked with an explicit number while checked out on a different branch (exactly what the `argument-hint` promotes), an un-numbered call silently scopes the wrong PR. Always pass `<number>` explicitly.
+   - **If a PR number argument was provided:** use it as `<number>`.
+   - **Otherwise:** auto-detect from the current branch and read the number back out:
+     ```bash
+     gh pr view --json number,url,title,headRefName,baseRefName,body   # <number> = .number
+     ```
```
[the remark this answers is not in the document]
QuintonJason: Resolved on the current head (ddf7b5d) — this was reviewed against the earlier e83a33d, before the branch-match guard landed. Phase 1 **step 2** (lines 107–112) now compares `git branch --show-current` to the PR's `headRefName` and **stops on mismatch** ("do not auto-switch — branch changes are the user's call"), so capture can no longer render the local checkout while the diff/report target a different PR. Step 1 also captures `headRefName`/`baseRefName` via `gh pr view --json`, so those template vars are set. The skill deliberately guards-and-stops rather than auto-checking-out the PR head.

PR #620, opened 2026-01-05
style: update alert dark mode link hover style

QuintonJason:
# Description

- [x] in dark mode links on items with a colored background are hard to read

Fixes #(issue)

| Before | After |
|--------|--------|
|alertLinkHoverBefore|alertLinkHoverAfter|

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Click the dark mode toggle -> Visit the alert page -> Small with actions -> hover link

- [x] tested manually

**Test Configuration**:

- Pine versions:
- OS:
- Browsers:
- Screen readers:
- Misc:

# Checklist:

If not applicable, leave options unchecked.

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing tests pass locally with my changes
- [ ] Design has QA'ed and approved this PR

PR #791, opened 2026-08-12
feat(pds-link): add active prop for nav-tab styling

QuintonJason:
# Description

**Draft / proposal — not for merge as-is.** Concrete code for the DS team to react to on DSS-227. The API shape (a boolean `active` prop vs. a `nav`/`tab` variant, and naming) is the open decision.

Adds an `active` prop to `pds-link` so a per-URL **navigation tab** can express its current state without every consumer re-deriving the same tokens.

**Why:** `pds-tabs` is an in-page panel switcher (requires `pds-tabpanel` children), so real per-URL tab strips are built as a semantic `<nav>` of `pds-link`s. Today the active/inactive treatment is hand-tokened in each consumer — and the obvious reach, `color="secondary"`, maps to `--pine-color-text-*inverse*-secondary` (an on-dark-surface preset), so it renders white-on-light. See DSS-227.

**This change:**
- `@Prop() active = false` — applies the active treatment (persistent underline + `--pine-color-text-strong`) and sets `aria-current="page"` on the inner anchor.
- `.pds-link--active` scss rule.
- Spec coverage (active adds class + `aria-current`; default omits both) — **pds-link.spec passes locally**.
- A `NavTabs` story showing an active/inactive strip.

**Not included / open:**
- A muted/neutral color option for the *inactive* tabs (currently `variant="plain"`), and/or a docs callout renaming the `secondary` on-dark preset — flagged in DSS-227, deferred pending API direction.

## ⚠️ `components.d.ts` intentionally not regenerated here

My local Stencil regen drops `OneOf` + the `attr:`/`prop:` mapped types across **all** components (a local-toolchain artifact, not a real change), so committing it would regress unrelated types. The source of truth is the `@Prop() active` in `pds-link.tsx`; **please let CI / a maintainer regenerate `components.d.ts` with the canonical toolchain.** (The per-component `readme.md` — which regenerates cleanly in isolation — is included and shows the new prop.)

## Type of change

# How Has This Been Tested?

- [x] unit tests — added active-prop specs; `pds-link.spec.tsx` passes locally.
- [x] tested manually — `NavTabs` story renders the active underline + strong text; `aria-current="page"` present on the active anchor.

**Test Configuration**:

- Pine versions: main @ 4f6f765d
- Misc: `@stencil/core` 4.38.3

# Checklist:

- [x] I have added tests that prove my feature works
- [ ] New and existing tests pass locally with my changes — pds-link passes; unrelated pre-existing spec failures on main are untouched

---

> [!NOTE]
> **Low Risk**
> Additive, default-off API on a single component with styling and accessibility only; no auth, data, or breaking changes.
> 
> **Overview**
> Adds an **`active`** boolean to **`pds-link`** so per-URL nav tab strips (a `<nav>` of links, not `pds-tabs`) can mark the current destination without hand-rolled tokens or misusing `color="secondary"`.
> 
> When **`active`** is set, the host reflects the attribute for **`:host([active])`** styles: persistent underline and strong text that beat **`color`** presets while still honoring context link CSS variables. The inner anchor gets **`aria-current="page"`**.
> 
> Also updates generated **`components.d.ts`**, component **readme**, a **`NavTabs`** Storybook example, and unit tests for **`aria-current`** and reflected **`active`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a632f3fccc3f0e04258224ca36a8e112387e1332. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/core/src/components/pds-link/pds-link.scss
```diff
@@ -88,3 +88,15 @@
...
+// text so the current page reads without a color prop; pairs with aria-current.
+.pds-link--active {
+  color: var(--pine-color-text-strong);
+  text-decoration: underline;
+
+  &:hover {
+    color: var(--pine-color-text-strong);
+    text-decoration: underline;
+  }
+}
```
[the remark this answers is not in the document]
QuintonJason: Fixed. `active` now reflects to the host and the treatment is scoped to `:host([active]) .pds-link` (and `:host([active]) .pds-link:hover`), which outranks the `:host([color=...]) .pds-link` presets. So the current item always gets the strong-text treatment and never falls back to the on-dark `secondary` hue on hover, even with a color prop set. Added a spec asserting the reflected host attribute.

libs/core/src/components/pds-link/pds-link.scss
```diff
@@ -88,3 +88,19 @@
...
+// Active navigation item (tab strips / nav rails). Persistent underline + strong
+// text so the current page reads without a color prop; pairs with aria-current.
+// Active navigation item (tab strips / nav rails). Scoped to :host([active]) so
+// it outranks the `:host([color=...]) .pds-link` presets — the current item
+// always gets the strong-text treatment (and never falls back to the on-dark
+// `secondary` hue on hover), even when a color prop is also set.
+:host([active]) .pds-link {
+  color: var(--pine-color-text-strong);
+  text-decoration: underline;
+}
```
pixelflips, line 102: This hardcodes `color`, bypassing the resolution chain the base rule documents at [line 11](https://github.com/Kajabi/pine/blob/5ca7a4a6e30756a17bed662cefbf3a44665a675f/libs/core/src/components/pds-link/pds-link.scss#L10-L12): `// Priority: 1) explicit --color prop, 2) context variable, 3) default`.

Raising specificity to (0,3,0) correctly wins against the `:host([color=...])` presets, but it also discards the *inherited* context variable — a separate case from the explicit `color` prop the new comment mentions. `--pds-context-link-color` is set by `pds-alert` for all five variants ([pds-alert.scss#L1-L29](https://github.com/Kajabi/pine/blob/5ca7a4a6e30756a17bed662cefbf3a44665a675f/libs/core/src/components/pds-alert/pds-alert.scss#L1-L29)) and by `pds-dropdown-menu-item`, and `pds-alert`'s docs state:

> Links (`pds-link`) automatically inherit the correct colors from the alert variant - no manual color prop needed.

([pds-alert.mdx#L91-L93](https://github.com/Kajabi/pine/blob/5ca7a4a6e30756a17bed662cefbf3a44665a675f/libs/core/src/components/pds-alert/docs/pds-alert.mdx#L91-L93))

So `<pds-alert variant="danger"><pds-link active ...>` now renders `--pine-color-text-strong` instead of the alert's link color. That chain was introduced by `ed4a457f` ("style: update alert dark mode link hover style (#620)") as a dark-mode contrast fix, so this reopens that risk.

Consulting only the context variable preserves alert inheritance while still overriding an explicit `color` prop, which matches the stated intent:

```scss
:host([active]) .pds-link {
  color: var(--pds-context-link-color, var(--pine-color-text-strong));
}

:host([active]) .pds-link:hover {
  color: var(--pds-context-link-color-hover, var(--pine-color-text-strong));
}
```

Minor, same block: the previous comment was not removed — lines 92-97 now say "Active navigation item (tab strips / nav rails)" twice, with two different explanations.

pixelflips (commented):
Reviewed at 5ca7a4a6. The move to `:host([active]) .pds-link` plus `@Prop({ reflect: true })` resolves the specificity conflict with the `:host([color=...])` presets, and reflection is the right second half of that fix. Three inline comments below.

libs/core/src/components/pds-link/pds-link.tsx
```diff
@@ -121,6 +129,10 @@ export class PdsLink {
       classNames.push('pds-link--' + this.variant);
     }
 
+    if (this.active) {
+      classNames.push('pds-link--active');
```
pixelflips: Now that the styling lives in `:host([active]) .pds-link`, this class is emitted but nothing targets it — there is no `.pds-link--active` rule left in [pds-link.scss](https://github.com/Kajabi/pine/blob/5ca7a4a6e30756a17bed662cefbf3a44665a675f/libs/core/src/components/pds-link/pds-link.scss) at this head.

Because the component is `shadow: true`, consumers can't target it either (`part="link"` is the public hook), so it is dead code.

The bigger issue is the test: [pds-link.spec.tsx#L241](https://github.com/Kajabi/pine/blob/5ca7a4a6e30756a17bed662cefbf3a44665a675f/libs/core/src/components/pds-link/test/pds-link.spec.tsx#L240-L242) still asserts the class, so that assertion now enforces a class with no styling contract. It passes while certifying nothing, and reads as coverage for the active treatment.

Either drop the class and rely on the reflected `active` attribute assertion the test already makes on the following line, or keep the class and style it instead of `:host([active])`.

libs/core/src/components/pds-link/pds-link.tsx
```diff
@@ -150,6 +162,7 @@ export class PdsLink {
         href={this.href}
         id={this.componentId}
         part="link"
+        aria-current={this.active ? 'page' : undefined}
```
pixelflips, line 161: `aria-current` is hardcoded to `page`, which in ARIA means "the current page within a set of pages." The new `NavTabs` story links to in-page fragments — [`#chat`, `#resources`, `#members`](https://github.com/Kajabi/pine/blob/5ca7a4a6e30756a17bed662cefbf3a44665a675f/libs/core/src/components/pds-link/stories/pds-link.stories.js#L66-L70) — where the correct token is `location`.

As written the canonical example demonstrates the wrong semantic to screen-reader users. Either let the token be chosen (e.g. `active` accepting `boolean | 'page' | 'location' | 'step'`, or a companion prop), or keep `page` and change the story to per-URL hrefs so the example matches what the component asserts.

libs/core/src/components/pds-link/pds-link.scss
QuintonJason: Good catch on the inheritance case — fixed in a632f3fc. The active rule now consults the context variable first and only falls back to strong text:

```scss
:host([active]) .pds-link {
  color: var(--pds-context-link-color, var(--pine-color-text-strong));
}
:host([active]) .pds-link:hover {
  color: var(--pds-context-link-color-hover, var(--pine-color-text-strong));
}
```

So `<pds-alert variant="danger"><pds-link active>` keeps the alert's link color (no regression to `ed4a457f`), while an explicit `color` prop is still overridden — which matches the stated intent. Also removed the duplicated "Active navigation item" comment; there's a single explanation now.

libs/core/src/components/pds-link/pds-link.tsx
QuintonJason: Fixed in a632f3fc. Dropped the `pds-link--active` class entirely — the styling contract is the reflected `active` host attribute (`:host([active]) .pds-link`), which is the public, shadow-safe hook. Updated the spec to assert the reflected `active` attribute (and `aria-current`) instead of the class, so the test now certifies the actual treatment rather than a class with no rule behind it.

libs/core/src/components/pds-link/pds-link.tsx
QuintonJason: Kept `page` and aligned the story instead. The `NavTabs` example now links to per-URL destinations (`/clubs/1/chat`, `/clubs/1/resources`, `/clubs/1/members`), so `aria-current="page"` is the correct token for what's demonstrated — a tab strip navigating between distinct pages, which is the pattern this prop is for (pds-tabs already covers the in-page panel-switch case). If an in-page/`location` variant is wanted later, exposing the token can be a fast follow, but I'd rather not widen the API in this PR. Fixed in a632f3fc.

pixelflips (approved):
Thanks for addressing the feedback, LGTM! 👍🏼

PR #801, opened 2026-08-25
feat(pds-combobox): add invalid, error-message and helper-message props

pixelflips:
# Description

> **Stacked on #800**, which is stacked on #799. Base branch is `fix/select-error-announcement`, so this diff shows only the `pds-combobox` work. Merge order: **#799 → #800 → this**.

`pds-combobox` had no `invalid`, `error-message` or `helper-message`, so a consumer needing a validation state had to hand-roll all three — and the hand-rolled version could not be announced to assistive technology at all, because the trigger lives in the shadow root and the component forwards no host ARIA.

This adds the three props, matching `pds-multiselect`'s implementation as the ticket asks.

Fixes DSS-235

## What the workaround looked like

From the report, in `app/views/super_admin/dispatches/_targeting_fields.html.erb`:

```
style="--pine-border: var(--pine-border-width) solid var(--pine-color-border-danger);
       --pine-color-background-container: var(--pine-color-red-050);"
```

That override was only safe while `trigger="input"`, because `--pine-border` is also read by `.pds-combobox__button-trigger` — a precondition that lived in a comment. The report counted three separate places in `kajabi-products` hitting this, plus a React wrapper that had institutionalised rendering the error as a sibling. All of that can now be deleted in favour of the props.

## The change

**Three props**, with JSDoc matching `pds-multiselect` (the named reference) and the terse phrasing this file already uses for `disabled`:

| Prop | Type | Description |
| -- | -- | -- |
| `errorMessage` | `string?` | Error message to display. |
| `helperMessage` | `string?` | Helper message to display below the combobox. |
| `invalid` | `boolean?` | If true, the combobox is in an invalid state. |

`helperMessage` says "below the combobox" rather than multiselect's "below the input" because a combobox trigger is not always an input — it can be a button or a chip.

**Message elements** render after the trigger, with `messageId()` ids so the trigger can reference them. Helper and error are mutually exclusive, matching `pds-multiselect`.

**ARIA is wired on all three trigger variants** (`input`, `button`, `chip`), via the same helper both attributes share:

```ts
// Mirrors pds-multiselect and pds-select: an error message on its own puts the
// field in an invalid state, so assistive tech agrees with the visual treatment.
private isInvalid(): boolean {
  return Boolean(this.invalid) || Boolean(this.errorMessage);
}
```

## Scope: the danger *visual* treatment is on the input trigger only

The other two triggers already own their border and background through separate colour systems — `button-trigger--primary/accent/secondary/destructive`, and `chip-trigger--neutral/danger/success/warning/info` driven by a `--chip-border` custom property. Layering an invalid override on top of those would fight the variant system, and picking a winner is a design decision rather than a code one.

So:

| | `input` | `button` | `chip` |
| -- | -- | -- | -- |
| `aria-describedby` / `aria-invalid` | yes | yes | yes |
| helper / error message rendered | yes | yes | yes |
| danger border + background | **yes** | no | no |

`trigger="input"` is exactly the variant the report was overriding by hand, so the reported need is fully covered. For button and chip the error message still renders in danger colour with the danger icon, and chip already has `chipSentiment="danger"` for the trigger itself. **Flagging for design**: if button/chip should also take a danger treatment, that wants its own ticket with a decision on how it composes with the variant colours.

Two details worth noting in the styles:

- The invalid rule outranks the plain hover rule (`&.pds-combobox__input--invalid:hover:not(:disabled)`), so the error treatment survives a mouseover. Without that, the generic `&:hover` would repaint the border grey and the error indication would vanish on hover — which is the behaviour `pds-select` currently has.
- `--pine-color-red-050` is a fixed light red, so a `:host-context([data-theme="dark"])` override to `--pine-color-red-950` is included, following `pds-select`. `pds-multiselect` uses the same token with no dark override, which looks like a gap on its side.

## Type of change

Additive only — three optional props. No existing prop, event, method or behaviour changes, and all 2999 pre-existing tests pass untouched.

# How Has This Been Tested?

- [x] unit tests

Added 10 tests. Each asserts the attribute value **and** that `aria-describedby` resolves to a rendered element containing the expected text, so a dangling reference fails rather than passes. The three trigger variants are covered by an `it.each`.

**Verified the tests actually detect the absence.** With `pds-combobox.tsx` reverted and only the tests applied, 8 of 10 fail — including all three trigger variants, which confirms the wiring is genuinely on each one:

```
● renders the error message with an id the trigger can reference
● treats an error message on its own as an invalid state
● renders the helper message and describes it when valid
● hides the helper message while an error message is shown
● applies the invalid modifier to the input trigger
● wires the error association on the input trigger
● wires the error association on the button trigger
● wires the error association on the chip trigger

Tests: 8 failed, 3001 passed, 3009 total
```

The two that pass either way are the intended controls — `describes nothing and is not invalid without messages` and `does not apply the invalid modifier when valid`.

With the change applied:

```
Tests:       3009 passed, 3009 total
```

Also clean: `stylelint` on the SCSS, `eslint` on all changed files with **0 errors and the warning count unchanged from `main`** (25, all pre-existing `strict-boolean-conditions`).

Added `HelperMessage` and `ErrorMessage` stories, matching the error-state coverage `pds-select` and `pds-multiselect` stories already have.

**Test Configuration**:

- Pine versions: reported against `@pine-ds/core@3.29.0`; built on top of #800
- OS: macOS
- Browsers: n/a (spec layer)
- Screen readers: not verified manually — this PR verifies the DOM associations that drive announcement
- Misc: generated files intentionally included, see below

# Checklist:

---

### About the first commit — `0073aac2 chore: re-sync generated types and readme`

**This commit touches `pds-avatar/readme.md` and unrelated parts of `components.d.ts`. That is deliberate, and it is correct. Here is the evidence, so you do not have to dig for it.**

New props mean `components.d.ts` and `pds-combobox/readme.md` genuinely regenerate, so unlike the earlier PRs in this batch those changes belong in the diff. The complication is that **`main`'s committed generated artifacts are already stale**, so a build dirties files that have nothing to do with this PR.

Specifically: [#795](https://github.com/Kajabi/pine/pull/795) — `fe4a93da`, the **current HEAD of `main`** — added an `initials` shadow part to `pds-avatar.tsx` but never regenerated the readme:

```
$ git show --stat fe4a93da
 libs/core/src/components/pds-avatar/pds-avatar.tsx     |  3 ++-
 libs/core/src/components/pds-avatar/test/…spec.tsx     | 16 ++++++++++++++--
 2 files changed, 16 insertions(+), 3 deletions(-)      ← no readme.md
```

The source declares the part, so the readme has been wrong on `main` ever since:

```
$ grep -n "@part" libs/core/src/components/pds-avatar/pds-avatar.tsx
8: * @part initials - The text element that renders the avatar's initials, for restyling their color.
```

The added row is byte-identical to that JSDoc, and rebuilding after this commit produces **no further change** — it is exactly what `stencil build --docs` emits, not a hand edit.

> **One thing that trips people up:** the Shadow Parts table looks misaligned in the diff. It is not. Every row is 103 characters with cell widths `[19, 81]`; the `+`/`-` diff prefixes shift the columns by one and make it read as ragged. The `pds-avatar` props table was already complete (`status`, `statusLabel`, `statusRing` all present), so Shadow Parts was the only stale section.

**Why it is a separate commit.** Hand-separating interleaved hunks out of `components.d.ts` would mean manually editing a generated file, which the contributing guidelines forbid. Instead the two regenerations were produced by two builds:

- **`0073aac2`** — built from a *clean* tree. Contains **zero** combobox references (`git diff | grep -ci combobox` → `0`); it is purely the pre-existing drift.
- **`4c7f0750 feat(pds-combobox): …`** — built on top, so its `components.d.ts` diff is exactly the three new props and nothing else.

Reviewing those two commits separately is the easy way to read this PR. `0073aac2` also cherry-picks cleanly onto `main` if you would rather it land on its own — say so and I will split it out.

Either way, this will keep happening until CI fails when a build leaves generated output dirty. Worth its own ticket.

---

## UX/Product Change Summary

### Detected Changes

| Error Handling | Medium | `pds-combobox` gains public `errorMessage` and `invalid` props, adding a user-visible validation message below the field plus an invalid-state visual treatment (accompanying SCSS changes in the same diff). Consumers can now surface validation errors on comboboxes where previously no error affordance existed, so any form already passing these attribute names will start rendering error text and error styling. | `components.d.ts:984`, `pds-combobox.scss` |
| Data Display | Low | `pds-combobox` gains a public `helperMessage` prop that renders supporting text below the combobox, changing the component's vertical footprint when set. | `components.d.ts:1001` |

<details>

- `libs/core/src/components.d.ts:984`
- `libs/core/src/components.d.ts:1001`
- `libs/core/src/components.d.ts:1009`
- `libs/core/src/components/pds-combobox/pds-combobox.scss`

</details>

### Risk Assessment

Low-to-moderate risk: the changes are additive opt-in props on a single component, so existing consumers see no change unless they set them. Verify that helper/error text is announced via `aria-describedby`, that the invalid treatment meets contrast requirements in both light and dark themes, and that adding messages below the trigger does not shift or clip the dropdown positioning.

### Action Items

---

> [!NOTE]
> **Low Risk**
> Additive optional props on one component; behavior changes only when consumers set the new attributes, with no changes to selection, dropdown, or form submission logic.
> 
> **Overview**
> **`pds-combobox`** now supports **`invalid`**, **`error-message`**, and **`helper-message`**, aligned with **`pds-multiselect`**: helper or error text appears under the trigger (errors win over helpers), and **`error-message`** alone counts as invalid so **`aria-invalid`** and **`aria-describedby`** stay in sync for assistive tech on **input**, **button**, and **chip** triggers.
> 
> The **danger border/background** invalid styling applies only when **`trigger="input"`**; button and chip triggers still show the error row with icon but keep their variant styling. Input invalid styles include dark-theme background and hover rules so the error border does not disappear on hover.
> 
> The diff also **refreshes generated artifacts**: **`pds-avatar`** readme documents the **`initials`** shadow part (stale on main), and **`pds-tab` / `pds-tabs`** JSDoc in **`components.d.ts`** is expanded for navigation mode. Stories and spec tests cover helper/error rendering and ARIA wiring across trigger types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d119ecead7650ebb107b12a28f2244130b9090df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


QuintonJason (approved):
Nice work

PR #802, opened 2026-08-27
feat(pds-modal): add disableTopLayer to opt out of the browser top layer

QuintonJason:
# Description

`pds-modal` opens via the native `dialog.showModal()`, which promotes the dialog to the **browser top layer**. The top layer always paints above every normal-flow element regardless of `z-index`, so any overlay a consumer mounts elsewhere in the DOM (typically appended to `document.body`) renders **behind** an open `pds-modal` and is unreachable.

This is a recurring, cross-team issue — commerce file picker / media-library card / Adobe Express "Edit image", TinyMCE "Formats" dropdown, and now Clubs' image uploader (its own `z-index` modal sits behind the Pine edit modal). Patching `z-index` can't beat the top layer, moving nodes into the modal via a `MutationObserver` was abandoned (kajabi-products#51160, closed), and the only reliable workaround has been falling back to the legacy Sage modal.

This adds the escape hatch phillip proposed: a prop to render the modal outside the top layer.

**What changed:** new boolean prop **`disableTopLayer`** (default `false`):
- Default (`false`): unchanged — `dialog.showModal()` (top layer, page inert). **No behavior change for existing consumers.**
- `true`: `dialog.show()` — a non-modal dialog that stays in the normal stacking context at `--pine-z-index-modal`, so consumer overlays with a higher `z-index` can display above it.

The dimming backdrop still renders (it's the `<dialog>` element styled via `.pds-modal__backdrop`, not the `::backdrop` pseudo) and outside-click dismiss still works — this mode is for letting a higher-`z-index` overlay appear *above* the modal, not for making the modal non-blocking. What changes in this mode: the page is not made `inert`, `aria-modal` is `false`, and **the Tab focus trap is disabled** so controls in an overlay stacked above the modal stay reachable by keyboard/AT. **Escape** closes the modal while focus is inside it, but is left to an overlay above the modal when that overlay owns focus — so a keyboard user can dismiss the overlay without losing the modal underneath it. The prop is read when the modal opens (documented as open-time-only; changing it mid-open is unsupported). Caveats (documented): the dialog now participates in normal stacking (a page element with a higher stacking context can overlap it), and don't mix a `disable-top-layer` modal with a default top-layer modal in one stack (dismiss targeting is `z-index`-based).

Fixes DSS-240 (https://linear.app/kajabi/issue/DSS-240). Consumer follow-up: adopt `disable-top-layer` on the Clubs post edit modal (kajabi-products#56131).

## Type of change

# How Has This Been Tested?

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [x] other: verified end-to-end in a downstream app (kajabi-products, production build) — the previously-trapped picker displays above the modal with `disable-top-layer`.

Spec (stubbed `modalRef`): default → `showModal()`, `disableTopLayer` → `show()`, `aria-modal` reflection. E2e (real Chromium — the only place the top-layer contract and focus/Escape behavior are observable): default is `:modal` true, `disableTopLayer` is `:modal` false + open + `aria-modal="false"`, a higher-`z-index` body overlay paints above the non-modal dialog, and Escape is left to an overlay that owns focus (modal stays open) but closes the modal when focus is not held outside it. Full `@pine-ds/core` suite green (**3018 tests**). Storybook **DisableTopLayer** story + MDX example demonstrate a **focusable, self-cleaning** `document.body` overlay stacking above the modal.

_Reviewed via `pine-run-gauntlet` (code + security + design/a11y) plus Cursor Bugbot + @pixelflips. Findings addressed: Tab focus trap disabled in non-top-layer mode; Escape left to an overlay that owns focus; `aria-modal`/prop documented as open-time-only; MDX heading blank-line parse fix; demo overlay now self-removes (click / Escape / modal close) so it can't orphan over the docs page._

**Test Configuration**:

- Pine versions: `@pine-ds/core` (main)
- OS: macOS
- Browsers: Chromium

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Changes modal open semantics, focus trapping, and Escape handling for opt-in consumers; default top-layer behavior is preserved but a11y/stacking caveats apply when the flag is set.
> 
> **Overview**
> Adds **`disableTopLayer`** (default `false`) on **`pds-modal`** so consumers can open the dialog with **`dialog.show()`** instead of **`showModal()`**, keeping it in normal stacking so body-mounted overlays (file pickers, editor menus) can paint above it via **`z-index`**.
> 
> When enabled, the modal sets **`aria-modal="false"`**, skips **Tab focus trapping**, and only handles **Escape** while focus stays inside the modal (overlays above can keep Escape). Default behavior is unchanged. Docs, Storybook, generated types, and unit/e2e tests cover the new mode.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c16b5acddeb00301981a728d530c72084e82fee9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


QuintonJason:
@cursor review

libs/core/src/components/pds-modal/pds-modal.tsx
```diff
@@ -163,8 +174,15 @@ export class PdsModal {
...
-        this.modalRef.showModal();
+        // showModal() promotes the dialog to the browser top layer (and makes the
+        // rest of the page inert), which prevents any overlay outside the dialog
+        // from ever painting above it. show() opens a non-modal dialog that stays
+        // in the normal stacking context so those overlays can stack above it.
+        if (this.disableTopLayer) {
+          this.modalRef.show();
+        } else {
+          this.modalRef.showModal();
+        }
```
pixelflips, line 185: `show()` fixes paint order, but *reachability* is governed by the focus trap, which this branch leaves untouched. The result is that the body-mounted overlays this prop exists to reveal are mouse-only.

- `updateFocusableElements()` queries `this.modalRef.querySelectorAll(selector)`, so only nodes inside the dialog are ever collected — [pds-modal.tsx#L113-L115](https://github.com/Kajabi/pine/blob/a3c820d7fb6bb9c739e4287a2bd02769ffed2002/libs/core/src/components/pds-modal/pds-modal.tsx#L113-L115)
- The `Tab` branch of `handleKeyDown` has no `disableTopLayer` guard and wraps at **both** boundaries: `Tab` on the last element calls `preventDefault()` and returns focus to the first, `Shift+Tab` on the first returns to the last — [pds-modal.tsx#L304-L313](https://github.com/Kajabi/pine/blob/a3c820d7fb6bb9c739e4287a2bd02769ffed2002/libs/core/src/components/pds-modal/pds-modal.tsx#L304-L313)

Repro: open a `disable-top-layer` modal and trigger a `document.body`-appended picker. It paints above the modal as intended, but `Tab` can never leave the modal, so keyboard and screen-reader users cannot operate it.

The prop JSDoc documents the `inert` tradeoff but not this one. Either scope the trap to allow `Tab` out while `disableTopLayer` is set, or state the keyboard limitation explicitly alongside the `inert` note.

libs/core/src/components/pds-modal/pds-modal.tsx
```diff
@@ -303,7 +321,7 @@ export class PdsModal {
           'pds-modal__backdrop': true,
           'open': this.open
         }}
-        aria-modal="true"
+        aria-modal={this.disableTopLayer ? 'false' : 'true'}
```
pixelflips, line 336: `aria-modal` is recomputed on every render, but there is no `@Watch('disableTopLayer')` and the only `@Watch` is on `open` ([pds-modal.tsx#L80-L88](https://github.com/Kajabi/pine/blob/a3c820d7fb6bb9c739e4287a2bd02769ffed2002/libs/core/src/components/pds-modal/pds-modal.tsx#L80-L88)).

So flipping the prop while the modal is open re-renders this attribute without re-invoking `show()`/`showModal()`. The announced modality then contradicts the dialog's actual modality and the page's actual inertness for the rest of that open session — e.g. open with the default, set `disableTopLayer = true`, and AT is told the dialog is non-modal while the browser still holds the page inert.

Since this is an escape hatch normally set once at open time, documenting it as open-time-only would be a reasonable fix instead of adding a watcher.

pixelflips (commented):
Reviewed through a3c820d7. Four high-confidence findings left inline on the relevant lines: two in the component, two in the new docs section. The `aria-modal` prose fix in a3c820d7 resolves the earlier inaccuracy in the Technical Notes.

Lower-confidence items (no e2e coverage for the real `dialog.show()` path, the `isInnermostModal()` z-index comparison across mixed top-layer/non-modal modals, and the unrelated `pds-link` `turbo*` drift in `components.d.ts`) are not raised inline.

libs/core/src/components/pds-modal/docs/pds-modal.mdx
```diff
@@ -1318,9 +1318,141 @@ Modals can be nested, with only the topmost modal responding to escape key and b
     </pds-modal>
   </div>
 </DocCanvas>
+### Overlays Above the Modal
```
pixelflips, line 1322: Missing blank line between `</DocCanvas>` on the preceding line and this heading.

With `@mdx-js/rollup: ^3.1.1`, MDX v3 requires a blank line between a JSX flow element and a following markdown block. Without it this is not parsed as a heading and ships as the literal string `### Overlays Above the Modal` on the docs site.

It fails **silently** — `storybook build` exits 0 and the Netlify preview will look green, so CI won't catch it.

Every other heading following a `</DocCanvas>` in this file has the blank line (L388-390, L655-657, L792-794, L935-937, L1090-1092, L1451-1453). One-line fix.

libs/core/src/components/pds-modal/docs/pds-modal.mdx
```diff
@@ -1318,9 +1318,141 @@ Modals can be nested, with only the topmost modal responding to escape key and b
...
+
+By default a modal opens with the native `dialog.showModal()`, which promotes it to the browser [top layer](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer). Nothing outside the dialog can paint above the top layer &mdash; not even with a higher `z-index` &mdash; so an overlay mounted elsewhere in the DOM (a file picker, a rich-text editor menu, a third-party widget appended to `document.body`) renders _behind_ the modal and becomes unreachable.
+
+Set `disable-top-layer` to open the modal as a non-modal dialog (`dialog.show()`) in the normal stacking context instead. Overlays with a higher `z-index` can then display above it.
+
+> **Note:** In this mode the browser does not make the rest of the page `inert`. Use it only when you specifically need an external overlay to appear above the modal. Focus trapping, Escape handling, and the backdrop still work.
```
pixelflips: "Focus trapping, Escape handling, and the backdrop still work" is literally true, but reads as reassurance directly above an example that appends an overlay to `document.body` — and the focus trap continuing to work is precisely why that example is broken for keyboard users.

The trap wraps at both boundaries with no `disableTopLayer` guard ([pds-modal.tsx#L304-L313](https://github.com/Kajabi/pine/blob/a3c820d7fb6bb9c739e4287a2bd02769ffed2002/libs/core/src/components/pds-modal/pds-modal.tsx#L304-L313)) and `updateFocusableElements()` only collects nodes inside `modalRef` ([pds-modal.tsx#L113-L115](https://github.com/Kajabi/pine/blob/a3c820d7fb6bb9c739e4287a2bd02769ffed2002/libs/core/src/components/pds-modal/pds-modal.tsx#L113-L115)), so the overlay this section demonstrates is unreachable by `Tab`.

Suggest replacing the reassurance with the actual constraint: focus stays trapped inside the modal in this mode, so an external overlay is reachable by pointer only unless the consumer manages focus themselves.

Separately, the overlay in the example below is removed only by clicking it — closing the modal via Escape, the footer Close, or the header X leaves a `position: fixed; z-index: 2147483647` div orphaned over the page. Since the live block renders on this docs page, a reader can permanently park it over the documentation.

libs/core/src/components/pds-modal/pds-modal.tsx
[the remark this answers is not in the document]
QuintonJason: Fixed in 1d43a156. In `disable-top-layer` mode the Escape branch of `handleKeyDown` now bails when `document.activeElement` is a real node outside the modal, so a body-mounted overlay that owns focus keeps Escape and this modal is not dismissed out from under it. When focus is inside the modal (or on `document.body`), Escape still closes it. New e2e test in 2ef45d4e covers both paths.

QuintonJason: Good catch — this was addressed after the commit you reviewed (a3c820d7, since replaced). The current tip guards the Tab branch on `disableTopLayer` (6abab99c, `pds-modal.tsx` handleKeyDown), so focus is no longer trapped and Tab reaches an overlay stacked above. I also closed the related Escape-ownership gap (1d43a156): when an external overlay owns focus, Escape is left to it rather than closing the modal, so a keyboard user can dismiss the overlay without losing the modal. e2e coverage in 2ef45d4e.

libs/core/src/components/pds-modal/pds-modal.tsx
QuintonJason: Went with the open-time-only documentation route you suggested rather than adding a `@Watch`. The `disableTopLayer` JSDoc now states it is read when the modal opens and that changing it while the modal is open is not supported (1d43a156, regenerated into readme.md / components.d.ts). This is an escape hatch set once at open time, so a live toggle is not a real use case and a watcher would just add surface area.

libs/core/src/components/pds-modal/docs/pds-modal.mdx
QuintonJason: Fixed in c16b5acd — added the blank line between `</DocCanvas>` and the `### Overlays Above the Modal` heading so MDX v3 parses it as a heading instead of shipping the literal string.

libs/core/src/components/pds-modal/docs/pds-modal.mdx
QuintonJason: Both fixed in c16b5acd. Reworded the note to the actual constraint — Escape closes the modal while focus is inside it, and is left to an overlay above the modal when that overlay owns focus (matching the new component behavior in 1d43a156). The demo overlay now cleans itself up: it removes on click, on Escape, and on the modal `pdsModalClose` event (`{ once: true }`), so closing via Escape / footer Close / header X can no longer orphan it over the docs page. Same cleanup mirrored in the Storybook `DisableTopLayer` story.

pixelflips (approved):
Thanks for addressing the feedback. LGTM! 👍🏼

QuintonJason:
@cursor review

PR #805, opened 2026-09-21
fix(pds-tab): flatten a nested control left by a page-cache restore

QuintonJason:
# Description

Surfaced in kajabi-products PR #56992: pressing Back on an admin tab strip (e.g. Clubs Posts/Resources/Activity) could show two underlines under the active tab.

`pds-tab` is `shadow: false` — its rendered `<a>`/`<button>` wrapper lives in this element's own light DOM, not behind a shadow boundary. A Turbo page-cache restore (or a native browser bfcache restore) can reconnect this element with that previous render still sitting in its children. Stencil's light-DOM slot relocation treats that stale wrapper as "the" slotted content and nests it inside the freshly rendered wrapper's `.pds-tab__content` — a second `<a>`/`<button>`, and a second active-tab underline, nested inside the first.

kajabi-products worked around this at the consumer level with a `turbo:before-cache` listener that manually tears every `pds-tab` back down to its server-rendered shape before Turbo snapshots the page. That's Turbo-specific and has to be reinvented by every consumer — the same bug would hit a native browser bfcache restore with no Turbo involved at all.

This fixes it in the component: `componentDidRender` now checks whether the wrapper it just rendered contains another `a.pds-tab`/`button.pds-tab` instead of plain content, and if so, pulls the real (innermost) content out and drops the stale nested one. It runs on every render, so it self-heals regardless of what caused the reconnect — Turbo cache, bfcache, or a manual DOM move — with no `shadow: true` migration and no consumer-side code required.

(An earlier attempt hooked `componentWillLoad` to unwrap the stale markup before `render()`, but Stencil captures the light-DOM content it relocates into the slot *before* `componentWillLoad` runs, so that approach lost the tab's label text instead of fixing the nesting. Cleaning up after `render()` avoids depending on that internal timing.)

Fixes #(issue)

## Type of change

# How Has This Been Tested?

- Unit tests (`pds-tab.spec.tsx`): both nav (anchor) and panel (button) tab shapes reconnecting over already-hydrated markup end up with exactly one control and the original label text preserved; a "pristine, never-hydrated" tab is left untouched; running the cleanup twice on the same snapshot is a no-op (idempotent).
- E2E test (`pds-tab.e2e.ts`, real Puppeteer/Chrome): renders a `pds-tabs nav` tab, captures its hydrated `outerHTML`, then re-parses that same markup into the DOM the way a page-cache restore does — reproducing the actual bug end-to-end and confirming both the single-anchor structure and label text survive.
- `npx nx run @pine-ds/core:test`: full suite green (92 suites / 3037 tests).
- `npx nx run @pine-ds/core:lint`: 0 errors, no new warnings.

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions: `@pine-ds/core` main (post pds-tabs nav mode, #792)
- OS: macOS
- Browsers: Chrome (Puppeteer, via Stencil e2e)
- Screen readers: n/a
- Misc:

# Checklist:

---

> [!NOTE]
> **Low Risk**
> Targeted post-render DOM cleanup in one Stencil component with broad test coverage; no auth, data, or API changes.
> 
> **Overview**
> Fixes **double active-tab underlines** when users navigate back to a page with nav-mode `pds-tab` strips (Turbo cache or browser bfcache). Because the component uses light DOM (`shadow: false`), a restore can leave a stale rendered `<a>`/`<button>` inside the element; Stencil then nests that wrapper inside the new one’s `.pds-tab__content`.
> 
> **`componentDidRender`** now detects nested `a.pds-tab` / `button.pds-tab` inside `.pds-tab__content`, moves the real label nodes up, and removes the stale control—repeatedly for doubly-nested snapshots. Cleanup runs after render so slot relocation timing doesn’t strip tab labels (unlike an earlier `componentWillLoad` approach).
> 
> Unit and e2e tests cover nav and panel tabs, double reconnect, bfcache-style disconnect/reconnect, rich label content, idempotency, and normal re-renders after cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 369594c753c8209aeddad7d2d55a0c2d25fc7561. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/core/src/components/pds-tabs/pds-tab/test/pds-tab.e2e.ts
```diff
@@ -8,4 +8,34 @@ describe('pds-tab', () => {
     const element = await page.find('pds-tab');
     expect(element).toHaveClass('hydrated');
   });
+
+  describe('reconnecting over an already-hydrated snapshot', () => {
+    // Reproduces a Turbo page-cache (or browser bfcache) restore: the DOM that
```
QuintonJason: 1 line comment

libs/core/src/components/pds-tabs/pds-tab/test/pds-tab.spec.tsx
```diff
@@ -239,4 +239,75 @@ describe('pds-tabs', () => {
       expect(anchor?.getAttribute('role')).toBe('link');
     });
   });
+
+  describe('reconnecting over an already-hydrated snapshot', () => {
+    // A page-cache restore (Turbo, browser bfcache) can reconnect this element
```
QuintonJason: 1 line comment

libs/core/src/components/pds-tabs/pds-tab/pds-tab.tsx
```diff
@@ -8,6 +8,28 @@ import { Component, Element, Fragment, Host, h, Prop, Event, EventEmitter } from
 export class PdsTab {
   @Element() el: HTMLPdsTabElement;
 
+  // A page-cache restore (Turbo, browser bfcache) or any other DOM move can
```
QuintonJason: 1 line comment

QuintonJason: Trimmed to one line in 30b15ff5.

libs/core/src/components/pds-tabs/pds-tab/test/pds-tab.spec.tsx
QuintonJason: Trimmed to one line in 30b15ff5.

libs/core/src/components/pds-tabs/pds-tab/test/pds-tab.e2e.ts
QuintonJason: Trimmed to one line in 30b15ff5.

libs/core/src/components/pds-tabs/pds-tab/pds-tab.tsx
```diff
@@ -8,6 +8,18 @@ import { Component, Element, Fragment, Host, h, Prop, Event, EventEmitter } from
...
+  // Flattens a nested <a>/<button> left when a page-cache restore (Turbo, bfcache) reconnects this light-DOM element over its own prior render.
+  componentDidRender() {
+    const contentDiv = this.el.querySelector('.pds-tab__content');
+    const nestedControl = contentDiv?.firstElementChild;
+    if (!nestedControl?.matches('a.pds-tab, button.pds-tab')) return;
+
+    const nestedContent = nestedControl.querySelector('.pds-tab__content');
+    const nodes = Array.from((nestedContent ?? nestedControl).childNodes);
+    while (contentDiv.firstChild !== null) contentDiv.removeChild(contentDiv.firstChild);
+    nodes.forEach((node) => contentDiv.appendChild(node));
```
pixelflips: **The cleanup can destroy the tab label when the restored snapshot is nested more than once.**

This clears every child of `.pds-tab__content`, but re-appends only what was extracted from the *first* nested control. Anything else in that div is lost.

Reproduced with `newSpecPage` on Stencil 4.43.5, feeding a doubly-nested snapshot — i.e. a restore of a page that was itself restored while this bug was live:

```html
<pds-tab selected="true" parent-component-id="foo" name="two">
  <button role="tab" class="pds-tab is-active">
    <div class="pds-tab__content">
      <button role="tab" class="pds-tab is-active">
        <div class="pds-tab__content">Content</div>
      </button>
    </div>
  </button>
</pds-tab>
```

Slot relocation splits that into two *sibling* controls inside the fresh content div, so at the top of `componentDidRender` the DOM is:

```html
<div class="pds-tab__content">
  <button class="pds-tab is-active"><div class="pds-tab__content"></div></button>
  <button class="pds-tab is-active"><div class="pds-tab__content">Content</div></button>
</div>
```

`nestedControl` is the first button, whose own content div is empty, so `nodes` is `[]`. The `while` loop then removes **both** buttons — including the second one holding the real text — and nothing is re-appended. Result:

```html
<button role="tab" id="foo__two" class="pds-tab is-active"><div class="pds-tab__content"></div></button>
```

The label is gone, leaving an unlabeled tab/link.

On reachability: once this ships, each restore flattens back to a single level, so it settles at a stable fixed point. The exposure is pages already sitting in a browser's bfcache or Turbo cache from before the deploy. Transitional — but a blank label is a worse failure mode than the doubled underline being fixed, and it is also an a11y defect.

Unwrapping in place rather than clearing wholesale avoids this by construction, since nothing is removed that wasn't first moved out. Sketch:

```ts
componentDidRender() {
  const contentDiv = this.el.querySelector('.pds-tab__content');
  if (!contentDiv) return;

  let nestedControl = contentDiv.firstElementChild;
  while (nestedControl?.matches('a.pds-tab, button.pds-tab')) {
    const nestedContent = nestedControl.querySelector('.pds-tab__content');
    Array.from((nestedContent ?? nestedControl).childNodes)
      .forEach((node) => contentDiv.insertBefore(node, nestedControl));
    nestedControl.remove();
    nestedControl = contentDiv.firstElementChild;
  }
}
```

On the snapshot above this yields `<div class="pds-tab__content">Content</div>`. Worth a spec case for the doubly-nested shape either way — the current tests only cover a single level, so this path is unguarded.

For reference, these cases were verified to behave correctly as written: single-nested nav and panel tabs, element-rich labels (icon + text + badge), the `availability` variant's edge spans, pristine never-hydrated content, and a real re-render after a successful flatten.

pixelflips (commented):
One issue on the new `componentDidRender` cleanup — details inline.

libs/core/src/components/pds-tabs/pds-tab/pds-tab.tsx
```diff
@@ -8,6 +8,28 @@ import { Component, Element, Fragment, Host, h, Prop, Event, EventEmitter } from
...
+  }
```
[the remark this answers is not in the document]
QuintonJason: Checked this against the mechanism Stencil actually uses for light-DOM slot relocation: it's an object-identity reference, not a DOM comment marker, so moving the node via appendChild doesn't detach it from what Stencil relocates on a later render.

Verified empirically in 8d10a2d4 with three added tests (not just assertion — each actually exercises the scenario):
- a genuine Stencil re-render after cleanup (prop change → real render() call, not just re-invoking the cleanup)
- a second full reconnect after the first cleanup already ran (repeated Turbo-style outerHTML reparse)
- the same element instance disconnected and reconnected twice (native-bfcache-accurate, not a fresh instance)

Label content and single-anchor structure survive all three. Didn't reproduce the described failure — happy to dig further if you can point to a specific sequence that triggers it.

libs/core/src/components/pds-tabs/pds-tab/pds-tab.tsx
QuintonJason: Good catch, and the sketch is correct — applied it as-is in 369594c7 (unwrap-in-place via insertBefore before removing, looping until no nested control remains). Traced through your doubly-nested repro and confirmed the loop settles at the single flat <button><div class="pds-tab__content">Content</div></button> you'd expect.

Added your doubly-nested case as a spec test, plus one for element-rich label content (icon + text) surviving the unwrap, alongside the existing single-nested/pristine/re-render coverage. All green.

pixelflips (approved):
Thanks for addressing the feedback, LGTM! 👍🏼

PR #806, opened 2026-09-21
fix(pds-modal): fix page-cache reconnect bug across modal, tabpanel, tooltip, radio and sortable

QuintonJason:
# Description

Follow-up to #805 (merged in 3.31.0). While fixing `pds-tab`'s page-cache reconnect bug, a full sweep of every `shadow: false` component using `<slot>` turned up five more components with the exact same vulnerable pattern: `render()` wraps slotted content in a newly-created element with a distinguishing class/tag, so a Turbo page-cache or browser bfcache restore reconnects the component over its own prior render, nesting a stale wrapper inside the fresh one.

Affected:
- `pds-tabpanel` — nested `.pds-tabpanel` div
- `pds-modal` — nested `dialog.pds-modal__backdrop` > `.pds-modal` div
- `pds-modal-header` — nested `header.pds-modal__header`
- `pds-modal-footer` — nested `footer.pds-modal__footer`
- `pds-tooltip` — nested trigger `span.pds-tooltip__trigger` **and** content wrapper `div.pds-tooltip__content-slot-wrapper`

Severity varies (pds-tab's was visually obvious — a doubled underline, plus interactive nested `<a>`/`<button>`), but all five have the same underlying defect: invalid nested landmark/dialog/role structure, duplicate `id`/`aria-labelledby`, or (for tooltip) content silently landing in the wrong slot.

`pds-tooltip` needed bespoke handling rather than the shared helper: its trigger and content wrappers are two separate slots that don't carry their own `slot` attribute, so on reconnect both stale wrappers land as *siblings* inside the fresh trigger — content meant for the content slot has to be explicitly rerouted there, not just flattened in place. It also required scoping the initial "find the fresh wrapper" lookup to `this.el`'s **direct children only** — `querySelector` would otherwise match a stale wrapper nested inside the trigger before it reaches the real, top-level content-slot-wrapper sibling (verified this by reproducing the actual bug: content briefly landed in and was then deleted from the wrong wrapper before the fix).

## Type of change

# How Has This Been Tested?

Added a `reconnecting over an already-hydrated snapshot` describe block to each component's spec suite (creating new spec files for `pds-modal-header`/`pds-modal-footer`, which had none), each reproducing the actual nested-snapshot shape and asserting a single flat wrapper with content intact, plus a "leaves pristine content alone" case per component.

- `npx nx run @pine-ds/core:test`: full suite green (94 suites / 3054 tests, up from 92/3037 at #805 merge — 6 new spec files/blocks).
- `npx nx run @pine-ds/core:lint`: 0 errors, no new warnings.

- [x] unit tests
- [ ] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions: `@pine-ds/core` main (post 3.31.0 / #805)
- OS: macOS
- Browsers: n/a (spec-level coverage; #805 already has e2e coverage for the shared unwrap pattern)
- Screen readers: n/a
- Misc:

# Checklist:

---

> [!NOTE]
> **Medium Risk**
> Touches render-time DOM mutation on modals, forms, and tooltips where bad reconnect state can affect structure and duplicate ids; changes are guarded by many unit tests and only run on reconnect-shaped markup.
> 
> **Overview**
> Follow-up to the `pds-tab` page-cache fix: adds shared **`unwrapReconnectedContent`** and wires **`componentDidRender`** cleanup across other **`shadow: false`** components so Turbo/bfcache reconnects do not leave nested stale wrappers (duplicate dialogs, landmarks, or slot markup).
> 
> **Modal family** (`pds-modal`, `pds-modal-content`, header/footer) and **tabs** (`pds-tab` now uses the helper, `pds-tabpanel`) call the shared unwrap. **`pds-tooltip`** gets custom sibling rerouting between trigger and content wrappers, run **once per connect** via `connectedCallback`. **`pds-radio`**, **`pds-radio-group`**, and **`pds-sortable-item`** remove or flatten Stencil-hidden stale siblings with component-specific logic.
> 
> Adds **`expectReconnectSafe`** and broad unit specs (including new modal header/footer spec files) that simulate reconnecting over an already-hydrated snapshot.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 510a63e631052ec6f7235b327456e33e54f88787. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/core/src/components/pds-tabs/pds-tab/pds-tab.tsx
```diff
@@ -8,20 +9,13 @@ import { Component, Element, Fragment, Host, h, Prop, Event, EventEmitter } from
...
-    while (nestedControl?.matches('a.pds-tab, button.pds-tab')) {
-      const nestedContent = nestedControl.querySelector('.pds-tab__content');
-      Array.from((nestedContent ?? nestedControl).childNodes).forEach((node) =>
-        contentDiv.insertBefore(node, nestedControl)
-      );
-      nestedControl.remove();
-      nestedControl = contentDiv.firstElementChild;
-    }
+    unwrapReconnectedContent(
+      this.el.querySelector('.pds-tab__content'),
```
pixelflips, line 15: Scope: `pds-tab` is the one component in this PR with **no** behavior change — this is a straight extraction of the #805 logic into the shared helper. The actual fixes land in `pds-modal`, `pds-modal-header`, `pds-modal-footer`, `pds-tabpanel` and `pds-tooltip`, none of which the scope names.

`CLAUDE.md` (Git Conventions): "Commit scope should match the component name (e.g., `feat(pds-table):`, `fix(pds-button):`)."

Since the PR title becomes the squash commit, this is worth retitling before merge — `fix(pds-modal)` covers the largest affected surface, or drop to a broader scope given it spans five components.

libs/core/src/utils/reconnected-content.ts
```diff
@@ -0,0 +1,28 @@
...
+export function unwrapReconnectedContent(
+  container: Element | null,
+  selector: string,
+  contentSelector?: string,
+): void {
+  if (container === null) return;
+
+  let stale = container.firstElementChild;
+  while (stale?.matches(selector)) {
+    const source = contentSelector !== undefined ? (stale.querySelector(contentSelector) ?? stale) : stale;
```
pixelflips, line 23: This is the only file in `libs/core/src/utils/` without a matching spec in `libs/core/src/utils/test/` — `attributes.ts`, `form.ts`, `overlay.ts`, `scroll.ts` and `truncation-tooltip.ts` each have one.

A `reconnected-content.spec.ts` would be the natural place to pin the two behaviors that are currently only reachable indirectly through five different component suites: the `contentSelector` present/absent split on this line, and the multi-iteration loop. As it stands a regression in the shared loop has to be caught five times over rather than once, and the doubly-nested shape that drives the loop past one iteration isn't exercised by any of the five.

Applies to the new reconnect blocks in `pds-modal-header`, `pds-modal-footer` and `pds-tabpanel` too — each has a single-level case and a pristine case, where [`pds-tab.spec.tsx`](https://github.com/Kajabi/pine/blob/032eed1c7947ea7118255c4020963bd6f1a14bed/libs/core/src/components/pds-tabs/pds-tab/test/pds-tab.spec.tsx#L243-L367) ended up with seven. I verified all three doubly-nested shapes flatten correctly today; they're just unguarded.

libs/core/src/components/pds-tooltip/test/pds-tooltip.spec.tsx
```diff
@@ -55,6 +55,49 @@ describe('pds-tooltip', () => {
     `);
   });
 
+  describe('reconnecting over an already-hydrated snapshot', () => {
```
pixelflips, line 59: This is the adopter I'd most want a re-render test on. `unwrapReconnectedSlots` moves nodes into a **different** Stencil-rendered element (`contentWrapper.appendChild`), whereas `pds-tab` only ever moved nodes within a single container. That cross-container reroute is new in this PR and wasn't covered by #805's verification.

It's also close to what Bugbot raised on #805 (detaching whatever reference Stencil uses to relocate slotted content), which is why [`pds-tab.spec.tsx`](https://github.com/Kajabi/pine/blob/032eed1c7947ea7118255c4020963bd6f1a14bed/libs/core/src/components/pds-tabs/pds-tab/test/pds-tab.spec.tsx#L343-L356) carries "survives a genuine Stencil re-render after cleanup". There's no equivalent here.

I ran it locally — change `placement`, `await page.waitForChanges()`, and the content stays in the right wrapper with no duplicate trigger — so this is asking for a guard on working behavior. Same for the doubly-nested shape, also verified passing.

pixelflips (commented):
Mechanism looks right. I traced all four `unwrapReconnectedContent` call shapes against each component's real `render()` output, and ran throwaway spec probes for the scenarios this PR doesn't cover — doubly-nested snapshots for modal, modal-header, tabpanel and tooltip, plus a genuine re-render after the tooltip's reroute. They all pass, so nothing below is a bug report.

Two things worth addressing before merge: the commit/PR scope, and regression coverage matching the bar set on #805. Details inline.

libs/core/src/components/pds-modal/test/pds-modal.spec.tsx
```diff
@@ -258,4 +258,38 @@ describe('pds-modal', () => {
     // the spec mock-doc environment does not implement — it is covered in the e2e
     // suite where a real browser exercises focus and key events.
   });
+
+  describe('reconnecting over an already-hydrated snapshot', () => {
```
pixelflips, line 263: Missing the doubly-nested case. The bar for this pattern got set on [#805](https://github.com/Kajabi/pine/pull/805#discussion_r4064839436): "Worth a spec case for the doubly-nested shape either way — the current tests only cover a single level, so this path is unguarded."

That matters most here, because `pds-modal` is the only call site where the container class (`.pds-modal`) differs from the staleness selector (`dialog.pds-modal__backdrop`) — it's the only adopter with two wrapper levels in `render()`, so the container/selector/`contentSelector` triple is manually coupled to that nesting with nothing enforcing it. A later change to the modal's wrapper structure could silently break the unwrap.

Verified locally that a doubly-nested snapshot does flatten to one `dialog` and one `.pds-modal` with body content intact. A test would keep it that way.

QuintonJason:
Thanks for the thorough re-read — all five confirmed and fixed, plus verified the actual reconnect shape for each empirically (not just from reading the compiled output) before writing a fix:

- **`pds-modal-content`** — confirmed vulnerable, same shape as header/footer. Simple fix via the shared util. (This one was a pure miss on my part — it has no `shadow`/`scoped` key at all, so my original "shadow: false" text-match sweep skipped it. Redid the sweep as "anything not `shadow: true`" and confirmed the corrected list is exactly your 5 + the 6 already fixed — nothing else missed.)
- **`pds-radio-group`** — confirmed. Reconnect sweeps the label, radios wrapper, *and* helper/error messages into one blob inside the fresh radios div, duplicating the helper/error ids exactly as you predicted. Fix unwraps the radios wrapper and discards the stale label/message siblings outright (their fresh counterparts already reflect current props).
- **`pds-sortable-item`** — confirmed, the sibling variant. Bare default slot means Stencil doesn't relocate anything for it — stale handle/actions divs just sit in place. Fix discards the stale handle (fresh is always first, JSX-declared first) and rescues the real `[slot="sortable-item-actions"]` content out of the stale actions wrapper into the fresh one (fresh is always last).
- **`pds-radio`** — resolved the "unclear": genuinely vulnerable. Nothing here is slotted except the conditional `image` slot, so Stencil hides (via `hidden`, doesn't remove) the entire stale label/input/message copy — leaving a duplicate `id` sitting in the DOM even though it's invisible. Also found the real `[slot="image"]` content gets stranded inside the stale hidden image-container; the fix rescues it before discarding the rest.
- **`pds-sortable`** — confirmed safe as you said (bare slot, no wrapper at all), locked in with a test.

Also built the generic reconnect-safety guard you suggested (`expectReconnectSafe` in `libs/core/src/utils/test/reconnect-safety.ts`): hydrates a fixture, re-parses its own output as a reconnect would, and asserts no duplicate ids and no content loss/duplication. Wired it into all 11 affected components alongside the specific tests, so it's there for future components too.

Separately, a code review on the PR caught that my tooltip fix (from the earlier round) didn't recursively drain the content-side wrapper the way the trigger side did, and ran on every render instead of once per connect — fixed both.

libs/core/src/components/pds-tabs/pds-tab/pds-tab.tsx
QuintonJason: Retitled to fix(pds-modal) — good catch, the squash commit would've mis-scoped this otherwise.

libs/core/src/utils/reconnected-content.ts
QuintonJason: Added libs/core/src/utils/test/reconnected-content.spec.ts (510a63e6) — direct coverage of both the contentSelector present/absent split and the doubly-nested drain-in-one-call loop, plus a null-container and no-nested-content-element fallback case. Caught a real fixture bug of my own while writing it: an innerHTML string with two nested <a> elements gets auto-corrected by the HTML parser (no nested interactive content), which doesn't reflect how Stencil's live-DOM slot relocation actually produces this shape — switched that one case to build the DOM via createElement/appendChild instead.

Left pds-modal-header/footer/pds-tabpanel's component-level blocks at single-level + pristine, now that the shared loop's doubly-nested behavior is pinned once at the util level rather than needing re-coverage in every consumer.

libs/core/src/components/pds-tooltip/test/pds-tooltip.spec.tsx
QuintonJason: Added 'keeps content in the right wrapper after a genuine re-render post-cleanup' (510a63e6) — same pattern as pds-tab's re-render guard, but asserting the actual DOM (single trigger, single content wrapper, content in the right one, none leaked into the trigger) after a placement change, not just that unwrapReconnectedSlots wasn't re-invoked.

libs/core/src/components/pds-modal/test/pds-modal.spec.tsx
QuintonJason: Added 'drains a doubly-nested dialog to a single one with content intact' (510a63e6), asserting single dialog + single .pds-modal + header/body present after flattening the two-level structure — locks in the container/selector/contentSelector coupling you flagged.

pixelflips (approved):
Thanks for addressing the feedback! LGTM! 👍🏼

PR #808, opened 2026-09-23
feat(pds-chip): add maxWidth prop to auto-truncate the slotted label

QuintonJason:
# Description

Surfaced from a kajabi-products consumer trying to ellipsize a long filename in a tag-variant chip: `pds-chip` has no size limit of its own, and there was no way to cap it from outside at all. Two commits:

**fix(pds-chip): let label shrink below its content width when the host is constrained**

`.pds-chip__label` (the shadow-internal flex wrapper around the default slot) never set `min-width`, so it kept the browser's default `min-width: auto` on a flex item — which refuses to shrink below the slotted content's unwrapped width no matter what a consumer set on the host (`max-width`, `min-width: 0`, even an inline style with maximum specificity had no effect). There was also no `::part()` exposed to reach around this from outside.

**feat(pds-chip): add maxWidth prop to auto-truncate the slotted label**

Rather than ship the above as a documented "set these 4 properties yourself on both the host and your slotted node" workaround, this adds a real `maxWidth` prop that does the whole thing internally:

- `@Prop() maxWidth?: string` sets `max-width` + `min-width: 0` on the host.
- The default slot (in both the label and dropdown-button render paths) is now always wrapped in an internal `<span class="pds-chip__label-text">` with `overflow: hidden; text-overflow: ellipsis; white-space: nowrap; min-width: 0`. This is unconditional and harmless when nothing overflows — it only matters once `maxWidth` is set.

The internal wrapper matters because `::slotted()` can only select an assigned *element*, never a bare text node — and the common case is `<pds-chip>Some text</pds-chip>`, slotting raw text with nothing for `::slotted()` to ever target. Wrapping the slot ourselves means truncation works for that case too, not just when a consumer happens to wrap their own label in a span.

Consumer usage collapses to `<pds-chip max-width="200px">A very long label</pds-chip>` — no styling of their own required.

Fixes #(issue)

## Type of change

# How Has This Been Tested?

- e2e: a bare-text tag-variant chip and a bare-text dropdown-variant chip both shrink well below their natural width (~403px unconstrained) when `max-width` is set; a control asserts an unconstrained chip stays wide. Verified the original assertion fails without the `pds-chip.scss` fix (constrained width == natural width).
- spec: `maxWidth` prop sets/omits the expected inline host style; the internal `.pds-chip__label-text` wrapper is present under `.pds-chip__button` for the dropdown variant too. All 24 existing shadow-DOM snapshot assertions updated for the new wrapper span.
- Storybook: `TruncatedLabel` story updated to use the real prop with a bare-text label. Docs section rewritten to describe the prop instead of the manual workaround. Verified locally in a real browser (not just that the build succeeds) that the docs page renders without the React error a string-typed JSX `style` prop caused in an earlier revision of this PR.
- Full `@pine-ds/core` suite: 3093/3093 passing. Lint and stylelint clean (0 errors).

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [x] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions: main (post-3.31.1)
- OS: macOS
- Browsers: headless Chromium (Puppeteer, via `newE2EPage`); manual check in a real browser against a local Storybook build
- Screen readers:
- Misc:

# Checklist:

---

> [!NOTE]
> **Low Risk**
> Additive, opt-in API with extensive tests; default chip layout/behavior is preserved when maxWidth is unset.
> 
> **Overview**
> Adds a **`maxWidth`** prop (`max-width` attribute) so `pds-chip` can cap its width and ellipsize long slotted labels (e.g. filenames in tag chips) without consumer CSS.
> 
> When set, the host gets `--pds-chip-max-width`, `box-sizing: border-box`, and flex `min-width: 0` so the cap includes padding; truncation styles apply only under `:host([max-width])`. The default slot is always wrapped in an internal **`.pds-chip__label-text`** span so bare text (not just slotted elements) can clip, with **`flex-shrink: 0`** on the status dot so it does not squash under pressure.
> 
> Truncated labels reuse **`setupTruncationTooltip`** (same as `pds-text` / table cells): hover/focus shows the full label, **`tabindex="0"`** on the label span when truncating (dropdown uses the button for focus), and **`componentDidRender`** re-binds the tooltip after variant remounts. Behavior without `maxWidth` is unchanged (wrap in narrow containers; no forced nowrap).
> 
> Docs, Storybook (`TruncatedLabel`), generated types/readme, and broad unit/e2e coverage are included.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38cdd45cef995495463909daae3291d6f3b31005. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/core/src/components/pds-chip/stories/pds-chip.stories.js
```diff
@@ -150,3 +151,17 @@ TagWithRemoveTarget.args = {
   removeUrl: "/clear-filters",
   removeTarget: "_blank",
 }
+
+// maxWidth ellipsizes plain slotted text with no styling of its own required
```
QuintonJason: remove comment

libs/core/src/components/pds-chip/test/pds-chip.e2e.ts
```diff
@@ -34,6 +34,49 @@ describe('pds-chip', () => {
...
+    const page = await newE2EPage();
+    await page.setContent(`
+      <pds-chip variant="tag" max-width="100px">
+        A very long label that would otherwise overflow the chip
+      </pds-chip>
+    `);
+
+    const width = await page.$eval('pds-chip', (el) => el.getBoundingClientRect().width);
+
+    // Not pixel-exact against the 100px cap: the tag variant's own padding,
```
QuintonJason: remove comment test is self explanatory

libs/core/src/components/pds-chip/pds-chip.scss
```diff
@@ -88,12 +88,28 @@ $pds-chip-sentiment-hover: (
...
+  // content's unwrapped size, which would make maxWidth on the host do
+  // nothing.
+  min-width: 0;
 
   pds-icon {
     margin-inline-end: var(--pine-dimension-2xs);
   }
 }
 
+// Wraps the default slot so maxWidth can ellipsize it regardless of what's
```
QuintonJason: remove comment, unnecessary

libs/core/src/components/pds-chip/pds-chip.scss
```diff
@@ -113,6 +129,9 @@ $pds-chip-sentiment-hover: (
   border-radius: var(--pine-dimension-sm);
   cursor: pointer;
   display: flex;
+  // Same reason as .pds-chip__label: without this, maxWidth on the host has
```
QuintonJason: remove comment

libs/core/src/components/pds-chip/pds-chip.tsx
```diff
@@ -82,6 +82,11 @@ export class PdsChip {
    */
   @Prop() dismissLabel = 'Remove';
 
+  /**
+   * Truncates the slotted label with an ellipsis once the chip reaches this width, instead of growing to fit it. Accepts any CSS length (e.g. '200px', '20ch'). The slot is documented as label text; slotting richer markup with maxWidth set may truncate it unexpectedly.
```
QuintonJason: remove comment truncation is implied

QuintonJason: Kept a short JSDoc rather than deleting it entirely — Pine's CLAUDE.md requires a description on every @Prop, and removing it would leave maxWidth as the only undocumented prop / blank cell in the readme table. Trimmed it to just the accepted-value format and the slot caveat, dropped the part restating what truncation does. Let me know if you wanted it gone entirely instead.

libs/core/src/components/pds-chip/pds-chip.scss
```diff
@@ -88,12 +88,24 @@ $pds-chip-sentiment-hover: (
...
   }
 }
 
+.pds-chip__label-text {
+  display: block;
+  min-width: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
```
pixelflips: These rules apply to **every** chip, not just ones with `maxWidth` set — only the host's inline style is gated on the prop.

Measured with no `maxWidth`, label "Quarterly revenue report final" in a 160px container:
- before: 160 × 36px — wrapped, fits
- after: 222 × 24px — overflows by 62px, and *not* ellipsized, since nothing caps the width

`overflow: hidden` + `min-width: 0` also caps this box's contribution to every ancestor's min-content, so the chip's shrink floor drops from "longest word" to padding-only.

Suggest gating on `:host([max-width])` — the prop already reflects.

libs/core/src/components/pds-chip/pds-chip.scss
```diff
@@ -88,12 +88,24 @@ $pds-chip-sentiment-hover: (
 .pds-chip__label {
   align-items: center;
   display: flex;
+  // A flex item's default min-width: auto refuses to shrink below its
+  // content's unwrapped size, which would make maxWidth on the host do
+  // nothing.
+  min-width: 0;
```
pixelflips: This lets the label be squeezed for the first time, and `.pds-chip__dot` has no `flex-shrink: 0`. It's an empty `<i>`, so its min-content is 0 and it shrinks while its height stays fixed.

Measured with icon + dot: dot 6.00px unconstrained → 3.13px at `max-width: 120px` → 2.44px at 60px, rendering as an oval. `pds-icon` is unaffected — its svg's intrinsic size floors it.

Same guard used at `pds-filter.scss:33`, `pds-copytext.scss:36`, `pds-alert.scss:132`, `pds-switch.scss:41`.

libs/core/src/components/pds-chip/pds-chip.tsx
```diff
@@ -82,6 +82,11 @@ export class PdsChip {
    */
   @Prop() dismissLabel = 'Remove';
 
+  /**
+   * Accepts any CSS length (e.g. '200px', '20ch'). The slot is documented as label text — slotting richer markup may not truncate the way you expect.
```
pixelflips: This never states what the prop does — it leads with the value format. Every other prop in this file leads with the behavior.

`pds-tooltip.tsx:74` and `pds-popover.tsx:103` both already document a `maxWidth` as "Sets the maximum width of the {tooltip,popover} content".

CLAUDE.md: "When adding a prop or event that exists in a similar form on another component, match the wording and phrasing of the existing description."

pixelflips (commented):
Six inline findings below, highest-impact first at `pds-chip.scss` L101-107.

libs/core/src/components/pds-chip/pds-chip.tsx
```diff
@@ -82,6 +82,11 @@ export class PdsChip {
...
+   */
+  @Prop({ reflect: true }) maxWidth?: string;
```
pixelflips: Nothing reveals the clipped text — no tooltip, no `tabIndex`, no `title`.

Pine has `setupTruncationTooltip` (`libs/core/src/utils/truncation-tooltip.ts`), used by `pds-text.tsx:130`, `pds-table-cell.tsx:109`, `pds-copytext.tsx:75`. The first two also set `tabIndex={... ? 0 : undefined}` — both parts are needed here, since the label span isn't focusable, so a focus-triggered tooltip would never fire on the text/tag variants.

Screen readers are unaffected; `text-overflow` doesn't truncate the accessible name.

libs/core/src/components/pds-chip/pds-chip.tsx
```diff
@@ -137,24 +142,39 @@ export class PdsChip {
...
     );
 
     return chipContent;
   }
 
+  private get hostStyles() {
+    // min-width: 0 overrides a flex item's default min-width: auto, which
+    // otherwise refuses to shrink below the slotted content's unwrapped
+    // width and makes maxWidth alone do nothing.
+    return this.maxWidth ? { maxWidth: this.maxWidth, minWidth: '0' } : {};
```
pixelflips: `:host` is content-box — no `box-sizing: border-box` here and no global reset — so `padding-inline` is added on top of `max-width`.

`max-width="200px"` renders:
- text/tag, md: **216px** (`--pine-dimension-100`, 8px)
- text/tag, lg: **224px** (`--pine-dimension-150`, 12px)
- dropdown: **200px** (`padding: var(--pine-dimension-none)`)

Same prop value, three different widths. The e2e tests can't catch it — a 100px cap asserted only as `toBeLessThan(150)`.

libs/core/src/components/pds-chip/pds-chip.tsx
```diff
@@ -216,7 +236,7 @@ export class PdsChip {
 
   render() {
     return (
-      <Host class={this.classNames()} id={this.componentId}>
+      <Host class={this.classNames()} id={this.componentId} style={this.hostStyles}>
```
pixelflips: Writing `max-width`/`min-width` as inline styles means a consumer can't override them from a stylesheet without `!important` — e.g. `@media (max-width: 600px) { pds-chip { max-width: 120px } }` is silently ignored.

`pds-box` handles the same kind of prop via a custom property (`--sizing-min-width-box`, consumed at `pds-box.scss:21`). Noting `pds-row` does set it inline, so the convention isn't universal.

libs/core/src/components/pds-chip/pds-chip.scss
QuintonJason: > **Comment from kp-polish-pr bot** — I'm a bot. I'm responding for the author.
---

Fixed in 49c66bf9. Moved the truncation rules (display/overflow/text-overflow/white-space/min-width) plus the label & button `min-width: 0` under `:host([max-width])` — the prop reflects, so a chip with no maxWidth sizes to content and wraps as before. Added e2e coverage: a no-maxWidth chip in a 160px container stays ≤160 with `white-space: normal`, and only truncates (`nowrap`) when maxWidth is set.

libs/core/src/components/pds-chip/pds-chip.scss
QuintonJason: > **Comment from kp-polish-pr bot** — I'm a bot. I'm responding for the author.
---

Fixed. Added `flex-shrink: 0` to `.pds-chip__dot` (matching pds-filter/pds-copytext/pds-alert/pds-switch). e2e asserts the dot stays ≥5px wide under `max-width: 120px` so it can't collapse into an oval.

libs/core/src/components/pds-chip/pds-chip.tsx
QuintonJason: > **Comment from kp-polish-pr bot** — I'm a bot. I'm responding for the author.
---

Fixed. The description now leads with behavior — "Sets the maximum width of the chip, truncating the label with an ellipsis when it overflows." — matching pds-tooltip/pds-popover's "Sets the maximum width of the {…} content" phrasing, then keeps the accepted-value format + slot caveat. readme.md + components.d.ts regenerated.

libs/core/src/components/pds-chip/pds-chip.tsx
QuintonJason: > **Comment from kp-polish-pr bot** — I'm a bot. I'm responding for the author.
---

Fixed. Wired `setupTruncationTooltip` in `componentDidLoad` (plus a `@Watch('maxWidth')`), torn down in `disconnectedCallback`, with `contentEl` = the label span and `getTooltipText` = the host's textContent — same as pds-text/pds-table-cell/pds-copytext. Added `tabindex="0"` on the text/tag label span when maxWidth is set (the dropdown's own button already takes focus, so focusin bubbles), so the focus-triggered tooltip fires. e2e covers hover→tooltip and the conditional tabindex.

libs/core/src/components/pds-chip/pds-chip.tsx
QuintonJason: > **Comment from kp-polish-pr bot** — I'm a bot. I'm responding for the author.
---

Fixed. `:host([max-width])` now sets `box-sizing: border-box`, so the cap is the outer width including padding rather than sitting on top of it — one maxWidth value renders one width across text/tag/dropdown. Tightened the e2e width assertions from `<150` to `≤101` for a 100px cap so this can't regress unnoticed.

libs/core/src/components/pds-chip/pds-chip.tsx
QuintonJason: > **Comment from kp-polish-pr bot** — I'm a bot. I'm responding for the author.
---

Fixed. maxWidth is now handed to the stylesheet as a `--pds-chip-max-width` custom property (consumed by `:host([max-width])` in the SCSS) instead of an inline `max-width` — matching pds-box's `--sizing-min-width-box` — so a consumer can override the cap from a stylesheet/media query without `!important`. spec updated to assert the custom property + reflected attribute.

libs/core/src/components/pds-chip/pds-chip.tsx
```diff
@@ -82,11 +89,57 @@ export class PdsChip {
...
+    this.destroyTruncationTooltip();
+
+    if (this.labelTextEl) {
+      this.truncationCleanup = setupTruncationTooltip({
+        hostEl: this.el,
+        contentEl: this.labelTextEl,
+        getTooltipText: () => this.el.textContent || '',
+      });
+    }
+  }
```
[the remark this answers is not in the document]
QuintonJason: > **Comment from kp-polish-pr bot** — I'm a bot. I'm responding for the author.
---

Good catch — fixed in 2514c9be. The tooltip now re-binds in `componentDidRender` keyed on the label node's identity: when `variant` (or size) remounts `.pds-chip__label-text`, it tears down the old listeners/observer and re-attaches to the current node; an unrelated re-render (same node) leaves the working tooltip in place, and `disconnectedCallback` still cleans up. Added an e2e test that sets `max-width`, switches `variant` to `dropdown` at runtime, then hovers and asserts the full label still surfaces.

QuintonJason:
@cursor review

pixelflips:
Not a blocker, but something seems odd in the docs example. Doesn't truncate with ellipsis, and height is off. Prob something with the canvas styling, but just calling it out.

Screenshot 2026-09-23 at 9 29 40 AM

pixelflips (approved):
Left one non-blocker about the story preview, otherwise LGTM! Thanks for addressing the feedback. 👍🏼

QuintonJason:
> Not a blocker, but something seems odd in the docs example. Doesn't truncate with ellipsis, and height is off. Prob something with the canvas styling, but just calling it out.
> 
> Screenshot 2026-09-23 at 9 29 40 AM

Resolved, it was the issue with the `<p>` getting injected

Screenshot 2026-09-23 at 11 48 44 AM

PR #809, opened 2026-09-23
fix(pds-text): make truncate actually shrink the host, not just style its inner tag

QuintonJason:
# Description

Surfaced auditing for the pds-chip min-width bug class (DSS-250) — `pds-text`'s own `truncate` prop has the same root cause, with a bigger blast radius since it's the most widely-used truncation mechanism in Pine.

`:host { display: inline; }` unconditionally — an inline element ignores `width`/`max-width` entirely per spec, so the documented contract ("Must add a `width` to the element") silently does nothing *unless* the host happens to be a flex/grid item, which auto-blockifies it regardless of its specified display. Even then, a flex item's default `min-width: auto` still refuses to shrink below its content's own width. Net effect: `truncate` only ever worked by accident, in specific parent-layout contexts, and the accessible truncation-tooltip feature (`initTruncationTooltip`, gated on `scrollWidth > clientWidth`) never fired either, since a never-shrinking element is never "truncated" by that check.

The existing e2e test only asserted that `text-overflow: ellipsis` is a computed property on the internal tag — never that anything actually shrinks. That's how this shipped invisibly.

Fix: `:host([truncate]) { display: block; min-width: 0; }`. Gated entirely on the `truncate` attribute, so nothing changes for any consumer not using it — this only starts working where it previously silently didn't.

Verified no regression in every current consumer of `pds-text truncate`: `pds-alert` (small variant), `pds-copytext`, `pds-table-cell`.

Fixes #(issue)

## Type of change

# How Has This Been Tested?

Added two e2e tests measuring real layout, not just computed CSS: a width-constrained `truncate` host now actually has `scrollWidth > clientWidth` on its rendered tag (previously false, verified by reverting the SCSS fix and re-running — test failed exactly as expected); an explicit `width` with `truncate` unset still doesn't constrain anything (unchanged control). Ran the full `pds-alert`, `pds-copytext`, and `pds-table-cell` suites — all green, no regression. Full `@pine-ds/core` suite green, lint and stylelint clean.

- [x] unit tests
- [x] e2e tests
- [ ] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions: main (post-3.31.1)
- OS: macOS
- Browsers: headless Chromium (Puppeteer, via `newE2EPage`)
- Screen readers:
- Misc:

# Checklist:

---

> [!NOTE]
> **Low Risk**
> Scoped to elements with the `truncate` attribute; fixes layout where truncation was already intended, with e2e coverage and no change to default inline display.
> 
> **Overview**
> **`pds-text` `truncate` now honors width on the host** instead of only applying ellipsis styles on the inner tag while the custom element stayed `display: inline` (so `width`/`max-width` had no effect outside flex/grid luck).
> 
> When `truncate` is set, the host gets **`display: block`** and **`min-width: 0`** so a constrained width can shrink and ellipsis + truncation detection (`scrollWidth > clientWidth`) work in normal layouts. Consumers without `truncate` are unchanged.
> 
> E2E coverage adds **layout assertions**: truncated text in a 100px-wide host must overflow horizontally, and explicit width without `truncate` must still not clamp the host.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cb002ac33a15fbcbd613a3c4e93d6bb677361d0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/core/src/components/pds-text/pds-text.scss
```diff
@@ -66,6 +66,15 @@ strong {
   font-weight: bolder;
 }
 
+// A plain `display: inline` host (the default) ignores width/max-width
```
QuintonJason: 1 line comment

QuintonJason: Trimmed to one line — pushed in cb002ac3.

pixelflips (approved):
LGTM! Thanks for adding the test as well! 👍🏼

PR #811, opened 2026-09-23
feat(pds-box): support per-side border control

pixelflips:
# Description

Fixes: [DSS-249](https://linear.app/kajabi/issue/DSS-249/featpds-box-support-per-side-border-control)

`pds-box` only exposed `border` as a boolean — all four sides or none. There was no way to render a bottom-only divider, a sidebar rail, or a stacked list where the last item drops its bottom border.

This was inconsistent with the rest of the component: `pds-box` already models per-side spacing exhaustively (`paddingBlockStart`, `paddingInlineEnd`, `marginBlockEnd`, …), so the per-side logical-property vocabulary already existed in the component's design language. Borders just never got it.

This PR adds four optional props using that same vocabulary:

- `borderBlockStart`
- `borderBlockEnd`
- `borderInlineStart`
- `borderInlineEnd`

Semantics follow CSS longhand-over-shorthand, so there's nothing new to learn: `border` sets all four sides, and a per-side prop overrides the side it names. `true` forces the border on, `false` forces it off, unset defers to `border`.

```html
<pds-box border-block-end="true">                  <!-- divider row -->
<pds-box border-inline-end="true">                 <!-- sidebar rail -->
<pds-box border="true" border-block-end="false">   <!-- last item in a stacked list -->
```

**No new dependencies.**

### Why four props instead of widening `border`

Widening `border` to a `boolean | string` union looks additive in TypeScript but is a breaking change at the HTML attribute layer. Stencil's `propTypeFromTSType` falls back to `any` when a union spans more than one primitive kind, and `parsePropertyValue` only special-cases the literal string `"false"` inside its `Boolean` branch. Under `any` that branch never runs, so `border="false"` would become a truthy string and start drawing a border where it previously drew none. `border="false"` is a live idiom in this library — `pds-copytext` both documents and tests it.

Four new optional props keep `border` typed `boolean`. The generated `components.d.ts` confirms the new props landed as `boolean` too, so `border-block-end="false"` coerces correctly.

### Implementation note

The per-side "off" rules use an element-qualified selector (`pds-box.pds-box--border-block-end-none`) so they outrank the all-sides `.pds-box--border` rule on specificity rather than on stylesheet source order. That's the one non-obvious line in the diff and it carries a comment explaining why.

Tracking: https://linear.app/kajabi/issue/DSS-249/featpds-box-support-per-side-border-control

## Type of change

# How Has This Been Tested?

Full core suite run locally: **97 suites, 3104 tests, all passing.** ESLint reports 0 errors and adds no new warnings against the repo's existing 340-warning baseline. Stylelint clean on the component's SCSS.

**Spec tests** (`test/pds-box.spec.tsx`) — a new `per-side border props` block covering each side set to `true`, each side set to `false`, all sides unset, a per-side prop combined with `border`, and two sides set together. These assert the class output and the tri-state behavior.

**e2e tests** (`test/pds-box.e2e.ts`, new file) — the spec tests only assert class names, so these assert computed styles in a real browser, which is the only layer that can verify the SCSS. Most importantly, the second test (`border="true" border-block-end="false"` → block-end `0px`, other three `1px`) proves the specificity-based override actually resolves as designed.

To reproduce:

```
npx nx run @pine-ds/core:test
npx nx run @pine-ds/core:lint
```

- [x] unit tests
- [x] e2e tests
- [x] accessibility tests
- [ ] tested manually
- [ ] other:

**Test Configuration**:

- Pine versions: 3.31.1
- OS: macOS (Darwin 25.6.0)
- Browsers: Chromium (Puppeteer, via Stencil e2e)
- Screen readers: none
- Misc: axe-core via the shared `runAxe` helper, WCAG 2.0/2.1 A + AA

# Checklist:

---

### Out of scope

- **Breakpoint variants** (`borderBlockEndMd`, …). Following the spacing precedent literally would mean 4 sides × 6 breakpoints = 24 props. `border` itself has no breakpoint variants either, so the new props are consistent with what they extend. The naming leaves room if demand appears.
- **Per-side border color and width.** All sides share `--color-border-box` and `--border-width-default`.

---

## UX/Product Change Summary

### Detected Changes

| API Contract | Low | Four new public props (`borderBlockStart`, `borderBlockEnd`, `borderInlineStart`, `borderInlineEnd`) are added to `pds-box` and exposed in the generated type declarations, expanding the component's public API for every consumer. All four default to `undefined`, so existing `pds-box` usage renders identically — this is purely additive. | `pds-box.tsx:123`, `components.d.ts:275` |
| Data Display | Medium | The render method now emits per-side border classes, and new SCSS applies `border-<side>: var(--border-width-default) solid var(--color-border-box, var(--pine-color-border))` for each side, so consumers can draw a border on one edge without `border`. This is a visible change to rendered box chrome anywhere the new props are used. | `pds-box.tsx:739`, `pds-box.scss:132` |
| Default Value | Medium | Setting a side to `false` emits a `-none` class whose selector is deliberately scoped to the element (`pds-box.pds-box--border-block-end-none`), raising specificity to 0-1-1 so it outranks `.pds-box--border`. That also outranks any single-class override a consuming app has written against `pds-box`, meaning app-level border styling can now be silently suppressed on that edge. | `pds-box.scss:139` |

<details>

- `libs/core/src/components/pds-box/pds-box.tsx:123`
- `libs/core/src/components/pds-box/pds-box.tsx:739`
- `libs/core/src/components/pds-box/pds-box.scss:132`
- `libs/core/src/components/pds-box/pds-box.scss:139`
- `libs/core/src/components.d.ts:275`

</details>

### Risk Assessment

Low regression risk for existing consumers since all four props default to `undefined` and emit no classes, but the `-none` rules intentionally escalate selector specificity above `.pds-box--border`, which can also override consumer-authored border styles on `pds-box`. Verify in a real consumer app that a `false` side does not defeat app-level border overrides, and confirm `border-color` resolves correctly when a per-side shorthand and the `border` prop are combined.

### Action Items


pixelflips:
@coderabbitai full review.

QuintonJason (approved):
This is good to go

PR #814, opened 2026-09-23
feat(pds-divider): add label prop

pixelflips:
# Description

Adds a `label` prop to `pds-divider` that renders text centered between two lines, so a labeled separator (a chat date divider like "──── Today ────", or a "Your sites" section break) takes one component instead of two `pds-divider`s and a hand-centered text node.

```html
<pds-divider label="Today"></pds-divider>
```

- **Rendering:** with no label, the divider renders the same `<hr>` as today, so existing usage and snapshots don't change. With a label, it renders a `div` with `role="separator"` and `aria-label` set to the label, so screen readers announce "Today, separator". The visible label `<span>` is `aria-hidden="true"`. Chrome keeps a separator's child text in the accessibility tree, so without it the label could be read twice. A whitespace-only label falls back to the plain `<hr>`. The two lines are `::before`/`::after` flex pseudo-elements.
- **Styling:** the label uses `--pine-typography-body-sm-medium` (12px), a step smaller than the 14px in the Figma mock, with a 16px (`--pine-dimension-sm`) gap and `--pine-color-text`. The lines use `--pine-color-border`, which already swaps in dark mode.
- **`label` shadow part:** consumers can restyle the label text, e.g. `pds-divider::part(label) { color: var(--pine-color-text-secondary); }`.
- **`offset`:** works with the label. The labeled variant drops the offset padding, because padding would inset the pseudo-element lines instead of bleeding them.
- **`vertical`:** the label is ignored when `vertical` is `true`, and this is documented.

|  Sceenshot   |
|--------|
| Screenshot 2026-09-23 at 12 46 48 PM|

No new dependencies.

Fixes [DSS-233](https://linear.app/kajabi/issue/DSS-233/pds-divider-add-a-labeled-centered-content-variant)

**Follow-up:** `libs/figma/components/pds-divider.figma.ts` isn't updated. The Figma divider component has no Label property yet (the mock builds the pattern from two dividers plus a text layer), and mapping a property that doesn't exist would break the Code Connect publish.

## Type of change

# How Has This Been Tested?

- [x] unit tests: labeled render with `role`/`aria-label`/`part`, label + offset, label ignored when vertical, and empty or whitespace-only label falls back to `<hr>`
- [x] e2e tests: both lines render at equal width in a 400px container, `offset="lg"` bleeds the labeled divider to 464px, and `::part(label)` overrides the computed color
- [x] accessibility tests: axe run on the labeled divider, plus an accessibility-tree check that the separator is named "Today" and has no duplicate text node
- [x] tested manually: new **components/Divider → With Label** story checked in Storybook

**Test Configuration**:

- Pine versions: 4.0.0 (rebased on `main`)
- OS: macOS
- Browsers: Chrome (Storybook); headless Chromium (e2e)
- Screen readers: not tested manually. The accessible name is covered by the `aria-label` spec assertion and axe
- Misc:

# Checklist:

---

## UX/Product Change Summary

### Detected Changes

| UI State Changes | Low | `pds-divider` has a new optional `label` prop. When `label` is set and `vertical` is false, the divider shows centered text (body-sm-medium typography, `--pine-color-text`) with a line on each side instead of one plain rule. The change is additive: dividers without a label, and all vertical dividers, still render the same `<hr>`. | `pds-divider.tsx:18` |
| Data Display Changes | Low | A labeled divider renders a `<div role="separator" aria-label={label}>` in place of an `<hr>`, so screen readers now announce the label text. The inner `<span>` exposes a new `label` CSS part that consumers can use to restyle the text, such as its color. | `pds-divider.tsx:60` |
| UI State Changes | Low | The offset behavior changes for labeled dividers. They keep the negative margin that makes them bleed past their container, but their padding is set to 0 so the pseudo-element lines reach the full bleed width instead of being inset. Unlabeled offset dividers keep the existing margin plus padding. | `pds-divider.scss:57` |

<details>

- `libs/core/src/components/pds-divider/pds-divider.tsx:18`
- `libs/core/src/components/pds-divider/pds-divider.tsx:60`
- `libs/core/src/components/pds-divider/pds-divider.scss:25`
- `libs/core/src/components/pds-divider/pds-divider.scss:57`

</details>

### Risk Assessment

Risk is low. The new prop is opt-in and the unlabeled `<hr>` path is unchanged, so existing uses shouldn't change visually. Check that labeled dividers look right in dark mode (they use the `--pine-color-border` and `--pine-color-text` tokens), that long labels wrap cleanly, and that screen readers announce the `aria-label` on `role="separator"` as expected.

### Action Items

---

> [!NOTE]
> **Low Risk**
> Opt-in UI component API; existing unlabeled `<hr>` rendering is unchanged, with tests covering accessibility and layout.
> 
> **Overview**
> Adds an optional **`label`** prop to **`pds-divider`** so horizontal dividers can show centered text between two lines (e.g. chat date breaks) without composing separate rules and text.
> 
> When **`label`** is non-empty and **`vertical`** is false, the component renders a **`role="separator"`** with **`aria-label`** and flex **`::before`/`::after`** lines instead of a plain **`hr`**; unlabeled and vertical dividers behave as before. Consumers can restyle the text via the new **`label`** shadow part (**`pds-divider::part(label)`**). **`offset`** still bleeds labeled dividers past the container by clearing padding on the labeled variant so the pseudo-element lines span the full bleed width.
> 
> Docs, Storybook (**With label**), generated types/readme, and unit/e2e/a11y tests cover layout, offset bleed, part styling, and accessible naming.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f8271d354b04e780abdd8995a5d3f1f706d7973. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


libs/core/src/components/pds-divider/pds-divider.tsx
```diff
@@ -40,7 +57,13 @@ export class PdsDivider {
   render() {
     return (
       <Host id={this.componentId}>
-        <hr class={this.classNames()} />
+        {this.hasLabel ? (
+          <div class={this.classNames()} role="separator" aria-label={this.label}>
+            <span class="pds-divider__label" part="label">{this.label}</span>
```
QuintonJason: **Should-fix:** this inner `<span>` has no `aria-hidden="true"`. `role="separator"` is the right call (a separator's accessible name is author-supplied only, never from content — that's exactly why `aria-label` is set on the host), but the label text is still visible child content, and real screen readers (esp. NVDA/JAWS in browse mode) sometimes read a non-focusable container's child text nodes regardless of role-based suppression. Without `aria-hidden`, that risks a double-announcement: "Today, separator... Today."

The PR itself discloses screen readers weren't tested manually, so this can't be confirmed either way without a real AT pass — `aria-hidden="true"` on this span is the standard, cheap mitigation that removes the ambiguity outright.

libs/core/src/components/pds-divider/pds-divider.tsx
```diff
@@ -22,13 +31,21 @@ export class PdsDivider {
    */
   @Prop() vertical = false;
 
+  private get hasLabel() {
+    return Boolean(this.label) && !this.vertical;
```
QuintonJason: **Nit:** `Boolean(this.label)` is truthy for a whitespace-only label (`label="   "`), so it renders the full labeled layout with a visually-empty gap where the text should be — reads as a rendering glitch rather than the clean "no label" `<hr>` fallback. `Boolean(this.label?.trim())` would fold that case into the same fallback path as an empty string. Low-likelihood in practice, cheap to close.

libs/core/src/components/pds-divider/pds-divider.tsx
pixelflips: @QuintonJason Fixed in d56619da. The label `<span>` now has `aria-hidden="true"`.

I checked Chrome's accessibility tree before changing it, and you're right about the risk. The separator is named "Today" from `aria-label`, but a `StaticText "Today"` node is still underneath it. Children-presentational removes the span's role (it shows up as `none`), not its text, so a browse-mode reader could hit the label twice. Hiding the span doesn't change the accessible name because that comes only from `aria-label`.

To keep it from regressing, 2f8271d3 adds an e2e test that snapshots the accessibility tree. It asserts the separator's name is "Today" and that it has no text descendants. I confirmed the test fails with `aria-hidden` removed. Still no manual NVDA/JAWS/VoiceOver pass, so that part of the PR description stands.

libs/core/src/components/pds-divider/pds-divider.tsx
pixelflips: @QuintonJason Fixed in d56619da. `hasLabel` is now `Boolean(this.label?.trim()) && !this.vertical`, so `label="   "` falls back to the plain `<hr>`, the same as an empty string. That also avoids a whitespace-only `aria-label`. 2f8271d3 adds a spec test for the whitespace case next to the existing empty-label test.

QuintonJason (approved):
Thanks for addressing the feedback. Good to go

